PR #5, opened 2022-01-17
Implement screen source modal

streamer45:
#### Summary

PR implements a modal for selecting screen sources from the Desktop app. As this required some changes upstream, the minimum support version for this to show will be 5.1.0.

I opted to implement the simplest design as currently there's no way to determine the type of source shared so no way to group them either. 

#### Screenshot

2022-01-17-144124_2560x1080_scrot

#### Video

https://user-images.githubusercontent.com/1832946/149780935-a535eb4d-c583-460a-bd08-800a9e1b11d9.mp4


abhijit-singh (commented):
Thanks @streamer45! The functionality looks good!

A few things (It's okay if you think all or some of these should be done in a follow-up PR):
1. Can we pre-select the first option by default?
2. Limit the max-width and max-height of the modal as per the mockup below. Here are the [general guidelines for modals](https://zeroheight.com/29be2c109/p/426c9b-modals) across the product.
image
3. Make header text left aligned and change the font size as per [these guidelines](https://zeroheight.com/29be2c109/p/426c9b-modals/i/46851901).
4. Can we make the aspect ratio for the thumbnails close to 16:9? Windows of random sizes can be centred vertically/horizontally within that space, as shown in the mockup attached above. 
5. Add 8px of space between the thumbnail and the name as marked in the mockup. The selection border can still be on the 16:9 frame.

streamer45:
Thanks for the thorough review @abhijit-singh . Now I see what you meant when you suggested to be added as a reviewer for this kind of PRs. I definitely want to align to our design principles which I personally found fascinating (thanks for the links). So much I need to plan some retroactive work to polish everything else that was merged already.

I tried to address all your points, please let me know if there's anything else.

image

abhijit-singh (approved):
Thank you, @streamer45! Glad you found them valuable!

The changes look good but listing a couple more minor things that I could spot. They're non-blocking so approving the PR. 

1. What happens if the title of a window is long? I can see the title of the second window in your screenshot (Developer tools...) is already extending beyond the bounds I'd expect. I'd suggest we truncate the title after a certain width or fixed number of characters (about 36 based on the space available). If that's high effort, I'd be okay with wrapping as well. Just want to avoid overlapping of text. 
2. Can we vertically centre align the close [icon button](https://zeroheight.com/29be2c109/p/00b099-icon-buttons)? Attaching a screenshot with the spacing marked up and also showing the hover state.
image

streamer45:
@abhijit-singh Yeah, long text should truncate, I forgot to set a width and to hide the overflow. Also vertically aligned the close icon. Should be all good now:

image

PR #7, opened 2022-02-24
[MM-42078] Fix possible multiple component registrations

streamer45:
#### Summary

I originally thought this bug was somehow due to the how community operates (pods rolling during upgrades) cause I wasn't able to reproduce but after a good amount of pure try and fail I was able to figure out it was switching channels/teams very quickly that caused it. This is because I failed to account of an async call which was delaying registering operations eventually leading to a leak.

The other issue (duplicate call icon next to channel name) is still a bit of a mystery since the component is only ever registered at plugin initialization so trying to force an unsubscribe although I don't believe it to be necessary. Will monitor for further occurrences.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-42078

### DIFF
--- a/webapp/src/index.tsx
+++ b/webapp/src/index.tsx
@@ -235,7 +235,8 @@ export default class Plugin {
 
     public async initialize(registry: PluginRegistry, store: Store<GlobalState>): Promise<void> {
         registry.registerReducer(reducer);
-        registry.registerSidebarChannelLinkLabelComponent(ChannelLinkLabel);
+        const sidebarChannelLinkLabelComponentID = registry.registerSidebarChannelLinkLabelComponent(ChannelLinkLabel);
+        this.unsubscribers.push(() => registry.unregisterComponent(sidebarChannelLinkLabelComponentID));
         registry.registerChannelToastComponent(ChannelCallToast);
         registry.registerPostTypeComponent('custom_calls', PostType);
         registry.registerNeedsTeamRoute('/expanded', ExpandedView);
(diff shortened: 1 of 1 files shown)


webapp/src/index.tsx
```diff
@@ -416,14 +417,14 @@ export default class Plugin {
             const roles = getMyRoles(store.getState());
             const cms = getMyChannelMemberships(store.getState());
 
-            registry.unregisterComponent(channelHeaderMenuID);
-
             try {
                 const resp = await axios.get(`${getPluginPath()}/config`);
+                registry.unregisterComponent(channelHeaderMenuID);
```
cpoile, line 422: I'm not really sure what this change does... We're putting the `unregisterComponent` call in the `try`, just incase the `unregisterComponent` throws? But the `unregister` is a `dispatch` call, which can't throw (unless we use middleware that does?) -- so I'm not really sure if this change does anything? 
But anyway, I'll approve because it doesn't really hurt I guess. :)

cpoile (approved):

webapp/src/index.tsx
streamer45: It's a bit tricky but I'll try to explain. This line change is the fix for:

image

We need to do the unregistration after the async call that fetches the config otherwise it will risk of running multiple times (e.g. quickly switching channels) before we get a chance to register again down below which means we'll "lose" unregister calls. Then the register calls will all happen once those queued async calls finish which will cause a leak given the unregisters already happened with stale ids.

This is a sample sequence of the bug happening:

- Channel switch 1
  - unregistering (this is a no-op as there's nothing registered yet, `channelHeaderMenuID` is empty)
  - registered ID A
- Channel switch 2:
  - unregistering ID A
  - async call taking some time
  - Channel switch 3
    - unregistering ID A again (wasting a call)
    - async call
    - async call returned
    - registering ID C
  - async call returned
  - registering ID B

streamer45:
@cpoile Going to merge this so we get the fix on Community today but please feel free to keep commenting if anything is unclear.

webapp/src/index.tsx
cpoile: Ah, gotcha. Very clear explanation, thank you!

PR #8, opened 2022-02-24
Add userID to call started telemetry event

streamer45:
#### Summary

This didn't get included originally cause I thought we could figure it out by looking at the first user who joined a call but a little redundancy makes for easier queries on the analytics side so we are adding it here as well.


cpoile (approved):
👍

PR #9, opened 2022-02-25
[MM-42102] Fix tracks sending goroutine getting stuck due to missing signaling message

streamer45:
#### Summary

This was a rather odd bug that happened today for the first time (as far as I can recall). Some minutes into the call new users joining were not able to be heard from other participants even though they were unmuted and sending voice data.

Luckily I was able to reproduce this behaviour on a test cluster and examined the goroutine dump which revealed a goroutine stuck sending tracks to some users:

```
goroutine 15365 [chan send, 7 minutes]:
main.(*Plugin).initRTCConn.func5.1(0xc003544480)
        github.com/mattermost/mattermost-plugin-calls/server/sfu.go:369 +0x69
main.(*Plugin).iterSessions(0xc000434180, {0xc00594cd80, 0x1a}, 0xc005257f00)
        github.com/mattermost/mattermost-plugin-calls/server/utils.go:51 +0x14a
main.(*Plugin).initRTCConn.func5(0xc005a3ea20, 0xc0046e6798)
        github.com/mattermost/mattermost-plugin-calls/server/sfu.go:365 +0x7b6
created by github.com/pion/webrtc/v3.(*PeerConnection).onTrack
        github.com/pion/webrtc/v3@v3.1.16/peerconnection.go:458 +0x17d
```
https://github.com/mattermost/mattermost-plugin-calls/blob/2532e83dd686a5817d311d1ffb1c41f1cef33801/server/sfu.go#L369

The way the code is written didn't account for this to happen. `tracksCh` had some buffer but given there's a reader dedicated to this (`handleTracks`) eventually it should unblock. What I forgot to account for was a user stuck during initial signaling, meaning the reader goroutine never got created in the first place:

```
goroutine 6427 [chan receive, 8 minutes]:
main.(*Plugin).initRTCConn(0xc000434180, {0xc003131020, 0x101edec})
        github.com/mattermost/mattermost-plugin-calls/server/sfu.go:464 +0x705
main.(*Plugin).startSession.func2()
        github.com/mattermost/mattermost-plugin-calls/server/plugin.go:55 +0x1d8
created by main.(*Plugin).startSession
        github.com/mattermost/mattermost-plugin-calls/server/plugin.go:51 +0x165
```
https://github.com/mattermost/mattermost-plugin-calls/blob/2532e83dd686a5817d311d1ffb1c41f1cef33801/server/sfu.go#L464

I believe this bug triggered partially due to how the simulated client was written, not accounting for failures. I wouldn't expect this to happen with a well implemented client (e.g. a browser). Either way no client should cause a stuck goroutine (plus leak) so it's always a good idea to fix it.

Main code changes are:

- Adding a timeout (10s) to receiving signaling messages. 
- Making the track sending operation unblocking. If the channel gets full an error gets logged. This shouldn't generally happen as it means someone will not be receiving a track but as things stand we cannot block as it's running in a loop.
- Increasing the channel size to account for the case where a lot of users unmute at the same time.

There are still things to consider and improve as timing out without letting the client know will leave the connection pretty much unusable as it will be stuck in an unexpected signaling state. As the change here would be more tricky I will defer it for now to be addressed during the new service's implementation.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-42102

### DIFF
--- a/server/session.go
+++ b/server/session.go
@@ -11,7 +11,9 @@ import (
 )
 
 const (
-	msgChSize = 20
+	msgChSize        = 20
+	tracksChSize     = 10
+	signalingTimeout = 10 * time.Second
 )
 
 type session struct {
@@ -51,11 +53,11 @@ func newUserSession(userID, channelID, connID string) *session {
 		signalOutCh:   make(chan []byte, msgChSize),
 		wsMsgCh:       make(chan clientMessage, msgChSize*2),
 		wsCloseCh:     make(chan struct{}),
-		tracksCh:      make(chan *webrtc.TrackLocalStaticRTP, 5),
+		tracksCh:      make(chan *webrtc.TrackLocalStaticRTP, tracksChSize),
 		iceCh:         make(chan []byte, msgChSize*2),
 		closeCh:       make(chan struct{}),
 		doneCh:        make(chan struct{}),
-		trackEnableCh: make(chan bool, 5),
+		trackEnableCh: make(chan bool, tracksChSize),
 		rtpSendersMap: map[*webrtc.TrackLocalStaticRTP]*webrtc.RTPSender{},
 	}
 }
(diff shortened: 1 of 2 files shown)


cpoile (approved):
Wow, really nice find, and excellent summary!

PR #10, opened 2022-02-28
[MM-42101] Implement option to enable calls by default

streamer45:
#### Summary

We are adding a new `DefaultEnabled` config option that controls whether Calls should be enabled in all channels by default (opt-out model). 

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-42101

### DIFF
--- a/server/session.go
+++ b/server/session.go
@@ -64,10 +64,24 @@ func newUserSession(userID, channelID, connID string) *session {
 
 func (p *Plugin) addUserSession(userID, channelID string, userSession *session) (channelState, error) {
 	var st channelState
+
+	cfg := p.getConfiguration()
+
 	err := p.kvSetAtomicChannelState(channelID, func(state *channelState) (*channelState, error) {
 		if state == nil {
-			return nil, fmt.Errorf("channel state is missing from store")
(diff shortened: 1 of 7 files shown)


server/session.go
```diff
@@ -64,10 +64,24 @@ func newUserSession(userID, channelID, connID string) *session {
...
+					Enabled: true,
+				}
+			} else {
+				return nil, fmt.Errorf("channel state is missing from store")
+			}
 		}
+
+		if !state.Enabled {
+			return nil, fmt.Errorf("calls are not enabled")
+		}
```
streamer45, line 83: Interestingly, from a server perspective it was possible to start a call on a channel with calls disabled.

webapp/src/client.ts
```diff
@@ -112,8 +112,7 @@ export default class CallsClient extends EventEmitter {
         this.ws = ws;
 
         ws.on('error', (err) => {
-            console.log(`ws error: ${err}`);
-            this.ws = null;
```
streamer45, line 116: This was preventing the ws connection to disconnect in case of server side error (e.g. during join).

cpoile: Nice catch!

cpoile (approved):
👍

PR #13, opened 2022-03-02
[MM-42198] Fix wrapping of long channel names

streamer45:
#### Summary

PR fixes the issue although I am also looking for some general feedback to see if we need to plan extra changes.

As things stand we have space for up to 17 characters in single team mode and 26 with the sidebar. If we don't set a max width very long names tend to go a bit over the right almost touching the expand icon which may feel a little compressed (see picture). 

Test string was `Some very very long channel name`.

#### Single team, no sidebar

image

#### With sidebar

image

/cc @itao 

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-42198


cpoile (approved):

abhijit-singh (commented):
@streamer45 We should probably reduce the char limit a little more to prevent overlap with the hover state of the icon. I'd recommend 24 chars with the team sidebar and 14 char without it.
image

streamer45:
Thanks, updated. It's not 100% accurate in terms of number of chars displayed cause not every char has the same width but there's a hard limit now.

image

image

abhijit-singh (approved):

PR #14, opened 2022-03-03
[MM-42190] Auto injecting, versioning and releasing (support marketplace)

cpoile:
#### Summary
- Rudder credentials are injected
- Version is set based on the current tag (or last tag + short hash if the current commit isn't tagged)
- Using gotestsum for tests (it's cleaner easier during dev)
- Manifests are generated (and aren't tracked anymore)
- To generate a release: push a tag (that's it)
- This should make matterbuild work, but I still need to fix Matterbuild a bit (if we don't build for windows).

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-42190


Makefile
```diff
@@ -52,7 +56,7 @@ gomod-check:
 
 ## Runs eslint and golangci-lint
 .PHONY: check-style
-check-style: gomod-check golangci-lint webapp/node_modules
+check-style: apply gomod-check golangci-lint webapp/node_modules
```
streamer45, line 59: Unrelated but since we are here can we invert and do the `gomod-check` as last? I've figured out it's super annoying when you are working on some uncommitted changes that import new things and you can't run `make check-style` cause it will always fail.

.circleci/config.yml
```diff
@@ -134,12 +136,56 @@ jobs:
       - run:
           name: Building Plugin Bundle
           command: make dist
+      - run:
+          name: Generating Release Notes
+          command: |
+            printf "Supported Mattermost Server Versions: **$(cat plugin.json | jq .min_server_version -r)+** \n## Enhancements\n\n## Fixes\n"  >> dist/release-notes.md
```
streamer45, line 142: Are the "Enhancements" ever getting written this way? I am totally fine with just having the raw list of commits at this point.

streamer45 (approved):
Awesome :100:

.circleci/config.yml
cpoile: Good point! Removed that, and just renamed it to `Commits`.

cpoile:
Also added a `dist-ci` make target so we only  bundle and distribute the linux binary.

streamer45 (approved):
Perfect :+1:

PR #15, opened 2022-03-07
[MM-42347] Fix UDP conn not respecting deadlines

streamer45:
#### Summary

PR fixes an issue with the UDP server connection not respecting the defined deadlines and hence potentially getting stuck indefinitely in case of an unreachable host.

After some testing and looking into the standard library I've figure out this was actually caused by the call on [`File.Fd()`](https://pkg.go.dev/os#File.Fd) we do here:

https://github.com/mattermost/mattermost-plugin-calls/blob/4ac8eb9ee52ed2d83a76a7ed8e6740c2ec54d34c/server/activate.go#L55-L64

As suggested in the official documentation we are now using [`File.SyscallConn()`](https://pkg.go.dev/os#File.SyscallConn) instead which seems to avoid this problem.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-42347

### DIFF
--- a/server/activate.go
+++ b/server/activate.go
@@ -27,6 +27,10 @@ func (p *Plugin) OnActivate() error {
 	}
 
 	cfg := p.getConfiguration()
+	if err := cfg.IsValid(); err != nil {
+		p.LogError(err.Error())
+		return err
+	}
 
 	udpServerConn, err := net.ListenUDP("udp4", &net.UDPAddr{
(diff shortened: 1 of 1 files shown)


streamer45:
@agnivade This is strictly related to the issue I mentioned to you a while back.

cpoile (approved):
Whoa, nice debugging!

agnivade:
Yep saw it, nice work!

PR #16, opened 2022-03-07
[MM-42086] Use messaging to get screenshare sources for Desktop App

devinbinnie:
#### Summary
`desktopCapturer.getSources` has been deprecated for use in the renderer process in Electron v17: https://www.electronjs.org/docs/latest/breaking-changes#removed-desktopcapturergetsources-in-the-renderer 

So as part of the upgrade, we need to remove the use of the library in Calls and just pass a message up through the preload script like we do for notifications.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-42086

### DIFF
--- a/e2e/tests/start_call.spec.ts
+++ b/e2e/tests/start_call.spec.ts
@@ -9,7 +9,6 @@ declare global {
     interface Window {
         callsClient: any,
         desktop: any,
-        desktopCapturer: any,
     }
 }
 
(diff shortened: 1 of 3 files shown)


e2e/tests/start_call.spec.ts
```diff
@@ -70,16 +69,21 @@ test.describe('desktop', () => {
...
-                getSources: async () => {
-                    return [
+            window.addEventListener('message', (event) => {
+                if (event.data.type !== 'get-desktop-sources') {
+                    return;
+                }
+
+                window.postMessage({
+                    type: 'desktop-sources-result',
+                    data: [
```
streamer45, line 79: `data` should become `message` as that's the new expected format.

streamer45 (commented):
Thanks @devinbinnie for making Calls desktop compliant :)
Left a couple of minor suggestions.

streamer45 (approved):
Awesome :+1:

cpoile (approved):
Awesome!

PR #17, opened 2022-03-08
Fix UI regressions

streamer45:
#### Summary

PR fixes a couple of issues that were making e2e tests fail due to recent UI changes.
Looking forward to have them automated so we find out right away :p


cpoile (approved):
Ahhh... Missed that use of the border in the toast. Sorry!

PR #24, opened 2022-03-10
[MM-42427] Add a `/call announce` slash command to notify current calls

cpoile:
#### Summary
- Use `/call announce` to make an announcement to all channels with ongoing calls
- The call bot icon is just a placeholder until Abhijit gets back (https://mattermost.atlassian.net/browse/MM-42466)
- TODO: for some reason `@here` `@channel` etc. don't seem to trigger mentions -- there's something simple I'm missing here
- TODO: we should probably add some kind of security to the command. I was thinking of a whitelist in the system console for usernames that can use the announce command. What do you think?

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-42417


streamer45:
It's interesting how I envisioned this in a totally different way implementation wise. Not a bad thing though, I thought of a more custom ephemeral/notification kind of announce probably cause I am not a huge fan of bots.

Some limitations I see:

- Unless the user is looking in the channel they won't get notified. I am guessing this is addressable by making global mentions work. The reasons they are not working, I assume, is because the bot is not a member of the channel.
- This doesn't cover the case of users that are not in a call but may want to start one while the service is interrupted. We may consider this a lesser problem but just wanted to mention it.

As for permissions, I am thinking we should leave this to System Admins only at this point as its main purpose (for the time being that is) will be to announce potential service interruptions.

cpoile:
@streamer45 I was thinking of ephemeral too, but then you would have to loop and send an ephemeral message to every member of the channel individually -- and the `@here` won't send a notification. I did some experiments, and the only way the `@here` will trigger a notification is if the message is posted with the user's id (the bot won't trigger the notification, even if they've been added as a member of the channel beforehand). 

So, would it seem an invasion of privacy if some sysadmin user posted a message in a private channel or DM channel?

streamer45:
> @streamer45 I was thinking of ephemeral too, but then you would have to loop and send an ephemeral message to every member of the channel individually -- and the `@here` won't send a notification. I did some experiments, and the only way the `@here` will trigger a notification is if the message is posted with the user's id (the bot won't trigger the notification, even if they've been added as a member of the channel beforehand).

I see. I am wondering if it would be simpler to instead broadcast a special websocket event to the affected channels and handle the display on the client? I understand that leveraging the notification system would have been easier and probably cleaner but there seem to be a lot of limitations.

> So, would it seem an invasion of privacy if some sysadmin user posted a message in a private channel or DM channel?

It's borderline, not sure we want to get there if we can avoid it.

cpoile:
@streamer45 I just realized - we can have the bot DM all calls participants with the announcement message--that will cause a notification. Along with a public post in each calls channel (for people who weren't in the call at that moment), that should cover all cases. What do you think?

streamer45:
> @streamer45 I just realized - we can have the bot DM all calls participants with the announcement message--that will cause a notification. Along with a public post in each calls channel (for people who weren't in the call at that moment), that should cover all cases. What do you think?

I think that's a good solution that covers most cases with the small exception of calls that have not started yet. Maybe in the future we could refine this through some heuristic based on whether a channel has been having recent calls. Anyhow, this is more than enough to cover our current needs.

My only lasting concern is on whether we should include this in the main branch. I am thinking we could first merge it in community and see how that goes. What do you think?

cpoile:
Closing for now. If we find a need for it, we can reopen and I can: 1. merge, 2. add sysadmin security check, and 3. add the proposed DMing solution.

PR #25, opened 2022-03-11
[MM-42385] Fix closing popout while screensharing from it

streamer45:
#### Summary

Fixes https://github.com/mattermost/mattermost-server/issues/19710

We weren't properly de-initializing resources when the window got closed and screen sharing was initiated from it.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-42315

### DIFF
--- a/webapp/src/client.ts
+++ b/webapp/src/client.ts
@@ -361,9 +361,11 @@ export default class CallsClient extends EventEmitter {
             return;
         }
 
-        this.streams.push(screenStream);
         const screenTrack = screenStream.getVideoTracks()[0];
         this.localScreenTrack = screenTrack;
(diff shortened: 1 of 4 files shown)


cpoile (approved):
Cool! 
Do you think this is similar to the problem of the webapp freezing when closing the popout on Chrome?

streamer45:
> Cool! Do you think this is similar to the problem of the webapp freezing when closing the popout on Chrome?

I don't think it's strictly related cause that issue will trigger with or without screensharing. Also I literally had to test this locally on Firefox and on a remote server for Chrome cause it was freezing which made it impossible to reproduce the bug :grimacing:

PR #27, opened 2022-03-11
[MM-42433] Fix: Calls mobile: microphone state gets reset when app switching

cpoile:
#### Summary
- Turns out this was caused by forgetting to send over the state for each user in the GET `/channels` route 😄 

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-42433


streamer45 (approved):
Nice catch :+1:

PR #31, opened 2022-03-15
[MM-42416] Calls: RHS Custom post styling

cpoile:
#### Summary
- It looks like the new avatar styling affected the enclosing custom post styling. This fixes it. 

- How the custom post /used/ to look (before the avatar styling):
image image

- How the custom post looks on main (after the avatar styling):
image image

- How it looks now:
image image

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-42416


abhijit-singh (commented):
Thanks for fixing this @cpoile. Difficult to test it, but would the layout break when there are a couple more avatars to show over there? Here's a proposed layout I had in the designs if it's not too complicated to implement: 
image

cpoile:
@abhijit-singh I've tried to match the figmas, but I couldn't figure out a simple way to conditionally add the separator (either as a div or a border) based on whether the line wrapped. According to what I found online, it's not possible to detect whether a wrap occurred and apply css conditionally, unless you get into using JS (which I think is a bit of overkill for this). I also converted the hardcoded colors to theme colors.

@streamer45 I converted this component to styled-components -- once I did that, the rest was much easier.

Here are the various combinations and some theme versions:
image
image

image
image
image
image
image
image
image

webapp/src/components/post_type/component.tsx
```diff
@@ -21,163 +19,182 @@ interface Props {
...
-                    {`Lasted ${moment.duration(this.props.post.props.end_at - this.props.post.props.start_at).humanize(false)}`}
-                </span>
-            </div>
-        ) : (
-            <span style={this.style.duration}>{moment(this.props.post.props.start_at).fromNow()}</span>
-        );
-
-        return (
-            <div
-                className='call-thread'
```
streamer45, line 119: Could we keep the class? This is actually used by e2e tests to select the element.

webapp/.eslintrc.json
```diff
@@ -48,7 +48,7 @@
     "array-callback-return": 2,
     "arrow-body-style": 0,
     "arrow-parens": [
-      2,
+      1,
```
cpoile, line 51: this (and below) makes the prop arrow functions so much easier to read.

webapp/src/components/connected_profiles.tsx
```diff
@@ -17,11 +17,11 @@ interface Props {
 const ConnectedProfiles = ({pictures, profiles, maxShowedProfiles, size, fontSize, border}: Props) => {
     maxShowedProfiles = maxShowedProfiles || 2;
     const diff = profiles.length - maxShowedProfiles;
+    const maxRightShift = 8 * maxShowedProfiles;
     profiles = diff > 0 ? profiles.slice(0, maxShowedProfiles) : profiles;
 
-    const margin = -8;
     const els = profiles.map((profile, idx) => {
-        const off = diff > 0 ? (profiles.length - idx) * 8 : (profiles.length - 1 - idx) * 8;
+        const rightShift = 8 * idx;
```
streamer45, line 24: I am thinking this is going to regress the spaning issue we've seen in https://github.com/mattermost/mattermost-plugin-calls/pull/11 :thinking:

streamer45 (commented):
Great work @cpoile  :+1:   Left some minor comments.

webapp/src/components/post_type/component.tsx
cpoile: Oh! Hah, sorry. Will change that to a `data-testid` then (or I'll probably try to delete it in the future when I can't find the css class it belongs to ;) )

webapp/src/components/connected_profiles.tsx
cpoile: Ah, dammit. Well, we have to pick what we want then -- either:
1. The avatars shift slightly to the left horizontally in the single line post:
image
vs
image
vs
image

or:
2. The avatars shift slightly to the right in the double line post (they line up:
image
image
image

@abhijit-singh your choice!

cpoile: I think I prefer the horizontal shift (1), since it's easier to see the misalignment in (2).

webapp/src/components/post_type/component.tsx
cpoile: I can't seem to get the playwright tests to run (they keep stalling), what is the command you use?

streamer45: That would be `npx playwright test` from within the `e2e` directory. Just bear in mind I was getting some failures earlier on which are unrelated to this PR. I'd suggest to only test the relevant test by temporarily adding `.only`, here.:

https://github.com/mattermost/mattermost-plugin-calls/blob/f1a38f10c9b9523ab3150bb9ddc773b7ac9b6129/e2e/tests/join_call.spec.ts#L54


```js
test.only('call thread', async ({page}) => {  ...
```

webapp/src/components/connected_profiles.tsx
streamer45: Yeah, with the current logic I don't think we can easily cover both cases. I agree that the vertical misalignment is far more noticeable than an extra margin.

Of course I am still hoping that there's an easier (and proper) way to do this kind of positioning and I just don't know how :p

webapp/src/components/post_type/component.tsx
streamer45: Also that assumes MM is running at `http://localhost:8065` with the plugin deployed and enabled. You can pass a custom URL with the `MM_SITE_URL=` env var.

webapp/src/components/connected_profiles.tsx
streamer45: Brilliant, thanks @cpoile :tada:

PR #32, opened 2022-03-17
[MM-42590] Allow sending audio track along with screensharing

streamer45:
#### Summary

PR adds support for sending an audio track when screensharing. Turned out to be slightly more involved than anticipated due to this new track potentially conflicting with pre-existing voice tracks. We are now making the client send track ids so we can be certain of what's coming server side which makes it easy to handle everything properly.

This feature is only active if experimental features are enabled client-side. It also comes with some browser/platform limitations:

- Sharing audio only works on Chrome based browsers (no Firefox, no Safari) and only if sharing a tab, in which case only the audio from that tab is shared.
- Global system audio only works in Windows when sharing the whole screen. This applies to the Desktop App as well.

More details about audio support when screensharing can be found at https://caniuse.com/mdn-api_mediadevices_getdisplaymedia_audio_capture_support

@cpoile I haven't tested yet how this affects mobile. I am thinking it should work out of the box given we are just sending an additional audio track but please let me know if you encounter issues on that side.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-42590

### DIFF
--- a/server/channel_state.go
+++ b/server/channel_state.go
@@ -17,12 +17,14 @@ type callStats struct {
 }
 
 type callState struct {
-	ID              string                `json:"id"`
-	StartAt         int64                 `json:"create_at"`
-	Users           map[string]*userState `json:"users,omitempty"`
-	ThreadID        string                `json:"thread_id"`
-	ScreenSharingID string                `json:"screen_sharing_id"`
-	Stats           callStats             `json:"stats"`
+	ID                 string                `json:"id"`
+	StartAt            int64                 `json:"create_at"`
+	Users              map[string]*userState `json:"users,omitempty"`
+	ThreadID           string                `json:"thread_id"`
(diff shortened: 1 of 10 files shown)


cpoile (approved):
Awesome! It does work on mobile, great work!
It's a little finnicky -- sometimes I get into a state where I can't get the audio to work (on iPhone and on Android) but a restart fixes it. I think we'll have to keep an eye on it and then do a real thorough bug bash once the currently known bugs get fixed.

streamer45:
> Awesome! It does work on mobile, great work! It's a little finnicky -- sometimes I get into a state where I can't get the audio to work (on iPhone and on Android) but a restart fixes it. I think we'll have to keep an eye on it and then do a real thorough bug bash once the currently known bugs get fixed.

Agreed, this is totally experimental after all.

PR #33, opened 2022-03-17
Fix setting pop-out window title

streamer45:
#### Summary

Small bug: we were also setting the parent window's title due to the component being loaded there as well. Added a check to avoid this from happening and updated the e2e test.

### DIFF
--- a/e2e/tests/popout.spec.ts
+++ b/e2e/tests/popout.spec.ts
@@ -25,7 +25,7 @@ test.describe('popout window', () => {
         await devPage.leaveCall();
     });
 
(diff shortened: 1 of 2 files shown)


e2e/tests/popout.spec.ts
```diff
@@ -25,7 +25,7 @@ test.describe('popout window', () => {
         await devPage.leaveCall();
     });
 
-    test('window title matches', async ({page, context}) => {
+    test.only('window title matches', async ({page, context}) => {
```
streamer45, line 28:
```suggestion
    test('window title matches', async ({page, context}) => {
```

PR #34, opened 2022-03-17
[MM-42497] Implement audio ouputs selection

streamer45:
#### Summary

PR implements a new menu to select audio output devices. 

Again this is not something universally supported as the underlying API call [`HTMLMediaElement.setSinkId`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/setSinkId) is considered experimental.

- Chrome based browsers will support it natively, which also means Desktop App is covered
- On Firefox it needs to be enabled through a flag, more info at https://caniuse.com/mdn-api_htmlmediaelement_setsinkid
- No support on Safari

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-42497

#### Screenshot

image

### DIFF
--- a/e2e/tests/start_call.spec.ts
+++ b/e2e/tests/start_call.spec.ts
@@ -157,10 +157,10 @@ test.describe('setting audio input device', () => {
         await devPage.startCall();
         await devPage.wait(1000);
 
-        const currentAudioDevice = await page.evaluate(() => {
-            return window.callsClient.currentAudioDevice?.deviceId;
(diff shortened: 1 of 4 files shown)


cpoile (approved):
This is amazing, how did you do it so quickly?
I tested it locally, it works as expected, nicely done!

streamer45:
> This is amazing, how did you do it so quickly? I tested it locally, it works as expected, nicely done!

Mostly because the UI was already covered by the input case so it was just about reusing that and the actual changes in logic were limited. Glad it's working nicely :)

PR #35, opened 2022-03-21
Implement recording view

streamer45:
#### Summary

PR adds a custom recording view at `/com.mattermost.calls/recording/CHANNEL_ID`.
This is literally hacked from the existing expanded/pop out view, just removing all unnecessary parts and making the screen player as big as possible. Looking to improve further of course but I am thinking this should be good enough to enable recordings on upcoming meetings. Of course this is Community only, hence looking to merge directly on that branch.

image
New Project (1)

### DIFF
--- a/webapp/src/components/recording_view/component.tsx
+++ b/webapp/src/components/recording_view/component.tsx
@@ -1,0 +1,300 @@
+import React, {CSSProperties} from 'react';
+
+import moment from 'moment-timezone';
+
+import {Channel} from 'mattermost-redux/types/channels';
+import {UserProfile} from 'mattermost-redux/types/users';
+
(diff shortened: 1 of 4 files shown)


cpoile (approved):
I'm assuming this is a temporary hack, right? In that case it looks good to me! There are some style nits and I would probably do the component differently if it were going out publicly, but seems good for now. Nice!

streamer45:
> I'm assuming this is a temporary hack, right? In that case it looks good to me! There are some style nits and I would probably do the component differently if it were going out publicly, but seems good for now. Nice!

Yes this is totally a hack, and it's not going into the official build any time soon. I am mostly trying to merge it now to get some recordings going which will surely generate some feedback from viewers which can be addressed in a future iteration with the proper fixes as well.

PR #36, opened 2022-03-22
[MM-42498] Disable STUN if ICE host override is set

streamer45:
#### Summary

A few more beta customers are hitting this as server side STUN requires an additional outgoing UDP port to be open.
PR disables server side STUN requests if the ICE host override is set. It also allows to set empty `ICEServers` which means disabling it on client side as well.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-42498

### DIFF
--- a/server/activate.go
+++ b/server/activate.go
@@ -78,10 +78,13 @@ func (p *Plugin) OnActivate() error {
 		return err
 	}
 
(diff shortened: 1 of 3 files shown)


cpoile (approved):
👍

PR #38, opened 2022-03-24
[MM-42737] Use stream ids instead of track ids

streamer45:
#### Summary

 After some debugging (thanks @cpoile :raised_hands: ) and [knowledge gathering](https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/addTrack) I figured out that relying on local track ids was a **dangerous** move as they are not passed consistently. Stream IDs instead seem to be the way to go, so going with those this time around. 

This solution does also remove the annoying browser specific check. The moral of the story is: I should have listened to Firefox when it told me something was clearly wrong with my intended logic. 

Oh, and this is mostly why R&D call failed so emphatically.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-42737

### DIFF
--- a/server/channel_state.go
+++ b/server/channel_state.go
@@ -17,14 +17,13 @@ type callStats struct {
 }
 
 type callState struct {
-	ID                 string                `json:"id"`
-	StartAt            int64                 `json:"create_at"`
-	Users              map[string]*userState `json:"users,omitempty"`
-	ThreadID           string                `json:"thread_id"`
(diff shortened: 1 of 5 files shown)


streamer45:
> Awesome! Do you think this will solve the widget vs popout problem we were seeing?

Yes, as far as I could see it wasn't related at all to where the sharing was initiated. I was able to reproduce locally by unmuting first and then starting to share in a call where someone else shared already.

PR #39, opened 2022-04-04
[MM-42574] Single handler mode

streamer45:
#### Summary

PR implements a special single handler mode that allows calls to be hosted on a single instance in a HA cluster through the use of the `CALLS_IS_HANDLER=true` environment variable. This is how our community installation has been operating so far. The only relevant addition from the community branch is the extra expiry check which is needed to avoid potential issues with the state getting messed up due to the handler node crashing without cleaning up.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-42574

### DIFF
--- a/server/configuration.go
+++ b/server/configuration.go
@@ -50,6 +50,9 @@ func (is *ICEServers) UnmarshalJSON(data []byte) error {
 	if err != nil {
 		return err
 	}
+	if unquoted == "" {
+		return nil
+	}
 	*is = strings.Split(strings.TrimSpace(unquoted), ",")
 	return nil
 }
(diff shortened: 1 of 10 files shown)


server/configuration.go
```diff
@@ -50,6 +50,9 @@ func (is *ICEServers) UnmarshalJSON(data []byte) error {
 	if err != nil {
 		return err
 	}
+	if unquoted == "" {
+		return nil
+	}
```
streamer45, line 55: This fixes an issue when no STUN servers are passed. Instead of an empty slice of strings we were getting a slice with a single empty string which was making connection fail.

cpoile (approved):
👍

PR #40, opened 2022-04-06
Use shared rtc.Server

streamer45:
#### Summary

Removing all the WebRTC logic and instead start using the shared `rtc` package from `rtcd`. This way we can keep everything in sync between the two modes of operation (rtc included vs external rtcd service).

This doesn't yet address the second case, but only makes the plugin work as it did up until now.


server/performance/metrics.go
```diff
@@ -81,53 +80,35 @@ func NewMetrics() *Metrics {
...
+	return m.rtcMetrics
+}
+
 func (m *Metrics) Handler() http.Handler {
 	return promhttp.HandlerFor(m.registry, promhttp.HandlerOpts{})
 }
+
+func (m *Metrics) IncWebSocketEvent(direction, evType string) {
+	m.WebSocketEventCounters.With(prometheus.Labels{"direction": direction, "type": evType}).Inc()
+}
```
cpoile, line 98: Thought: could most of these be direct calls to an initialized struct from the rtcd package?

cpoile (approved):
Awesome! 👍

server/performance/metrics.go
streamer45: Not sure I understood the concern but we are passing the registry during initialization:

https://github.com/mattermost/mattermost-plugin-calls/blob/4b570fa53f800c0c0d2841f44ce5883d7cf01556/server/performance/metrics.go#L83

So what's really happening is that we are just extending the existing metrics with the rtc related counters. Anything on this side is called on the `Metrics` structure defined by the plugin.

streamer45:
@cpoile @oh6hay 

I am going to merge this as it's just becoming a bit of a blocker for other PRs that need to build on top of these changes.
Happy to address any further change in future PRs of course.

PR #41, opened 2022-04-12
Implement using external `rtcd` service

streamer45:
#### Summary

PR adds the necessary logic to enable running calls through a provided `rtcd` service.
A new `RTCDServiceURL` configuration field is available of the form:

```
https://clientID:authKey@rtcd.example.com
```

When set, the plugin will offload all calls to the external service.

Note that the authentication/authorization flow is still in under review by security so it's very likely subject to breaking changes.

### DIFF
--- a/server/activate.go
+++ b/server/activate.go
@@ -1,10 +1,13 @@
 package main
 
 import (
+	"fmt"
+	"net/url"
 	"os"
 	"time"
 
 	"github.com/mattermost/rtcd/logger"
+	rtcd "github.com/mattermost/rtcd/service"
 	"github.com/mattermost/rtcd/service/rtc"
 
 	"github.com/mattermost/mattermost-server/v6/model"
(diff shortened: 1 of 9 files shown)


server/websocket.go
```diff
@@ -533,7 +607,7 @@ func (p *Plugin) WebSocketMessageHasBeenPosted(connID, userID string, req *model
 	}
 
 	p.mut.RLock()
-	us := p.sessions[userID]
+	us := p.sessions[connID]
```
streamer45, line 610: This is a somewhat notable change. We are moving away from mapping call sessions to user ids but instead start using the more granular connection ids. This will let us support multiple, simultaneous sessions per user (e.g. both web and mobile clients connected)

server/activate.go
```diff
@@ -27,78 +30,113 @@ func (p *Plugin) OnActivate() error {
...
+
+	if cfg.RTCDServiceURL != "" {
+		u, err := url.Parse(cfg.RTCDServiceURL)
+		if err != nil {
+			p.LogError(err.Error())
+			return err
+		}
+
+		clientID := u.User.Username()
+		authKey, _ := u.User.Password()
```
cpoile, line 47: I was going to ask about validation, but then I saw it below. 👍

server/activate.go
```diff
@@ -27,78 +30,113 @@ func (p *Plugin) OnActivate() error {
...
-			ticker := time.NewTicker(handlerKeyCheckInterval)
-			defer ticker.Stop()
-			for {
-				select {
-				case <-ticker.C:
-					if err := p.setHandlerID(status.ClusterId); err != nil {
-						p.LogError(err.Error())
+		}()
+	} else {
+		if os.Getenv("CALLS_IS_HANDLER") != "" {
```
cpoile, line 72: I wonder if there's a way to simplify this nested if-else-if (and the rest of the function below)? I only ask because it's tough for me to tell what's being executed when only looking at the github review interface, which is usually a sign that things have gotten too nested.

cpoile: Hmm.. You could do the go solution of having a `Server` interface and then an `rtcdServer` and a `callsServer` implementation? Although that might cause too much duplication... 0/5, your call of course

cpoile (approved):
Awesome! A suggestion, but not blocking (up to you how you want to handle it). Thanks Claudio!

server/activate.go
streamer45: This needs some good thought cause interfaces would be an added layer of indirection which may just spread things further without actually aiding in understanding what's really going on. At some point we still need to make the distinction between what mode the plugin is operating on and currently we support at least three:


- Hosting calls on all plugin instances.
  - Single handler mode: only a single instance (cluster node) will host all the calls for the whole installation (this is how community operates today).
- Off-loading all calls to external `rtcd` service.

I am thinking of deferring any refactoring to a later point when the whole interaction with `rtcd` has been finalized.

PR #43, opened 2022-04-14
Implement self-registration to rtcd

streamer45:
#### Summary

PR implements the self-registration mechanism on plugin activation. The process works as follow:

Precedence is given to these environment variables which override anything else:

- `CALLS_RTCD_URL` - URL to a running `rtcd` service.
- `CALLS_RTCD_CLIENT_ID` -  unique client identifier (e.g. installation id).
- `CALLS_RTCD_AUTH_KEY` - authentication key for the client.

If these are not set we then look at the plugin's config by parsing the URL (it can be in the form `http://clientID:authKey@host`)

If no auth key is provided either we then try to fetch the `rtcd` client config from the k/v store. If missing there as well we finally attempt to register the client and store the returned auth key in the k/v store. We try to register also in case the provided client id or URL don't match what's saved in the k/v store.

@gabrieljackson Please let me know if this is enough for your side to work with. Happy to make changes as needed.


server/rtcd.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+
+	cfgData, err := json.Marshal(&cfg)
+	if err != nil {
+		return cfg, fmt.Errorf("failed to marshal rtcd client config: %w", err)
+	}
+
+	if err := p.API.KVSet(rtcdConfigKey, cfgData); err != nil {
+		return cfg, fmt.Errorf("failed to store rtcd client config: %w", err)
+	}
```
streamer45, line 110: To note, a limitation of the current design, if the application were to crash after the registration succeeded but before we were able to store the config (containing the authkey) the client would be permanently locked out until either the clientID changed or the client was deregistered, just like a user forgetting its own password.

cpoile: Maybe send an `ack` that signals the registration is complete and the authkey can be activated?

cpoile (approved):
Awesome! Just one non-blocking suggestion to the problem.

server/rtcd.go
streamer45: If we go to that complexity (having to keep additional state and coordination) I am thinking we may as well let the client provide the auth key, At the end of the day it's just like a password. We could still validate it properly on server side to make sure it has enough entropy.

cpoile: Since there's no real authentication happening here, maybe we should just do that? (have the client state their auth key, have the server record it) I'm not really sure what this is doing, security-wise, to be honest. You kind of brought it up in the call, but I wasn't totally convinced by the answer.

streamer45: I guess the advantage was to keep things simple having a central place that generates these keys. Now we are trusting clients to provide them which means we need to add some validation in place. But from a security perspective I don't think it's any different if implemented properly. I can bring this up today so we can converge on a decision asap.

streamer45: @oh6hay Wondering if you could weight on this from a security perspective. Is there any advantage in letting the server (rtcd) generate the authentication key (current implementation) vs having the client (plugin) proposing one, assuming some checks are in place to make sure the key is "strong" enough?

streamer45:
@cpoile I am going to merge this one so we only keep one branch active. Any further change can be made over there.

PR #44, opened 2022-04-19
load-test client

streamer45:
#### Summary

Looking to merge the load-test client into the main branch. Notably I've implemented a couple of additions which will be needed in the upcoming performance tests:

- Concurrent calls support
- Screen sharing support

There's some work still needed to polish (e.g. proper config) it but I feel it's a good time to include it, if anything to avoid constantly switching between branches.


lt/client.go
```diff
@@ -0,0 +1,639 @@
...
+		}
+		defer file.Close()
+
+		// Open on oggfile in non-checksum mode.
+		ogg, _, oggErr := oggreader.NewWith(file)
+		if oggErr != nil {
+			log.Fatalf(oggErr.Error())
+		}
+
+		<-time.After(2 * time.Second)
```
streamer45, line 185: @cpoile Could you remind me why we have to sleep for a bit here? I've kept the semantic you introduced but forgot what we were trying to achieve (or avoid that is).

cpoile (approved):
👍

streamer45:
I had to revert one small detail as setting up the audio track is actually required even if clients are muted.

streamer45:
@cpoile Will finally merge this so that we can update the documentation and link to the load-test docs as I've already heard a couple of customers in need for something to test Calls performance on their instances.

This should also unblock any further improvement such as the simulative logic you were working on.

PR #46, opened 2022-04-20
High concurrency fixes

streamer45:
#### Summary

Two main additions:

- Added some mitigation to a state consistency issue caused by replica lag and high concurrency. The result was that some calls were still running due to sessions not being removed correctly. 
- Updating the server deps to start using the `ReliableClusterSend` introduced in [. This should](https://github.com/mattermost/mattermost-server/pull/19594).


go.mod
```diff
@@ -4,11 +4,11 @@ go 1.17
 
 require (
 	github.com/gorilla/websocket v1.5.0 // indirect
-	github.com/mattermost/mattermost-server/v6 v6.4.0
+	github.com/mattermost/mattermost-server/v6 v6.6.0
```
cpoile, line 7: Will this require us to bump the minimum server?

server/session.go
```diff
@@ -110,7 +118,20 @@ func (p *Plugin) removeUserSession(userID, channelID string) (channelState, chan
 
 		currState = *state
 		return state, nil
-	})
+	}
+
+	var err error
+	maxTries := 5
+	for i := 0; i < maxTries; i++ {
+		err = p.kvSetAtomicChannelState(channelID, setChannelState)
```
cpoile, line 126: I wonder if we should start to use the `mattermost-plugin-api`? It has `SetAtomicWithRetries` and a few other convenience functions that are well tested.  (I can't remember if I've mentioned that already, apologies if I have)

cpoile (approved):
Cool! Non-blocking question and suggestion.

go.mod
streamer45: Nope, changes are backwards compatible so while the reliable send will only work on 6.5 onwards previous versions shouldn't be affected.

server/session.go
streamer45: Maybe. I am thinking we could probably use that in place of the `kvSetAtomic` util but this is different and would require a customized version because we need the attempts to continue under certain conditions that are implementation specific (in this case the session being there as we expect it to be).

PR #51, opened 2022-04-27
[MM-42712] Implement basic rate limiting for incoming WebSocket messages

streamer45:
#### Summary

I've tested this locally and while this is surely better than nothing, it feels like a proper solution should be either implemented on the server side or we should find a way to allow a plugin to disconnect a malicious connection. As things stand a high rate of messages will still produce some noticeable load given the events are routed to the plugin with RPC calls happening and all that's involved into that process.

Please let me know your thoughts.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-42712

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -17,6 +17,7 @@ require (
 	github.com/mattermost/rtcd v0.3.1-0.20220427165117-035048fda047
 	github.com/pion/stun v0.3.5
 	github.com/rudderlabs/analytics-go v3.3.2+incompatible
+	golang.org/x/time v0.0.0-20201208040808-7e3f01d25324
 )
 
 require (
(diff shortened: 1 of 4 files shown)


streamer45:
> I suppose a proper server side solution would be better after we move into the sever? Seems fine for now. :)

Yes, we could work something with Server Platform team and find a way to extend the existing rate limiter to address WebSocket events as well but the limit may be app/plugin specific so best solution is to keep the check there I feel. 

Exposing a way to trigger a disconnect would be an interesting approach but again it would require some coordination and likely bumping minimum server version as well. Within the current timeline I think this as a patch but happy to freeze it as well if it's not deemed worth it from a security perspective.

oh6hay (approved):
:+1:

PR #52, opened 2022-04-29
MM-43131: Add ability to add custom start message

agnivade:
https://mattermost.atlassian.net/browse/MM-43121


webapp/src/components/post_type/component.tsx
```diff
@@ -49,59 +49,62 @@ const PostType = ({post, connectedID, hasCall, pictures, profiles, showSwitchCal
...
-            <SubMain ended={Boolean(post.props.end_at)}>
-                <Left>
-                    <CallIndicator ended={Boolean(post.props.end_at)}>
-                        {!post.props.end_at &&
-                            <ActiveCallIcon
-                                fill='var(--center-channel-bg)'
-                                style={{width: '100%', height: '100%'}}
-                            />
+        <>
+            <Message>{post.message}</Message>
```
agnivade, line 53: This may need to by styled a bit better. Please tell me some magical css incantations :)

webapp/src/index.tsx
```diff
@@ -268,8 +268,13 @@ export default class Plugin {
 
             switch (subCmd) {
             case 'join':
```
agnivade, line 270: Need some guidance here to limit `/call join` to only join calls and not start them.

agnivade (commented):
Brace yourselves!

agnivade:
And writing tests isn't mandatory for OSF right? :sweat_smile:

webapp/src/index.tsx
streamer45: I think it's fine to keep the existing behaviour here. If anything I'd be making `/start` fail if the call is already ongoing.

To do that I think we need a new redux selector though, more or less a copy of [`voiceConnectedUsers`](https://github.com/mattermost/mattermost-plugin-calls/blob/f252b1e7a0216fb6f4320034cfca03c1ac47ecf7/webapp/src/selectors.ts#L12-L19) but that receives a channel id given it may not always be the current channel in this case. Then if the result has length > 0 we error out since it means the call is happening already.

webapp/src/index.tsx
```diff
@@ -378,7 +383,7 @@ export default class Plugin {
                 return;
             }
             if (ev.data && ev.data.type === 'connectCall') {
-                connectCall(ev.data.channelID);
+                connectCall(ev.data.channelID, '');
```
streamer45, line 386: This is JS we can omit the second parameter entirely :)

webapp/src/index.tsx
```diff
@@ -341,7 +346,7 @@ export default class Plugin {
         };
 
         let initializing = false;
-        const connectCall = async (channelID: string) => {
+        const connectCall = async (channelID: string, title: string) => {
```
streamer45, line 349: Let's make the title optional with `title?: string`

server/plugin.go
```diff
@@ -223,11 +223,12 @@ func (p *Plugin) startNewCallThread(userID, channelID string, startAt int64) (st
 	post := &model.Post{
 		UserId:    userID,
 		ChannelId: channelID,
-		Message:   postMsg,
+		Message:   title,
```
streamer45, line 226: Wondering if we should use this only if non-empty.

server/websocket.go
```diff
@@ -579,8 +579,13 @@ func (p *Plugin) WebSocketMessageHasBeenPosted(connID, userID string, req *model
 			p.LogError("missing channelID")
 			return
 		}
+		title, ok := req.Data["title"].(string)
+		if !ok {
+			p.LogError("missing title")
+			return
+		}
```
streamer45, line 586: Does this make it a requirement for a title to be there? I think we should still continue if it's missing.

webapp/src/client.ts
```diff
@@ -67,7 +67,7 @@ export default class CallsClient extends EventEmitter {
         };
     }
 
-    public async init(channelID: string) {
+    public async init(channelID: string, title: string) {
```
streamer45, line 70: Same thing, let's make the parameter optional with `title?: string`.

streamer45 (commented):
This is awesome :raised_hands: thank you!

server/websocket.go
streamer45: I see, we are sending an empty string either way. I guess we could make it completely optional to avoid even sending the field but it's okay really.

webapp/src/index.tsx
streamer45: We need to pass `args.channel_id` to the selector. That should fix it.

server/websocket.go
streamer45: I think this is now preventing us from starting a call without a message given we made it an optional parameter on the client. I think it's safe to remove the check completely given this is optional. Otherwise only check if the `title` field exists and it's empty at the same time, but again that may require client changes.

webapp/src/index.tsx
agnivade: Done

server/websocket.go
agnivade: Done

cpoile (commented):
Thanks @agnivade, looks great!

As for the message styling, can @abhijit-singh take a quick look? This is what it looks like when starting a call now:
image
image

Here it is with the call started with: `/call start This is the title of the call thread`
image
image

What do you think @abhijit-singh ?

abhijit-singh:
Ooh, this is great! Thanks for looping me in @cpoile. I think the title is getting lost in that message in the current state and could be given a little more prominence. 

We probably don't want to spend too much time doing something fancy, so I tried recreating it keeping the same layout, and using h3 (###) for the title text works for me. Would look something like this - 
image

agnivade:
Thanks Abhijit. This is easy enough CSS for me! Let me make the changes.

agnivade:
@cpoile @streamer45 - PTAL

webapp/src/components/post_type/component.tsx
```diff
@@ -49,59 +49,62 @@ const PostType = ({post, connectedID, hasCall, pictures, profiles, showSwitchCal
...
-            <SubMain ended={Boolean(post.props.end_at)}>
-                <Left>
-                    <CallIndicator ended={Boolean(post.props.end_at)}>
-                        {!post.props.end_at &&
-                            <ActiveCallIcon
-                                fill='var(--center-channel-bg)'
-                                style={{width: '100%', height: '100%'}}
-                            />
+        <>
+            <h3 className='markdown__heading'>{post.message}</h3>
```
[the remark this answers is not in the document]
agnivade: Would this be solved if we just move the title to a prop? Then post.message would still be under <Message>.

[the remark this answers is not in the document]
streamer45: As long as we add code to display the new prop it should work just fine.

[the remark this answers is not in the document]
cpoile: I agree with removing the `Call ended` message, because we would still want the thread title. 👍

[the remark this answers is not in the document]
agnivade: Ohh my bad, you want Call ended to be replaced by the title. Yeah that makes sense. Sorry I thought you were asking to revert the style of Call ended to the original style.

agnivade:
Bear with me on this one. I'll get to it as soon as I get free time.

webapp/src/components/post_type/component.tsx
```diff
@@ -48,60 +48,65 @@ const PostType = ({post, connectedID, hasCall, pictures, profiles, showSwitchCal
         <Duration>{moment(post.props.start_at).fromNow()}</Duration>
     );
 
+    const title = post.props.title ? (<h3 className='markdown__heading'>{post.props.title}</h3>): (<div></div>);
```
agnivade, line 51: Let me know if there's a better to way signify an empty variable other than `<div></div>`. :smiling_face_with_tear:

agnivade:
@streamer45 @cpoile - PTAL.

webapp/src/components/post_type/component.tsx
streamer45: I think we can use [inline conditional rendering](https://reactjs.org/docs/conditional-rendering.html#inline-if-with-logical--operator) and only add the case we are interested in. So no need to initialize a variable but simply use this on top of the render return body:

```tsx
            {post.props.title &&
                <h3 className='markdown__heading'>{post.props.title}</h3>
            }
```

agnivade: Nice. Learnt so much!

streamer45 (approved):
Looks good to me, brilliant effort :bowing_man: :100:

cpoile (approved):
:) Thanks @agnivade, this is great!

webapp/src/components/post_type/component.tsx
```diff
@@ -49,59 +49,64 @@ const PostType = ({post, connectedID, hasCall, pictures, profiles, showSwitchCal
...
-                <Left>
-                    <CallIndicator ended={Boolean(post.props.end_at)}>
-                        {!post.props.end_at &&
-                            <ActiveCallIcon
-                                fill='var(--center-channel-bg)'
-                                style={{width: '100%', height: '100%'}}
-                            />
+        <>
+            {post.props.title &&
+                <h3 className='markdown__heading'>{post.props.title}</h3>
```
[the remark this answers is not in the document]
abhijit-singh: @ mentions do work if we create an H3 using markdown. If it's not possible or complicated to make it work in an H3 title, I'd be supportive of removing h3 formatting to get @ mentions in.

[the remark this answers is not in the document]
itao: (Of course, this isn't a priority for v7.0 btw)

> overriding the usual post rendering

Ah I assumed we're already using the regular post rendering, just with a slack attachment that says call started. I think clicking on @ mention (and # hashtag to search) will be valuable especially as group mention gains the functionality to view group members on click.

I tried the `/echo` command and it supports all the markdown formatting (multi-line, hyperlink, @ mention etc). is tricky for `/call start` to do it the same way but with an attachmetn?

PR #55, opened 2022-04-29
[MM-43887] Attempt to re-register client if first connect attempt fails

streamer45:
#### Summary

Turned out that persisting `rtcd`'s auth store in k8s is a bit more complex than anticipated so we are allowing (tolerating really) it to reset on each pod restart. This means we need to support re-registering the client (plugin) which is what this PR is about.

This isn't ideal but hopefully it won't be needed for long as we are looking into moving to a decoupled auth service at some point.

/cc @angeloskyratzakos 

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-43887

### DIFF
--- a/server/rtcd.go
+++ b/server/rtcd.go
@@ -30,6 +30,85 @@ func (p *Plugin) getStoredRTCDConfig() (rtcd.ClientConfig, error) {
 	return cfg, nil
 }
 
+func (p *Plugin) registerRTCDClient(cfg rtcd.ClientConfig) (rtcd.ClientConfig, error) {
+	client, err := rtcd.NewClient(cfg)
+	if err != nil {
+		return cfg, fmt.Errorf("failed to create rtcd client: %w", err)
+	}
+	defer client.Close()
+
+	cfg.AuthKey, err = client.Register(cfg.ClientID)
+	if err != nil {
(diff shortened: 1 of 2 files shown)


server/rtcd.go
```diff
@@ -30,6 +30,85 @@ func (p *Plugin) getStoredRTCDConfig() (rtcd.ClientConfig, error) {
...
+	}
+
+	err = client.Connect()
+	if err == nil {
+		return client, nil
+	}
+	defer client.Close()
+
+	// If connecting fails we attempt to re-register once as the rtcd instance may
+	// have restarted, potentially losing stored credentials.
```
cpoile, line 77: Should we make a longer pause then, maybe wait long enough for an rtcd instance to restart? (Is that even a worry?)

cpoile (approved):
👍

server/rtcd.go
streamer45: The assumption here is that `rtcd` restarted before plugin activation and we'd be failing to authenticate because the client would be effectively unregistered again at that point. The case of `rtcd` restarting while the plugin is connected will be handled in https://mattermost.atlassian.net/browse/MM-43508.

PR #56, opened 2022-05-02
[MM-43789] Simplify client initialization

streamer45:
#### Summary

PR simplifies client initialization by decoupling object creation and call to the `init()` method.
This change was done already in the [community branch](https://github.com/mattermost/mattermost-plugin-calls/blob/7a6b5645f44bd6c636857b1436e4258ff35e17fb/webapp/src/index.tsx#L356) but I forgot to add the early return check which I think allowed for this bug to happen.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-43789

### DIFF
--- a/webapp/src/index.tsx
+++ b/webapp/src/index.tsx
@@ -340,19 +340,14 @@ export default class Plugin {
             );
         };
 
-        let initializing = false;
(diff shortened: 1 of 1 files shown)


cpoile (approved):
Awesome!

PR #57, opened 2022-05-02
[MM-43660] Expose some client-side metrics through `/call stats` command

streamer45:
#### Summary

Adding a new utility command `/call stats` to expose some client-side metrics.
Command will generate an ephemeral message with the collected metrics in JSON, e.g.:

image

This will be useful to debug possible client-side conditions (e.g. networking) affecting calls quality.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-43660

### DIFF
--- a/webapp/src/index.tsx
+++ b/webapp/src/index.tsx
@@ -9,6 +9,8 @@ import {getMyChannelMemberships} from 'mattermost-redux/selectors/entities/commo
 import {getChannel as getChannelAction} from 'mattermost-redux/actions/channels';
 import {getProfilesByIds as getProfilesByIdsAction} from 'mattermost-redux/actions/users';
 
+import {RTCStats} from 'src/types/types';
+
 import {isVoiceEnabled, connectedChannelID, voiceConnectedUsers, voiceChannelCallStartAt} from './selectors';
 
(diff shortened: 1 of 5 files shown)


webapp/src/index.tsx
```diff
@@ -292,6 +294,18 @@ export default class Plugin {
...
                 }
+                break;
+            case 'stats':
+                if (!window.callsClient) {
+                    return {error: {message: 'You are not connected to any call'}};
+                }
+                try {
+                    const stats = await window.callsClient.getStats();
+                    console.log(JSON.stringify(stats, null, 2));
+                    return {message: `/call stats "${JSON.stringify(stats)}"`, args};
```
cpoile, line 305: Heh, this is clever. I never would have thought to do this. Makes it simpler to display to the user. 👍

cpoile (approved):
Nice!

webapp/src/index.tsx
streamer45: It's horribly clever :see_no_evil: . To be honest I wanted a way to display an ephemeral message directly from the client but to my surprise it doesn't look like possible. The solution is hacky but it gets the job done so why not.

PR #58, opened 2022-05-03
MM042710 - Implement dependency pinning

cpoile:
#### Summary
- Ran an npm audit, there's still one high priority transitive dependency from webapp, so we don't have control over that.
- Using `npm ci` for all build steps
- Using pinned docker images
- Thanks @oh6hay!
- Update rtcd dependency

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-42710


oh6hay (approved):
:+1:

Makefile
```diff
@@ -125,7 +125,7 @@ endif
 webapp/node_modules: $(wildcard webapp/package.json)
 ifneq ($(HAS_WEBAPP),)
 	cd webapp && node skip_integrity_check.js
-	cd webapp && $(NPM) install
+	cd webapp && $(NPM) ci
```
streamer45, line 128: Is there any side effect of running `npm ci` locally? For example I am worried it could force a re-install of all modules every time. 

I guess the better question is: do we actually call this Makefile target from CI? If not, we may leave it unchanged or devise some additional check based on env variables.

streamer45 (commented):
Thanks @cpoile, left one question.

Makefile
cpoile: We do run this from the CI, but I initially thought there shouldn't be a side effect. But looking into the docs, it looks like in dev, new dependencies might not get persisted to the `package-lock.json`, so I split up the webapp build step into a `webapp-ci` version. That's cleaner too, because we are already doing the `npm ci` step in the `npm-dependencies` circleci job.  A little confusing.  Thanks for the question!

streamer45 (approved):
Looks great, thanks @cpoile 👍

streamer45:
> @streamer45 Not sure if you want to merge this before the next cut. Your call!

Let's include it, I am thinking it's unlikely to break things.

PR #59, opened 2022-05-04
[MM-43508] Implement basic reconnection support for rtcd client

streamer45:
#### Summary

PR adds reconnection support for the rtcd client as implemented in https://github.com/mattermost/rtcd/pull/24

- If a disconnection occurs, we first attempt to re-register the client in case the `rtcd` instance restarted and wiped its store.
- Whether that fails or not we keep trying up to `maxReconnectAttempts` (default to 8) with some incremental backoff (2 seconds). 
- If `maxReconnectAttempts` is reached we give up and disable the plugin.

To note, this doesn't cover initial connection on plugin startup. Adding that support makes the logic even more complex so I opted to keep the assumption that a running `rtcd` should be available when enabling the plugin.

Also, this whole logic is going to change a bit when implementing [MM-43872](https://mattermost.atlassian.net/browse/MM-43872) as we are going to deal with multiple concurrent clients but didn't want to couple the two pieces of work so take this as a first iteration.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-43508


cpoile (approved):
👍

go.mod
```diff
@@ -15,7 +15,7 @@ require (
 require (
 	github.com/mattermost/logr/v2 v2.0.15
 	github.com/mattermost/mattermost-plugin-api v0.0.27
-	github.com/mattermost/rtcd v0.3.1-0.20220427165117-035048fda047
+	github.com/mattermost/rtcd v0.3.1-0.20220504082845-a64f36fa09dc
```
cpoile, line 18: Oh, this will have to be updated once it's merged (you probably know) -- I was stung once by a branch disappearing on me once. :)

streamer45: Yes, I'll probably push a new tag since there are significant changes upstream.

PR #65, opened 2022-05-09
[MM-43140] Fix call icon color in channel link label

streamer45:
#### Summary

We were still using an hard-coded white value instead of following the theme.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-43140


cpoile (approved):
👍

PR #69, opened 2022-05-11
[MM-43872] Implement rtcd auto-scaling logic

streamer45:
#### Summary

This is where *the magic* happens :magic_wand:  It's still a bit rough with a couple of corner cases but it's a start. Looking to improve this in further iterations.

Simply put the logic goes as follows:

- Upon starting the plugin we create an `rtcdClientManager` which resolves the `rtcdURL` and connects to any returned host.
- We also start a `hostsChecker` goroutine to routinely resolve the hosts in case anything has changed. If new hosts are found we create clients for those as well. If some hosts are gone we flag them so that they won't be used for future calls.
- When a call is about to start we pick the host with the least number of calls started on it (from the plugin instance perspective that is, to be improved).
- Once a call has started, any future message related to that call is routed to the appropriate host. This is saved as part of the `callState` object in the k/v store.

As I mentioned above there are still some rough edges. As an example, we don't have yet a proper way of removing hosts from the mapping other than when clients have failed all their re-connection attempts. Ideally we'd like to remove those as soon as the last call routed to them has ended. To do that we need a fair bit more coordination though so opted to defer it for now and think it through a bit more.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-43872


server/rtcd.go
```diff
@@ -177,27 +438,94 @@ func (p *Plugin) getRTCDClientConfig(rtcdURL string) (rtcd.ClientConfig, error)
...
 	}
 
-	if cfg, err := p.registerRTCDClient(cfg); err != nil {
+	if cfg.AuthKey == "" {
+		m.ctx.LogDebug("auth key missing from rtcd config, generating a new one")
+		cfg.AuthKey, err = random.NewSecureString(32)
+		if err != nil {
+			return cfg, fmt.Errorf("failed to generate auth key: %w", err)
+		}
+	}
```
streamer45, line 465: Forgot to mention that this PR also includes the auth related changes proposed in https://github.com/mattermost/rtcd/pull/28 . This simplifies a bit and removes the corner case of getting locked out due to crashing after registration but before saving to the store. We now save the key prior to registration in `registerRTCDClient`.

server/rtcd.go
```diff
@@ -4,105 +4,351 @@ import (
...
 
-func (p *Plugin) registerRTCDClient(cfg rtcd.ClientConfig) (rtcd.ClientConfig, error) {
-	client, err := rtcd.NewClient(cfg)
-	if err != nil {
-		return cfg, fmt.Errorf("failed to create rtcd client: %w", err)
+// hostsChecker runs in a dedicated goroutine that routinely resolves all
+// the available hosts (ip addresses) pointed by the rtcd URL that are advertised through DNS.
+// When new hosts are found a client for them is created. Hosts that are missing
+// from the returned set are flagged and won't be used for new calls.
+func (m *rtcdClientManager) hostsChecker() {
```
cpoile, line 120: also cool!

cpoile (approved):
Awesome work, that's a pretty deep PR. 😅

server/rtcd.go
```diff
@@ -4,105 +4,351 @@ import (
...
+	h.callsCounter++
+	h.mut.Unlock()
+
+	return h.ip, nil
+}
+
+// Send routes the message to the appropriate host that's handling the given
+// call. If this is missing a new client is created and added to the mapping.
+func (m *rtcdClientManager) Send(msg rtcd.ClientMessage, callID string) error {
+	state, err := m.ctx.kvGetChannelState(callID)
```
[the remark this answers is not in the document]
streamer45: My first implementation was actually doing some in-memory caching and only fetch on join events. But with that it came more complexity as I had to keep a map of `callID -> host` and a few extra corner cases to handle so I reverted to the dumbest solution which I think is going to perform well enough for now. But certainly doing some caching would be a good first optimization.

PR #70, opened 2022-05-12
[MM-43140] Use correct color for channel link label

streamer45:
#### Summary

This is why we badly need automated e2e tests :man_facepalming: 

In https://github.com/mattermost/mattermost-plugin-calls/pull/65 I used the wrong color. Now using `sidebarText` instead which makes more sense.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-43140

### DIFF
--- a/webapp/src/components/channel_link_label/component.tsx
+++ b/webapp/src/components/channel_link_label/component.tsx
@@ -44,7 +44,7 @@ const ChannelLinkLabel = (props: Props) => {
             >
 
                 <ActiveCallIcon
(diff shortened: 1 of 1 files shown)


streamer45:
We may even go as far as using `sidebarUnreadText` but not sure. 

@abhijit-singh would you know?

image

cpoile (approved):

abhijit-singh (approved):
`sidebarText` should work fine 👍

PR #72, opened 2022-05-16
[MM-43590] Simplify client state re-initialization

streamer45:
#### Summary

PR fixes a couple of issues with client state upon plugin re-initialization due to either:

- plugin disable/enable
- server restart 

This should fix the need for people to refresh every time pods rotate in Community. I tested various scenarios and everything seems okay now.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-43590

### DIFF
--- a/webapp/src/components/channel_header_button/index.ts
+++ b/webapp/src/components/channel_header_button/index.ts
@@ -9,7 +9,7 @@ import ChannelHeaderButton from './component';
 
 const mapStateToProps = (state: GlobalState) => ({
     hasCall: voiceConnectedUsers(state).length > 0,
-    inCall: connectedChannelID(state) && connectedChannelID(state) === getCurrentChannelId(state),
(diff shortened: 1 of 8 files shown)


webapp/src/components/channel_header_button/index.ts
```diff
@@ -9,7 +9,7 @@ import ChannelHeaderButton from './component';
 
 const mapStateToProps = (state: GlobalState) => ({
     hasCall: voiceConnectedUsers(state).length > 0,
-    inCall: connectedChannelID(state) && connectedChannelID(state) === getCurrentChannelId(state),
+    inCall: Boolean(connectedChannelID(state) && connectedChannelID(state) === getCurrentChannelId(state)),
```
streamer45, line 12: This is unrelated but I got an error while testing so including it here.

cpoile (approved):
Awesome, thanks! 👍

cpoile (approved):
Look good to me, thanks Claudio!

PR #76, opened 2022-05-18
MM-44055 - Beta limits for Cloud

cpoile:
#### Summary
- Implements the following beta limits for cloud:
  - On-prem: no limits to calls
  - Cloud Starter: DMs 1-1 only
  - Cloud Professional & Cloud Enterprise ("Cloud Paid"): DMs 1-1, GMs and Channel calls limited to 8 people.
- I had to split this from the upsell ticket, things were getting a little complicated. 
- Had to expose the license info in our own endpoint (after talking with Maria; they didn't expect to have to release that info to non-sysadmins)

#### UI:
Cloud starter when in a GM or Channel (this will be updated in the :
image
Cloud paid when in a GM or Channel with more than 8 participants:

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-44055

### DIFF
--- a/server/cloud_limits.go
+++ b/server/cloud_limits.go
@@ -1,0 +1,51 @@
+package main
+
+import (
+	"fmt"
+	"github.com/mattermost/mattermost-server/v6/model"
+)
+
+const cloudMaxParticipants = 8
+
+// JoinAllowed returns true if the user is allowed to join the call, taking into
+// account cloud limits
+func (p *Plugin) JoinAllowed(channelID string, state *channelState) (bool, error) {
+	license := p.pluginAPI.System.GetLicense()
+
+	// Rules are:
+	// On-prem: no limits to calls
+	// Cloud Starter: DMs 1-1 only
+	// Cloud Professional & Cloud Enterprise: DMs 1-1, GMs and Channel calls limited to 8 people.
+
+	if !isCloud(license) {
+		return true, nil
+	}
+
+	channel, err := p.API.GetChannel(channelID)
+	if err != nil {
+		return false, fmt.Errorf("get channel failed: %w", err)
+	}
+
+	if isCloudStarter(license) {
+		return channel.Type == model.ChannelTypeDirect, nil
+	}
+
+	// we are cloud paid (starter or enterprise)
+	if len(state.Call.Users) >= cloudMaxParticipants {
+		return false, nil
+	}
+
+	return true, nil
+}
+
+func isCloud(license *model.License) bool {
+	if license == nil || license.Features == nil || license.Features.Cloud == nil {
+		return false
+	}
+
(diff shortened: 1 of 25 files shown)


webapp/src/selectors.ts
```diff
@@ -66,3 +70,60 @@ export const screenSourceModal = (state: GlobalState) => {
     return getPluginState(state).screenSourceModal;
 };
 
+const sku = (state: GlobalState): string => {
+    return getPluginState(state).cloudInfo.sku_short_name;
+}
+
+//
+// Selectors for Cloud and beta limits:
```
cpoile, line 78: also includes selectors that we'll need later

webapp/src/index.tsx
```diff
@@ -452,7 +466,8 @@ export default class Plugin {
                 channel = getChannel(store.getState(), channelID);
             }
 
-            const roles = getMyRoles(store.getState());
+            const systemRoles = getMySystemRoles(store.getState());
+            const channelRoles = getMyChannelRoles(store.getState());
```
cpoile, line 470: this was changed in webapp, so updated here

Phrynobatrachus (approved):
[created Zephyr case](https://mattermost.atlassian.net/projects/MM?selectedItem=com.atlassian.plugins.atlassian-connect-plugin:com.kanoah.test-manager__main-project-page#!/testCase/MM-T4812)

Testing following that LGTM.

streamer45 (commented):
Thanks @cpoile, looks very good.

My only thought would be on whether we should make the limit dynamic instead. I think at some point there were talks to allow on-prem admins to set a limit as well from config. We don't have to do it all now but for cloud purposes allowing for the value to be overridden through environment would allow us to potentially raise the limit without pushing new code. What do you think?

server/cloud_limits.go
```diff
@@ -0,0 +1,71 @@
...
+	// Cloud Professional & Cloud Enterprise: DMs 1-1, GMs and Channel calls limited to 8 people.
+
+	if !isCloud(license) {
+		return true, nil
+	}
+
+	channel, err := p.API.GetChannel(channelID)
+	if err != nil {
+		return false, fmt.Errorf("get channel failed: %w", err)
+	}
```
streamer45, line 29: Since we are fetching the channel already in `handleJoin` maybe we can pass it through `addUserSession` in place of  the channel id and avoid some redundancy.

cpoile:
@streamer45 Good idea for the env override, let me know if this is how you'd want to handle it [a17d7b3](https://github.com/mattermost/mattermost-plugin-calls/pull/76/commits/a17d7b309b6acdca23d7891f4daa0ec39990606a)

streamer45 (approved):
Looks great :+1:

PR #79, opened 2022-05-24
[MM-44370] Fix state reset during ws reconnect

streamer45:
#### Summary

Resetting state in https://github.com/mattermost/mattermost-plugin-calls/pull/72 turned out to be a bit too aggressive with various side effects like the attached issue. Now resetting the whole state only if the user has disconnected from the call, otherwise simply updating it. 

At some point we'll want to send the whole current state through the websocket channel, mainly to avoid possible race conditions that could still cause inconsistencies at this point.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-44370

### DIFF
--- a/webapp/src/index.tsx
+++ b/webapp/src/index.tsx
@@ -583,6 +583,7 @@ export default class Plugin {
             if (window.callsClient) {
                 window.callsClient.disconnect();
             }
+            console.log('calls: resetting state');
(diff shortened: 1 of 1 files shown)


cpoile (approved):
I think I need to do a bug bash on mobile and see if if any of these issues are popping up there.

PR #84, opened 2022-05-25
Consolidate config, Reduce quantity of api calls

cpoile:
#### Summary
- While working on the beta limits for mobile, I anticipated that Elias would not like us adding another API call, even if it's only on connect/reconnect. Anyway, I think we should be sending over the `/cloud-info` data in the `/config` api call, since they're both "config" type of data.
- Then I thought we may as well put the config into the store -- it's been bugging me that we make an api call on every join, and 2 api calls on every channel switch. Now we only make one api call on a channel switch (and I'd like to get rid of that too, maybe later).

#### Ticket Link
- None

### DIFF
--- a/server/api.go
+++ b/server/api.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"encoding/json"
+	"github.com/pkg/errors"
 	"io/ioutil"
 	"net/http"
 	"net/http/pprof"
@@ -286,6 +287,33 @@ func (p *Plugin) handleDebug(w http.ResponseWriter, r *http.Request) {
 	res.code = http.StatusNotFound
 }
 
+// handleConfig returns the client configuration, and cloud license information
+// that isn't exposed to clients yet on the webapp
+func (p *Plugin) handleConfig(w http.ResponseWriter) error {
+	license := p.pluginAPI.System.GetLicense()
+	if license == nil {
+		p.handleErrorWithCode(w, http.StatusBadRequest, "no license",
+			errors.New("no license found"))
+		return nil
+	}
+
+	clientConfig := p.getConfiguration().getClientConfig()
+	config := map[string]interface{}{
+		"ice_servers":            clientConfig.ICEServers,
+		"allow_enable_calls":     clientConfig.AllowEnableCalls,
+		"default_enabled":        clientConfig.DefaultEnabled,
+		"sku_short_name":         license.SkuShortName,
+		"cloud_max_participants": cloudMaxParticipants,
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	if err := json.NewEncoder(w).Encode(config); err != nil {
+		return errors.Wrap(err, "error encoding config")
+	}
+
+	return nil
+}
+
 func (p *Plugin) ServeHTTP(c *plugin.Context, w http.ResponseWriter, r *http.Request) {
 	if strings.HasPrefix(r.URL.Path, "/version") {
 		p.handleGetVersion(w, r)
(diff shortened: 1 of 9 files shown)


server/api.go
```diff
@@ -286,6 +287,33 @@ func (p *Plugin) handleDebug(w http.ResponseWriter, r *http.Request) {
...
+		p.handleErrorWithCode(w, http.StatusBadRequest, "no license",
+			errors.New("no license found"))
+		return nil
+	}
+
+	clientConfig := p.getConfiguration().getClientConfig()
+	config := map[string]interface{}{
+		"ice_servers":            clientConfig.ICEServers,
+		"allow_enable_calls":     clientConfig.AllowEnableCalls,
+		"default_enabled":        clientConfig.DefaultEnabled,
```
cpoile, line 304: Start moving over to using json-standard snake case

webapp/src/client.ts
```diff
@@ -155,17 +159,8 @@ export default class CallsClient extends EventEmitter {
...
-                config = resp.data;
-            } catch (err) {
-                console.log(err);
-                this.ws?.close();
-                return;
-            }
-
-            const iceServers = config.ICEServers?.length > 0 ? [{urls: config.ICEServers}] : [];
+            const configServers = getIceServers(this.getState());
+            const iceServers = configServers?.length > 0 ? [{urls: configServers}] : [];
```
cpoile, line 163: Is there a specific reason we want to be requesting fresh iceServers on `join`, or was that just a way to get the data? We can still request it on join if we want to.

webapp/src/client.ts
```diff
@@ -155,17 +159,8 @@ export default class CallsClient extends EventEmitter {
...
-                const resp = await axios.get(`${getPluginPath()}/config`);
-                config = resp.data;
-            } catch (err) {
-                console.log(err);
-                this.ws?.close();
-                return;
-            }
-
-            const iceServers = config.ICEServers?.length > 0 ? [{urls: config.ICEServers}] : [];
+            const configServers = getIceServers(this.getState());
```
cpoile, line 162: Is there a specific reason we want to be requesting fresh iceServers on `join`, or was that just a way to get the data? We can still request fresh data on `join` if we want to.

webapp/src/index.tsx
```diff
@@ -506,9 +506,9 @@ export default class Plugin {
             }
 
             try {
-                const resp = await axios.get(`${getPluginPath()}/config`);
+                const allowEnable = allowEnableCalls(store.getState());
```
cpoile, line 509: This is the main benefit.

webapp/src/client.ts
```diff
@@ -31,7 +34,7 @@ export default class CallsClient extends EventEmitter {
     private audioTrack: MediaStreamTrack | null;
     public isHandRaised: boolean;
 
-    constructor() {
+    constructor(getState: GetStateFunc) {
```
streamer45, line 37: I am not that in love with this kind of coupling if it can be avoided. Could we pass the ice servers directly instead?

webapp/src/client.ts
streamer45: I think it was just about needing them there which automatically supported the case of changes at runtime but it's not a big requirement as it's mostly static information, configured once and left there.

server/api.go
streamer45: This is a breaking change though right? Are you thinking of handling it on mobile somehow or dropping support?

server/api.go
```diff
@@ -286,6 +287,33 @@ func (p *Plugin) handleDebug(w http.ResponseWriter, r *http.Request) {
...
 
+// handleConfig returns the client configuration, and cloud license information
+// that isn't exposed to clients yet on the webapp
+func (p *Plugin) handleConfig(w http.ResponseWriter) error {
+	license := p.pluginAPI.System.GetLicense()
+	if license == nil {
+		p.handleErrorWithCode(w, http.StatusBadRequest, "no license",
+			errors.New("no license found"))
+		return nil
+	}
```
streamer45, line 298: Surely we'd like to return the config to poor unlicensed servers as well right? :p

server/api.go
cpoile: Right, I must have blanked on that. Fixed.

webapp/src/client.ts
cpoile: Sure, passing the iceServers is cleaner, thanks

server/api.go
cpoile: haha, oops. Defaulting to `starter` seems like the best option.

PR #87, opened 2022-05-27
[MM-44155] Implement `/call end` slash command

streamer45:
#### Summary

PR introduces a new `/call end` slash command to end a call for all participants. This was made necessary as a way to fix some state issues we've been seeing lately. Originally a `cleanupState` functionality would run during plugin initialization to fix any potentially inconsistent state left by a sudden crash but this created an issue in HA deployments where one plugin instance would overwrite a perfectly healthy state on start up. Now removing the state cleanup call in case of HA deployment and relying on the end call command to fix potentially stuck calls.
 
Permissions wise the command is available to any system admin and to the call creator, meaning the user starting the call (aka the first participant to join).

@abhijit-singh Asking for your review since the PR introduces some minor UI but this shouldn't be considered a proper feature just yet, more like a fail-safe mechanism in case calls get stuck due to server crash or worse.

/cc @tao for visibility

#### Screenshots
image

#### Private/Public channels

image

#### DMs / GMs

image

#### Error case

error_end_call

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-44155

### DIFF
--- a/server/api.go
+++ b/server/api.go
@@ -8,12 +8,14 @@ import (
 	"net/http/pprof"
 	"regexp"
 	"strings"
+	"time"
 
 	"github.com/mattermost/mattermost-server/v6/model"
 	"github.com/mattermost/mattermost-server/v6/plugin"
 )
 
 var chRE = regexp.MustCompile(`^\/([a-z0-9]+)$`)
+var callEndRE = regexp.MustCompile(`^\/calls\/([a-z0-9]+)\/end$`)
 
 type Call struct {
 	ID              string      `json:"id"`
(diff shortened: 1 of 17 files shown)


server/api.go
```diff
@@ -180,59 +182,141 @@ func (p *Plugin) handleGetAllChannels(w http.ResponseWriter, r *http.Request) {
...
+			return
+		}
+		if state == nil || state.Call == nil || state.Call.ID != callID {
+			return
+		}
+
+		p.LogInfo("call state is still in store, force ending it")
+
+		for connID := range state.Call.Sessions {
+			if err := p.closeRTCSession(userID, connID, channelID, state.NodeID); err != nil {
```
cpoile, line 253: ~~I kind of wonder why we don't just do this, since the clients can handle it...?~~ Oh I see, the UI modal to notify the users. Got it.

webapp/src/actions.ts
```diff
@@ -155,3 +164,8 @@ export const requestTrial = () => {
         return {};
     };
 };
+
+export const endCall = (channelID: string) => {
+    return axios.post(`${getPluginPath()}/calls/${channelID}/end`, null,
```
cpoile, line 169: Just curious why we're using axios instead of fetch?

webapp/src/components/end_call_modal/component.scss
```diff
@@ -0,0 +1,42 @@
+%button-end-call-modal {
```
cpoile, line 1: ahhh... nooo... scss.... :( 
But it doesn't matter I guess. I have to go through and convert to styled-components someday later.

webapp/src/components/end_call_modal/component.tsx
```diff
@@ -0,0 +1,205 @@
+import React, {CSSProperties} from 'react';
```
cpoile, line 1: I think we could do this a bit simpler by using the generic modal we just added. Probably only a dozen lines of code or so. If you agree, not sure if you want to do that now or leave it for later? Up to you.

cpoile (approved):
Just comments, nothing blocking (it's up to you if you want to just get this out quickly?)

server/api.go
```diff
@@ -180,59 +182,141 @@ func (p *Plugin) handleGetAllChannels(w http.ResponseWriter, r *http.Request) {
...
+		res.Code = http.StatusForbidden
+		return
+	}
+
+	p.API.PublishWebSocketEvent(wsEventCallEnd, map[string]interface{}{}, &model.WebsocketBroadcast{ChannelId: channelID, ReliableClusterSend: true})
+
+	go func() {
+		// We wait a few seconds for the call to end cleanly. If this doesn't
+		// happen we force end it.
+		time.Sleep(5 * time.Second)
```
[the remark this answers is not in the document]
streamer45: One reason is that we can't clients to disconnect but most importantly this covers the case where the server crashed so there's no client connected to ws at all but the state in the store is still saved as if a call was ongoing.

server/api.go
streamer45: We actually could but it's much less clean on the UI as the RTC connection could timeout which would make ending a call feel very "unsynchronized" between clients.

webapp/src/components/end_call_modal/component.scss
streamer45: Sorry :see_no_evil:

webapp/src/components/end_call_modal/component.tsx
streamer45: Agreed. I'll have a look and create a separate ticket if it becomes a blocker.

webapp/src/index.tsx
```diff
@@ -213,6 +216,18 @@ export default class Plugin {
             });
         });
 
+        registry.registerWebSocketEventHandler(`custom_${pluginId}_call_end`, (ev) => {
+            if (connectedChannelID(store.getState()) === ev.broadcast.channel_id && window.callsClient) {
+                window.callsClient.disconnect();
+            }
+            store.dispatch({
+                type: VOICE_CHANNEL_CALL_END,
```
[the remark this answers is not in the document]
cpoile: Yep, I didn't really think of that -- you'd have to send x number of events. Better to send just this one. 👍

abhijit-singh (commented):
Looks good! 

One question — if a user is not the call creator and they run the command, would they still see the confirmation popup? I would instead expect to see a error message right there something like `You do not have permission to end the call. Please ask the call creator to end call.`

Please feel free to move forward if it's urgent.

streamer45:
Thanks @abhijit-singh, that's great feedback. We can fail fast on sending the slash command as such:

error_msg

Would that work?

webapp/src/components/end_call_modal/component.tsx
streamer45: Had a good look and almost implemented it so it's totally doable but then there's a bit of time to spend on styling it properly, centering content and such so I am looking to defer this work for later as we need to convert a few of these anyway. Created [MM-44708](https://mattermost.atlassian.net/browse/MM-44708) to keep track of this work.

abhijit-singh (approved):
@streamer45 Yep that works, it's consistent with how we're showing other warnings. 

Maybe @justinegeffen can review the copy if we have time, but looks good to me!

server/slash_command.go
```diff
@@ -32,6 +33,7 @@ func getAutocompleteData() *model.AutocompleteData {
 	data.AddCommand(model.NewAutocompleteData(leaveCommandTrigger, "", "Leaves a call in the current channel"))
```
justinegeffen, line 33:
```suggestion
	data.AddCommand(model.NewAutocompleteData(leaveCommandTrigger, "", "Leave a call in the current channel."))
```

server/slash_command.go
```diff
@@ -32,6 +33,7 @@ func getAutocompleteData() *model.AutocompleteData {
...
 	data.AddCommand(model.NewAutocompleteData(linkCommandTrigger, "", "Generates a link to join a call in the current channel"))
```
justinegeffen, line 34:
```suggestion
	data.AddCommand(model.NewAutocompleteData(linkCommandTrigger, "", "Generate a link to join a call in the current channel."))
```

server/slash_command.go
```diff
@@ -32,6 +33,7 @@ func getAutocompleteData() *model.AutocompleteData {
...
 	data.AddCommand(model.NewAutocompleteData(statsCommandTrigger, "", "Shows some client-generated statistics about the call"))
```
justinegeffen, line 35:
```suggestion
	data.AddCommand(model.NewAutocompleteData(statsCommandTrigger, "", "Show client-generated statistics about the call."))
```

server/slash_command.go
```diff
@@ -32,6 +33,7 @@ func getAutocompleteData() *model.AutocompleteData {
...
+	data.AddCommand(model.NewAutocompleteData(endCommandTrigger, "", "End the call for everyone. All the participants will drop immediately"))
```
justinegeffen, line 36:
```suggestion
	data.AddCommand(model.NewAutocompleteData(endCommandTrigger, "", "End the call for everyone. All the participants will drop immediately."))
```

server/slash_command.go
```diff
@@ -32,6 +33,7 @@ func getAutocompleteData() *model.AutocompleteData {
...
 
 	experimentalCmdData := model.NewAutocompleteData(experimentalCommandTrigger, "", "Turns on/off experimental features")
```
justinegeffen, line 38:
```suggestion
	experimentalCmdData := model.NewAutocompleteData(experimentalCommandTrigger, "", "Turn experimental features on or off.")
```

webapp/src/index.tsx
```diff
@@ -306,6 +324,23 @@ export default class Plugin {
...
                 }
                 return {error: {message: 'You are not connected to a call in the current channel.'}};
+            case 'end':
+                if (voiceConnectedUsersInChannel(store.getState(), args.channel_id)?.length === 0) {
+                    return {error: {message: 'No ongoing call in the channel.'}};
+                }
+
+                if (!isCurrentUserSystemAdmin(store.getState()) &&
+                    getCurrentUserId(store.getState()) !== voiceChannelCallCreatorID(store.getState(), args.channel_id)) {
+                    return {error: {message: 'You do not have permission to end the call. Please ask the call creator to end call.'}};
```
justinegeffen, line 334:
```suggestion
                    return {error: {message: 'You don't have permission to end the call. Please ask the call creator to end the call.'}};
```

webapp/src/index.tsx
```diff
@@ -306,6 +324,23 @@ export default class Plugin {
                     return {};
                 }
                 return {error: {message: 'You are not connected to a call in the current channel.'}};
```
justinegeffen, line 326:
```suggestion
                return {error: {message: 'You're not connected to a call in the current channel.'}};
```

webapp/src/index.tsx
justinegeffen: @streamer45, can we replace "call creator" with "host" or "owner"?

justinegeffen (commented):
Some re-wording and a question about "call creator".

webapp/src/index.tsx
streamer45: We can, I guess it's more of a product question at this point as we don't really have a concept of "hosting a call". The "creator" is the one that originally started the call. I am okay with owner. /cc @itao

itao: I'm onboard with `owner` as it's implicitly transferrable, whereas `creator` can only be one person evern.

streamer45: Perfect. I'll update, thanks!

justinegeffen (approved):

PR #91, opened 2022-05-31
[MM-44679] Use config to determine WebSocket URL

### LINKED ISSUE #90: Inherit configuration of Mattermost config.json (or db config) for WebsocketURL
#### Description of problem:

Mattermost Calls plugin assumes the websocket URL is equal to the board URL, but in some deployments, the websocket URL must be configured to a subdomain in order to bypass a CDN which is serving static assets to users at the edge. In this case you may have service at chat.example.com, but websockets via only ws.chat.example.com. [1]

This was originally an issue for the Focalboard plugin of Mattermost but it was resolved at some point. I am trying to identify an exact commit.

#### Expected behavior:

Calls plugin works in mattermost server configured as above where `SiteURL` != `WebsocketURL`

#### Actual behavior:

Calls plugin attempts to start a call but is unable to open a websocket connection to the SiteURL, so fails. If the SiteURL's DNS record is changed locally (e.g., with a hosts file change) to bypass the CDN, Calls plugin appears to function as expected.

[1] https://docs.mattermost.com/configure/configuring-cloudfront-to-host-mattermost-static-assets.html

streamer45:
#### Summary

We should take into account if server's config has `WebsocketURL` set in order to properly ovverride the URL we connect to when starting/joining a call.

Fixes #90 

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-44679

### DIFF
--- a/webapp/src/utils.test.ts
+++ b/webapp/src/utils.test.ts
@@ -1,0 +1,30 @@
+import {
+    getWSConnectionURL,
+} from './utils';
+
+describe('utils', () => {
+    describe('getWSConnectionURL', () => {
+        const testCases = [
+            {
+                description: 'undefined',
+                config: {},
+                expected: 'ws://localhost:8065/api/v4/websocket',
+            },
+            {
+                description: 'empty config.WebsocketURL',
+                config: {WebsocketURL: ''},
+                expected: 'ws://localhost:8065/api/v4/websocket',
+            },
+            {
+                description: 'config.WebsocketURL',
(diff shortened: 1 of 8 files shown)


webapp/src/utils.test.ts
```diff
@@ -0,0 +1,30 @@
+import {
+    getWSConnectionURL,
+} from './utils';
+
+describe('utils', () => {
+    describe('getWSConnectionURL', () => {
```
cpoile, line 6: nice!

cpoile (approved):
👍

PR #93, opened 2022-06-02
[MM-44651] Implement `MaxCallParticipants` config setting

streamer45:
#### Summary

PR implements a new `MaxCallParticipants` config setting to optionally limit the maximum number of participants in calls.

- For on-prem the default is 0 meaning unlimited. 
- For Cloud the default is 8. 

Config value can be also be overridden through the `MM_CALLS_MAX_PARTICIPANTS` environment variable if needed.

@cpoile I made various changes to the logic to avoid having two values for essentially the same thing. It would be great if you could do some QA on this to make sure I haven't missed anything.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-44651

### DIFF
--- a/server/activate.go
+++ b/server/activate.go
@@ -53,6 +53,18 @@ func (p *Plugin) OnActivate() error {
 		return err
 	}
 
+	// On Cloud installations we want calls enabled in all channels so we
+	// override it since the plugin's default is now false.
+	if isCloud(p.pluginAPI.System.GetLicense()) {
+		cfg.DefaultEnabled = new(bool)
+		*cfg.DefaultEnabled = true
+		if err := p.setConfiguration(cfg); err != nil {
+			err = fmt.Errorf("failed to set configuration: %w", err)
+			p.LogError(err.Error())
+			return err
+		}
+	}
+
 	if cfg.RTCDServiceURL != "" {
 		rtcdManager, err := p.newRTCDClientManager(cfg.RTCDServiceURL)
 		if err != nil {
--- a/server/activate.go
+++ b/server/activate.go
@@ -3,7 +3,6 @@ package main
 import (
 	"fmt"
 	"os"
-	"strconv"
 	"time"
 
 	"github.com/mattermost/rtcd/service/rtc"
@@ -40,29 +39,6 @@ func (p *Plugin) OnActivate() error {
 
 	cfg := p.getConfiguration()
 	if err := cfg.IsValid(); err != nil {
-		p.LogError(err.Error())
-		return err
-	}
-
-	if isCloud(p.pluginAPI.System.GetLicense()) {
-		cfg.MaxCallParticipants = new(int)
-		*cfg.MaxCallParticipants = cloudMaxParticipantsDefault
-		// On Cloud installations we want calls enabled in all channels so we
-		// override it since the plugin's default is now false.
-		cfg.DefaultEnabled = new(bool)
-		*cfg.DefaultEnabled = true
-	}
-
-	if maxPart := os.Getenv("MM_CALLS_MAX_PARTICIPANTS"); maxPart != "" {
-		if max, err := strconv.Atoi(maxPart); err == nil {
-			*cfg.MaxCallParticipants = max
-		} else {
-			p.LogError("activate", "failed to parse MM_CALLS_MAX_PARTICIPANTS", err.Error())
-		}
-	}
-
-	if err := p.setConfiguration(cfg); err != nil {
-		err = fmt.Errorf("failed to set configuration: %w", err)
 		p.LogError(err.Error())
 		return err
 	}
--- a/server/activate.go
+++ b/server/activate.go
@@ -44,26 +44,27 @@ func (p *Plugin) OnActivate() error {
 		return err
 	}
 
-	// On Cloud installations we want calls enabled in all channels so we
-	// override it since the plugin's default is now false.
 	if isCloud(p.pluginAPI.System.GetLicense()) {
 		cfg.MaxCallParticipants = new(int)
 		*cfg.MaxCallParticipants = cloudMaxParticipantsDefault
-		if maxPart := os.Getenv("MM_CALLS_MAX_PARTICIPANTS"); maxPart != "" {
-			if max, err := strconv.Atoi(maxPart); err == nil {
-				*cfg.MaxCallParticipants = max
-			} else {
-				p.LogError("activate", "failed to parse MM_CALLS_MAX_PARTICIPANTS", err.Error())
-			}
-		}
-
+		// On Cloud installations we want calls enabled in all channels so we
+		// override it since the plugin's default is now false.
 		cfg.DefaultEnabled = new(bool)
 		*cfg.DefaultEnabled = true
-		if err := p.setConfiguration(cfg); err != nil {
-			err = fmt.Errorf("failed to set configuration: %w", err)
-			p.LogError(err.Error())
-			return err
+	}
+
+	if maxPart := os.Getenv("MM_CALLS_MAX_PARTICIPANTS"); maxPart != "" {
+		if max, err := strconv.Atoi(maxPart); err == nil {
+			*cfg.MaxCallParticipants = max
+		} else {
+			p.LogError("activate", "failed to parse MM_CALLS_MAX_PARTICIPANTS", err.Error())
 		}
+	}
+
+	if err := p.setConfiguration(cfg); err != nil {
+		err = fmt.Errorf("failed to set configuration: %w", err)
+		p.LogError(err.Error())
+		return err
 	}
 
 	if cfg.RTCDServiceURL != "" {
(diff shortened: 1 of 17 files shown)


server/api.go
```diff
@@ -298,13 +298,11 @@ func (p *Plugin) handleConfig(w http.ResponseWriter) error {
...
 	type config struct {
 		clientConfig
-		SkuShortName         string `json:"sku_short_name"`
-		CloudMaxParticipants int    `json:"cloud_max_participants"`
+		SkuShortName string `json:"sku_short_name"`
 	}
 	ret := config{
-		clientConfig:         p.getConfiguration().getClientConfig(),
-		SkuShortName:         skuShortName,
-		CloudMaxParticipants: cloudMaxParticipants,
```
streamer45, line 307: I realize this is a breaking change but given it only affects Cloud I think we can *safely* make it now.

webapp/src/types/types.ts
```diff
@@ -58,16 +58,16 @@ export type CallsConfig = {
     ICEServers: string[],
     AllowEnableCalls: boolean,
     DefaultEnabled: boolean,
+    MaxCallParticipants: number,
```
streamer45, line 61: Going snake case makes the server code more complex as we need yet another config wrapper, not sure if it's worth it to be honest but let me know.

cpoile: Nah, it's fine. We're kind of stuck with it for now.

cpoile (commented):
Looks good code-wise, I went through it carefully. I will approve after I do QA on it (will take awhile to get the server set up and run through all the combinations).

server/configuration.go
```diff
@@ -162,6 +172,10 @@ func (c *configuration) Clone() *configuration {
 		}
 	}
 
+	if c.MaxCallParticipants != nil {
+		cfg.MaxCallParticipants = model.NewInt(*c.MaxCallParticipants)
```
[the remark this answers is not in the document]
streamer45: Honestly, I feel like Hamlet now :skull:  

Part of me wants to avoid relying on `model` as much as possible but it's so tempting to save a line by using the util...

server/cloud_limits.go
```diff
@@ -7,36 +7,13 @@ import (
 	"net/http"
 )
 
-// cloudMaxParticipants defaults to 8, can be overridden by setting the env variable
-// MM_CALLS_CLOUD_MAX_PARTICIPANTS
-var cloudMaxParticipants = 8
+// cloudMaxParticipantsDefault is set to 8.
+// The value used can be overridden by setting the MM_CLOUD_MAX_PARTICIPANTS env variable.
```
cpoile, line 11: Shouldn't it be?:

```suggestion
// The value used can be overridden by setting the MM_CALLS_MAX_PARTICIPANTS env variable.
```

streamer45: Oh dear me :man_facepalming:

server/activate.go
```diff
@@ -56,6 +47,16 @@ func (p *Plugin) OnActivate() error {
 	// On Cloud installations we want calls enabled in all channels so we
 	// override it since the plugin's default is now false.
 	if isCloud(p.pluginAPI.System.GetLicense()) {
+		cfg.MaxCallParticipants = new(int)
+		*cfg.MaxCallParticipants = cloudMaxParticipantsDefault
+		if maxPart := os.Getenv("MM_CALLS_MAX_PARTICIPANTS"); maxPart != "" {
```
cpoile, line 52: Discovered in QA: Do we not want to allow on-prem to use the `MM_CALLS_MAX_PARTICIPANTS` flag?

streamer45: I think it's more like we need it on Cloud, not so much on on-prem although given it's not possible to use common configuration overrides due to the plugin id issue I think it makes sense to allow this to be used on non-cloud installations as well. I can update :+1:

itao (approved):

cpoile:
This has been a disaster to test. First, after setting up a mocked CWS system and a custom docker image, and spinning up a test server using that hash, you can only upload the plugin once. After that the plugin upload button is deactivated (but you can get around that by never leaving the system console). Second, often you'll get an error back from uploading a plugin: 
image

If you get that, your server is toast -- it will no longer accept plugin uploads, so you have to spin up a new one.

Third, the `MAX_CALLS_MAX_PARTICIPANTS` env var wasn't working at all. Finally figured out that `OnConfigurationChange` is getting called after the `onActivate` has set the overrides, which overwrites the config with the one from the system console. So [36c67f6](https://github.com/mattermost/mattermost-plugin-calls/pull/93/commits/36c67f6a8850410cf3ef533e1530ea915baa7073) fixes that and lets us have permanent overrides. (FYI: This happens with on-prem servers as well.) But that means cloud customers cannot turn off default calls, which we probably don't want. So we have to find a way to make the `setOverrides` override the plugin default, but not once the user has changed it manually. 

Fourth, I could not test whether the manual max participants configuration setting works in cloud, because it wouldn't let me change any config settings. It would respond with:
image

Finally, I could not test the max 8 default for cloud. This is because I could not change other settings like `Enable Open Server` because it responded with:
image

What I was able to test:
- with [36c67f6](https://github.com/mattermost/mattermost-plugin-calls/pull/93/commits/36c67f6a8850410cf3ef533e1530ea915baa7073) the env var works correctly and is sent to webapp (and mobile) and the UI changes as expected. 

What I was not able to test:
- changing the config setting in the system console works as expected (i.e., is transmitted to the clients and changes the limits and the UI as expected)
- that changing the config setting for default enabled in cloud is respected (it won't be, at the moment)
- that the cloud max of 8 participants was respected (the point above about not being able to change Enable Open Server)

Issues found:
- We should add the N/A cursor for the channel header button and channel header dropdown button.
- Config changes require the webapp and mobile to reload before they get the new config values, which means they'll try to join a call and get no response. I think this is probably "ok" for beta, but not later.

So at the moment I think we just have to ship it and hope? And then we need to get proper cloud freemium test servers with plugin uploads.

streamer45:
Thanks @cpoile ,

I am sorry you had to go through all that trouble just to test this but if anything it helped to reinforce even further the need to have better tools/permissions when testing/debugging in Cloud. I'll make sure to raise these concerns in our future discussions with Cloud/SREs.

For the sake of this PR could we test whatever is feasible using cloud licenses on a local env? I understand it won't be 100% accurate (as we have seen already) but it's better than nothing.

---

> Third, the MAX_CALLS_MAX_PARTICIPANTS env var wasn't working at all. Finally figured out that OnConfigurationChange is getting called after the onActivate has set the overrides, which overwrites the config with the one from the system console.

Interesting, there must be some racy behaviour with the two operations then given that locally I was consistently getting the activation after the initial config change trigger. I am almost regretting not going forward with the plugin id change since it would have prevented all this :grimacing: 

Anyhow, I think for the time being we can put in place a little check and apply the overrides as soon as the configuration has been loaded the first time. E.g.:

```go
func (p *Plugin) setConfiguration(configuration *configuration) error {
	p.configurationLock.Lock()
	defer p.configurationLock.Unlock()

	if p.configuration == nil && configuration != nil {
		setOverrides(configuration)
	}
```

What do you think?

---

> Config changes require the webapp and mobile to reload before they get the new config values, which means they'll try to join a call and get no response. I think this is probably "ok" for beta, but not later.

Yes this is expected and fine for now. Created [MM-44781](https://mattermost.atlassian.net/browse/MM-44781).

> We should add the N/A cursor for the channel header button and channel header dropdown button.

I think there may still be some issues with on-prem I guess where we should show the blocked cursor anyway since there's nothing to click but it's a minor thing.

cpoile:
No worries. :)

> For the sake of this PR could we test whatever is feasible using cloud licenses on a local env? I understand it won't be 100% accurate (as we have seen already) but it's better than nothing.

I've done that, and that would be the "let's just hope" option. In the end I don't think there is a way to finish testing this at the moment, so it's the only option. 

> What do you think?

That won't work, unfortunately, because the pluginAPI isn't set at that point either.

I think that putting the overrides in the `OnActivate` works only when the server is started/restarted with the EnvVars. Which I can't test because this PR's plugin isn't in the prepackaged plugins list for a test server. I could try to make that happen, but that would be a lot of work.

Instead, when you upload a plugin, that's when the overrides in the `OnActivate` don't work, because the `OnConfigurationChange` is called once more after the `OnActivate`, and overwrites the OnActivate overrides.
Maybe this is not intended and it's a bug?

But either way, maybe we can keep it in the `OnActivate` and then just know that any time the plugin is updated (either manually, which isn't possible, right? -- or through the marketplace, which is possible, right?) the env var and defaultEnabled will be lost.

cpoile:
Done. Was able to test everything, including mobile. I wasn't able to test that changing the config setting in the system console doesn't affect the overrides, because of the error I talked about above. But I'm assuming so, b/c of the code.

I think we're good to go @streamer45

streamer45:
Thanks @cpoile for the huge effort here :100:

cpoile (approved):

PR #98, opened 2022-06-07
[MM-44894] Check rtcd URL override prior to activation

streamer45:
#### Summary

This somehow slipped through our tests. In Cloud the config setting will be empty since we set it through env but the client was not being created unless the config value was set. Now checking for both config and env through a useful configuration getter.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-44894


cpoile (approved):
Nice!

PR #104, opened 2022-06-09
[MM-44965] Fix following threads

streamer45:
#### Summary

We were following any thread upon receiving call start event. Added a check to make sure we only follow the call we are currently in.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-44965

### DIFF
--- a/webapp/src/index.tsx
+++ b/webapp/src/index.tsx
@@ -128,6 +128,13 @@ export default class Plugin {
                     const audio = new Audio(getPluginStaticPath() + JoinUserSound);
                     audio.play();
                 }
+
+                if (window.callsClient.channelID === channelID) {
+                    const channel = getChannel(store.getState(), channelID);
+                    if (channel) {
+                        followThread(channel.id, channel.team_id);
+                    }
+                }
             }
 
             store.dispatch({
@@ -219,11 +226,6 @@ export default class Plugin {
                     rootPost: ev.data.thread_id,
                 },
             });
-
-            const channel = getChannel(store.getState(), ev.broadcast.channel_id);
-            if (channel) {
-                followThread(channel.id, channel.team_id);
-            }
         });
 
         registry.registerWebSocketEventHandler(`custom_${pluginId}_call_end`, (ev) => {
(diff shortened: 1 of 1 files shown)


cpoile (approved):
ahhh, I see. Nice :)

streamer45:
Actually we may want to add a check on the user id as well to avoid following multiple times. Will send an update.

streamer45:
@cpoile Move it back to the call start handler so it's handled only once, if we have started/joined the call that is. Please take a look.

webapp/src/index.tsx
```diff
@@ -211,21 +204,30 @@ export default class Plugin {
...
             store.dispatch({
                 type: VOICE_CHANNEL_ROOT_POST,
                 data: {
-                    channelID: ev.broadcast.channel_id,
+                    channelID,
                     rootPost: ev.data.thread_id,
                 },
             });
+
+            if (window.callsClient?.channelID === channelID) {
```
cpoile, line 225: Does this mean the `callsClient` has to be running before the `_call_start` event for the thread to be followed? Is that a possible race condition?

streamer45: If we are the user that started the call then it's more or less guaranteed as it means we already initialized the client to send the join message. The only weird case that could happen is:

- Init client
- Start a call
- Something fails and we drop
- Deinit client
- Start event is received, nothing happens

But again, if we were to join again (assuming the call is still on) we would follow anyway so I think it's okay.

cpoile: What if we're a user joining the call? (I think it would work, but I just want to double check I'm following the order of things.)

streamer45: If we are joining the call the call we fallback to the other cases which will just fetch the thread id from the store and follow the thread directly. The problem we were originally addressing here was that upon starting a call the thread id wasn't in the store yet so we moved the call to the call_start event handler.

cpoile: Ah, I see now, my mistake. I didn't read carefully enough. 👍

PR #105, opened 2022-06-09
MM-44946: Fix: Calls: start trial path from Start Call button not working

cpoile:
#### Summary
- Use the openCloudPricing flow instead of our homegrown trial path because the code changed on webapp/server and broke our flow; it's too late to export the webapp's flow for this, so we're using the CloudPricing modal as a temporary measure to make things work for launch.
- Next step tracked in: https://mattermost.atlassian.net/browse/MM-44992

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-44946


cpoile:
I'm not able to test this, unfortunately (split.io). So I'll ask for a review now and hope something changes tomorrow, but probably not in the time we have.

streamer45:
> I'm not able to test this, unfortunately (split.io). So I'll ask for a review now and hope something changes tomorrow, but probably not in the time we have. (Do we even have time? I'm unsure whether we can fix this before release.)

We don't have time to include it in 7.0/cloud build but what we can probably do, similarly to other bugs (e.g. the thread autofollow one) we can cut a new release later today, push it to the marketplace and force an update on cloud (through split). That should give us what we want without going through pre-packaging.

cpoile:
@streamer45 Ok, that sounds good, let's do that. 👍

streamer45:
@cpoile Do you want to test this before or after merging?

cpoile:
@streamer45 Do we need to test anything else in this batch? If not, I can cut it from this hash.

streamer45:
@cpoile I think this is the only PR requiring an actual Cloud instance to be verified so I'd say this is it. I don't have to merge anything else for v0.6.1 (or whatever it's going to be).

streamer45 (approved):

PR #107, opened 2022-06-09
MM-44567 - Fix: On first load, channel header button is not in correct state

cpoile:
#### Summary
- original problems from the ticket seem to be fixed. But now the problem is the channel header button does not have the correct config data from the server. The first config call was unauthorized, and we don’t make another one until reconnect. This leads to a user seeing the "Start call" button available, when it should be disabled (e.g., cloud starter in a public channel, cloud limits have been reached, or max participants settings). This will persist until a websocket reconnect.
- Now we fetch the config data if we haven't fetched it yet. The user sees the correct state on the first channel load.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-44567

### DIFF
--- a/webapp/src/index.tsx
+++ b/webapp/src/index.tsx
@@ -26,7 +26,6 @@ import {
     voiceChannelRootPost,
     allowEnableCalls,
     iceServers,
-    callsConfig,
 } from './selectors';
 
 import {pluginId} from './manifest';
@@ -636,8 +635,15 @@ export default class Plugin {
             }
         };
 
+        let configRetrieved = false;
         const onActivate = async () => {
-            store.dispatch(getCallsConfig());
+            const res = await store.dispatch(getCallsConfig());
+
+            // @ts-ignore
+            if (!res.error) {
+                configRetrieved = true;
+            }
+
             fetchChannels();
             const currChannelId = getCurrentChannelId(store.getState());
             if (currChannelId) {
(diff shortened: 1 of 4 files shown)


webapp/src/index.tsx
```diff
@@ -687,6 +688,12 @@ export default class Plugin {
             const currentChannelId = getCurrentChannelId(store.getState());
             if (currChannelId !== currentChannelId) {
                 currChannelId = currentChannelId;
+
+                // If we haven't retrieved config, user must not have been logged in during onActivate
+                if (!callsConfig(store.getState()).retrieved) {
+                    store.dispatch(getCallsConfig());
+                }
```
streamer45, line 695: Is there any benefit in "polluting" the config reducer instead of having a state variable in scope to keep track of whether we retrieved it or not?

cpoile: So yah, that's the benefit to putting the flag in the reducer -- it keeps everything clean and async. I personally don't mind it, and it has the benefit of keeping the logic in one place.

streamer45: Do we have to check the error? I was more like thinking of tracking the attempt, assuming in most cases it will work but do you think it could fail somehow and cause issues?

streamer45 (approved):
Thanks @cpoile , good conversation there :+1:

webapp/src/index.tsx
streamer45: Nothing wrong with your approach really, it just felt weird to me re-using the config object itself to keep track of that.

PR #108, opened 2022-06-10
[MM-44486] Support advanced ICE server configs through ICEServersConfigs

streamer45:
#### Summary

Adding a new `ICEServersConfigs` config setting to support passing to clients both STUN and TURN server configs (with credentials). This is a superset of the existing `ICEServers` which we are deprecating but not removing to maintain backwards compatibility.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-44486

### DIFF
--- a/server/configuration.go
+++ b/server/configuration.go
@@ -41,8 +41,9 @@ type configuration struct {
 }
 
 type clientConfig struct {
-	// A comma separated list of ICE servers URLs (STUN/TURN) to use.
+	// **DEPRECATED use ICEServersConfigs** A comma separated list of ICE servers URLs (STUN/TURN) to use.
 	ICEServers ICEServers
+
 	// A list of ICE server configurations to use.
 	ICEServersConfigs ICEServersConfigs
 	// When set to true, it allows channel admins to enable or disable calls in their channels.
--- a/server/configuration.go
+++ b/server/configuration.go
@@ -199,16 +199,12 @@ func (c *configuration) Clone() *configuration {
 
 	if c.ICEServers != nil {
 		cfg.ICEServers = make(ICEServers, len(c.ICEServers))
-		for i, u := range c.ICEServers {
-			cfg.ICEServers[i] = u
-		}
+		copy(cfg.ICEServers, c.ICEServers)
 	}
 
 	if c.ICEServersConfigs != nil {
 		cfg.ICEServersConfigs = make([]rtc.ICEServerConfig, len(c.ICEServersConfigs))
-		for i, u := range c.ICEServersConfigs {
-			cfg.ICEServersConfigs[i] = u
-		}
+		copy(cfg.ICEServersConfigs, c.ICEServersConfigs)
 	}
 
 	if c.MaxCallParticipants != nil {
--- a/server/configuration.go
+++ b/server/configuration.go
@@ -4,12 +4,15 @@
 package main
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
 	"os"
 	"reflect"
 	"strconv"
 	"strings"
+
+	"github.com/mattermost/rtcd/service/rtc"
 
 	"github.com/mattermost/mattermost-server/v6/model"
 )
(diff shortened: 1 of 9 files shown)


server/configuration.go
```diff
@@ -177,6 +204,13 @@ func (c *configuration) Clone() *configuration {
 		}
 	}
 
+	if c.ICEServersConfigs != nil {
+		cfg.ICEServersConfigs = make([]rtc.ICEServerConfig, len(c.ICEServersConfigs))
+		for i, u := range c.ICEServersConfigs {
+			cfg.ICEServersConfigs[i] = u
+		}
```
cpoile, line 211: nit: Could we do this?
```suggestion
		copy(cfg.ICEServersConfigs, c.ICEServersConfigs)
```
Same for 210

cpoile (approved):
Looks good!

streamer45:
@justinegeffen 

Here we are introducing a new config setting `ICEServerConfigs` that is set to replace the existing `ICEServers` which is getting deprecated internally but removed from the UI. We should update the official documentation accordingly. 

As for timing, it would be nice to have this in next server release (7.1). Let me know if I can help with anything.

streamer45:
@justinegeffen Let me know if you'd like to defer this to post merge.

justinegeffen:
Thanks, @streamer45! Will this be live now or only from 7.1?

justinegeffen:
To add to that, this can be done now if it'll be live now, otherwise the docs/needed label can stay and I'll pick it up for the 7.1 docs (which will be started in a week or so).

justinegeffen (approved):

streamer45:
> To add to that, this can be done now if it'll be live now, otherwise the docs/needed label can stay and I'll pick it up for the 7.1 docs (which will be started in a week or so). 

@justinegeffen If we merge now it would most likely be in 7.1 self-hosted which is probably the most relevant release as far as documentation. Plugin may go live on Cloud before depending on a few factors. 

> Either way, please don't let docs be a blocker for merging! :)

Thanks for clarifying, I'll keep adding labels so you can have a look in due time.

server/configuration.go
```diff
@@ -114,6 +140,7 @@ func (c *configuration) getClientConfig() clientConfig {
 		AllowEnableCalls:    c.AllowEnableCalls,
 		DefaultEnabled:      c.DefaultEnabled,
 		ICEServers:          c.ICEServers,
+		ICEServersConfigs:   c.getICEServers(),
```
cpoile, line 143: Leaving a comment here, but it applies to line 142 -- I realized that if this new version is merged, we could still have old versions of the mobile app using the new plugin. But if someone starts using the `ICEServersConfigs`, those won't be translated to old ICEServers list. (Actually, as far as I can tell, this PR removes ICEServers, so this will always be empty after this is merged, right?)

If I'm understanding this right, then we should add in a conversion to turn `ICEServersConfigs` into `ICEServers` when there isn't a user/credential field, and we need to add docs that adding a credentials server here is a breaking change for mobile apps < `1.54`.

cpoile (changes requested):
Requesting changes to make sure you see this before merging, thanks!

server/configuration.go
streamer45: > Actually, as far as I can tell, this PR removes ICEServers, so this will always be empty after this is merged, right?

We are only removing the field from the settings but the values would still be returned and configurable in `config.json` if needed. To be extra thorough we could put some migration in place to convert from deprecated `ICEServers` to `ICEServersConfigs` but it seems a bit overkill considering we are in beta and those fields aren't actively used.

As for backward mobile support I think that's a good point although I'd like to highlight a couple of things: 

- STUN servers aren't necessary on the client side to make calls work. 
- TURN servers were not usable at all before this PR.

So overall there's isn't much that could break due to this change if everything is configured appropriately. Only thing that comes to mind are cases where clients rely on STUN candidates but only because the network wasn't configured appropriately (opening UDP ports). But even then, mobile code still defaulted to the hardcoded google/twilio ones so it would still work.

I think that eventually it's up to us to decide if this is a breaking enough change to justify the overhead of conversion and then deprecating the code for future removal. Thoughts?

cpoile: Thanks for clarifying! I agree, I don't think it's breaking enough to warrant adding the conversion code and the debt. Let's merge. 👍

cpoile (approved):

PR #114, opened 2022-06-17
[MM-45155] Add support for using plugin when MM is served under a subpath

### LINKED ISSUE #112: Calls Plugin not working within subdirectory
We run Mattermost in a subdirectory, but apparently the Calls plugin can't handle it. No user icons are displayed and screen sharing, for example, refers to the main domain and ignores the subdirectory.

streamer45:
#### Summary

Fixes #112 

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-45155

### DIFF
--- a/webapp/src/websocket.ts
+++ b/webapp/src/websocket.ts
@@ -7,7 +7,7 @@ import {logWarn, logErr} from './log';
 
 export default class WebSocketClient extends EventEmitter {
     private ws: WebSocket | null;
-    private seqNo = 0;
+    private seqNo = 1;
     private connID = '';
     private eventPrefix: string = 'custom_' + pluginId;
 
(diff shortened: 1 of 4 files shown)


webapp/src/websocket.ts
```diff
@@ -7,7 +7,7 @@ import {logWarn, logErr} from './log';
 
 export default class WebSocketClient extends EventEmitter {
     private ws: WebSocket | null;
-    private seqNo = 0;
+    private seqNo = 1;
```
cpoile, line 10: just curious what this is for?

cpoile (approved):
👍

webapp/src/websocket.ts
streamer45: The change was to conform to what's expected on the server side as the zero value is treated as invalid. I found this during debugging but it doesn't affect anything at the moment. The sequence number is used by the server to figure out which events to resend to the client in case of an unexpected reconnection.

streamer45: Sorry, went into "teaching mode" there :sweat_smile:

Issue #49, opened 2022-04-25
Turn server config


justvanbloom:
Would be nice if we can insert shared cred and user-xx to make it use coturn.

streamer45:
Hi @justvanbloom,

are you looking to use a TURN server to relay the media to clients unable to connect directly?

We definitely have plans to support clients behind very strict firewalls but nothing has been decided yet on how exactly we want to implement it. While adding support for external TURN servers would be easy enough it's not something I am particularly excited about, especially as we are trying to avoid using 3rd party services where possible. So it may take a while before we support this functionality but it's definitely on the roadmap.

justvanbloom:
Uh. Okay. Perhaphs i can mod it to use a external turn. Yes clients behind corporate fw are unable to use calls right now

Sent via Commodore 64 in use with Dataphone s21-23d

Asgoret:
@streamer45 Would be great If u move it little up on roadmap and make in this year...or early :)

streamer45:
@Asgoret is your main requirement being able to set an external TURN server that you'll provide on your own or have an all in one solution?

Asgoret:
@streamer45 external provided by our own. In my opinion it's not worth it to do all features all-in-one. Some things can make by external solutions, you just need to create such opportunity. It will cost less hours to program and will be good metric for necessity of such feature. If somebody will use it as external and want in all-in-one they can create request.

streamer45:
Alright, thanks for the context. I created an internal ticket ([MM-44486](https://mattermost.atlassian.net/browse/MM-44486)) to keep track of this and hopefully include it sometime soon.

gecame:
Hi @all, @streamer45, +10 for this enhancement 👍  It's a pretty common VoIP problem and affects far more than just some corporate networks with strict firewalls. Actually, a significant fraction of mobile networks (and increasingly more and more home ISPs) use so-called symmetric NATs (e.g. [link1](https://dh2i.com/kbs/kbs-2961448-understanding-different-nat-types-and-hole-punching), [link2](https://webrtchacks.com/symmetric-nat/)) - these don't allow directly sending media end-to-end across network boundaries. Users behind such a mobile network/ISP will suffer from one-way audio or no audio at all without helpers such as TURN / ICE, even if they are connecting to a less restricted network.
[...]

PR #116, opened 2022-06-17
[MM-45001] Add support for generating short-lived TURN credentials

streamer45:
#### Summary

PR adds support for generating short-lived TURN credentials as described in https://datatracker.ietf.org/doc/html/draft-uberti-behave-turn-rest-00#section-2.2

Addresses https://github.com/mattermost/mattermost-plugin-calls/issues/49

#### Related PRs

https://github.com/mattermost/rtcd/pull/63

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-45001

### DIFF
--- a/server/activate.go
+++ b/server/activate.go
@@ -103,11 +103,16 @@ func (p *Plugin) OnActivate() error {
 		}()
 	}
 
-	rtcServer, err := rtc.NewServer(rtc.ServerConfig{
-		ICEPortUDP:      *cfg.UDPServerPort,
-		ICEHostOverride: cfg.ICEHostOverride,
-		ICEServers:      rtc.ICEServers(cfg.getICEServers()),
-	}, newLogger(p), p.metrics.RTCMetrics())
+	rtcServerConfig := rtc.ServerConfig{
+		ICEPortUDP:                       *cfg.UDPServerPort,
+		ICEHostOverride:                  cfg.ICEHostOverride,
+		ICEServers:                       rtc.ICEServers(cfg.getICEServers(false)),
+		TURNCredentialsExpirationMinutes: *cfg.TURNCredentialsExpirationMinutes,
+	}
+	if *cfg.ServerSideTURN {
+		rtcServerConfig.TURNStaticAuthSecret = cfg.TURNStaticAuthSecret
+	}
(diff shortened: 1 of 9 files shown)


server/activate.go
```diff
@@ -103,11 +103,16 @@ func (p *Plugin) OnActivate() error {
...
-	}, newLogger(p), p.metrics.RTCMetrics())
+	rtcServerConfig := rtc.ServerConfig{
+		ICEPortUDP:                       *cfg.UDPServerPort,
+		ICEHostOverride:                  cfg.ICEHostOverride,
+		ICEServers:                       rtc.ICEServers(cfg.getICEServers(false)),
+		TURNCredentialsExpirationMinutes: *cfg.TURNCredentialsExpirationMinutes,
+	}
+	if *cfg.ServerSideTURN {
+		rtcServerConfig.TURNStaticAuthSecret = cfg.TURNStaticAuthSecret
+	}
```
cpoile, line 114: Should we add text to the settings page that `changing this setting will require a server restart`?

webapp/src/index.tsx
```diff
@@ -460,9 +461,20 @@ export default class Plugin {
                     return;
                 }
 
+                const iceConfigs = [...iceServers(store.getState())];
+                if (needsTURNCredentials(store.getState())) {
+                    logDebug('turn credentials needed');
+                    try {
+                        const resp = await axios.get(`${getPluginPath()}/turn-credentials`);
```
cpoile, line 468: Just a note: we're going to have to replace these with Client4 requests. Not sure if you want to do that here, or wait for the ticket to do it bulk. Up to you!

cpoile (approved):
Thanks Claudio!

server/activate.go
streamer45: Great point. I actually had a similar thoughts when implementing this. I think we should do a review and document any setting (there are a few) that require a restart, a plugin restart that is (no need to restart the Mattermost server) but I guess that's what you meant. Will open a ticket to address this.

webapp/src/index.tsx
streamer45: I like `axios` :) I think we can defer to the ticket mostly because it's a bit more involved than just replacing a call. I'd like to make sure to have a proper object/superclass or something tidy. I like the approach we have on mobile, not sure if easily replicable here though.

oh6hay (approved):
LGTM :+1:

PR #117, opened 2022-06-17
e2e: Update snapshots + fix failure

Phrynobatrachus:
I did a quick audit of the existing tests to prepare for CI setup.


e2e/tests/popout.spec.ts
```diff
@@ -22,7 +22,7 @@ test.describe('popout window', () => {
         const text = await popOut.textContent('#calls-popout-mute-button');
         expect(text).toBe('Unmute');
 
-        await devPage.leaveCall();
+        await popOut.locator('.button-leave').click();
```
streamer45, line 25: Interesting. This should work regardless of whether the user leaves from the pop-out window or the widget.

streamer45:
Thanks @Phrynobatrachus, were you able to get the suite to pass with these changes?

A few comments:

[calls-widget-bottom-bar-dm-chromium-linux.png](https://github.com/mattermost/mattermost-plugin-calls/pull/117/files#diff-8cecbd76cbac527e312a69b4478bf911ffcf25be48702d8791e6260412a32925)

This seems wrong as the expectation here is that calls in DMs/GMs should be unmuted by default. Maybe it's just a matter of sleeping a little longer, not sure.

[e2e/tests/join_call.spec.ts-snapshots/channel-toast-chromium-linux.png](https://github.com/mattermost/mattermost-plugin-calls/pull/117/files#diff-ab38cef6c9429a9974798689e0243a3be6423ca4590f082b1a1be2f9c133ebcd)

I am not able to reproduce this change of width. The viewport is fixed so I am thinking something else may be affecting it. Are you making sure to run the tests on a fresh instance with nothing (plugins/teams?) that could affect sidebar width?

e2e/tests/start_call.spec.ts
```diff
@@ -241,7 +241,7 @@ test.describe('setting audio output device', () => {
         await devPage.leaveCall();
     });
 
-    test.only('setting default', async ({page}) => {
+    test('setting default', async ({page}) => {
```
streamer45, line 244: Oh wow, thanks :100:

Phrynobatrachus:
Yeah the suite passes, save for some flake on one of the `Join call` tests.

> This seems wrong as the expectation here is that calls in DMs/GMs should be unmuted by default. Maybe it's just a matter of sleeping a little longer, not sure.

Good to know, I'll double check this one.

> I am not able to reproduce this change of width. The viewport is fixed so I am thinking something else may be affecting it. Are you making sure to run the tests on a fresh instance with nothing (plugins/teams?) that could affect sidebar width?

I think the new App Bar might be the cause, I updated the snapshot with that enabled. I figure that's probably what we want going forward, do you agree?

streamer45:
> I think the new App Bar might be the cause, I updated the snapshot with that enabled. I figure that's probably what we want going forward, do you agree?

Oh that makes sense. Something to keep in mind and possibly document then. Thanks.

Phrynobatrachus:
I spent awhile looking at the mute/unmute snapshot issue today. It seems that some connection events don't fire when running the test:

```
peer.on('connect', () => {
  logDebug('rtc connected');
  this.emit('connect');
});
```
This `on()` doesn't execute, which in turn means that the `callsClient` logic that calls `unmute()` doesn't execute either.

```
window.callsClient.on('connect', () => {
  if (isDMChannel(this.props.channel) || isGMChannel(this.props.channel)) {
      window.callsClient.unmute();
  }
  this.setState({currentAudioInputDevice: window.callsClient.currentAudioInputDevice});
  this.setState({currentAudioOutputDevice: window.callsClient.currentAudioOutputDevice});
});
```
I believe the Chromium flag `--use-fake-ui-for-media-stream` has something to do with this, but not sure exactly why yet.

streamer45:
@Phrynobatrachus Thanks for having a look. Sounds like connectivity cannot be established somehow.

Can you try removing the STUN server from config, restarting the plugin and run the test again?

image

Phrynobatrachus:
Removing the STUN config worked, the test call unmutes correctly.

Phrynobatrachus:
Definitely cool with bumping the version, I'll look into it.

PR #120, opened 2022-06-20
[MM-45123] Fix fill color for disabled state of screen share icon

streamer45:
#### Summary

Fixing the fill of the screen sharing icon when button is disabled (someone else is already sharing).

#### Screenshots

**Before**

image

**After**

image

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-45123

### DIFF
--- a/webapp/src/components/call_widget/component.tsx
+++ b/webapp/src/components/call_widget/component.tsx
@@ -187,7 +187,7 @@ export default class CallWidget extends React.PureComponent<Props, State> {
                 height: '28px',
                 borderRadius: '4px',
                 color: '#D24B4E',
-                background: 'rgba(210, 75, 78, 0.08)',
+                background: 'rgba(var(--dnd-indicator-rgb), 0.08)',
                 marginRight: 'auto',
             },
             dotsMenu: {
@@ -510,7 +510,7 @@ export default class CallWidget extends React.PureComponent<Props, State> {
                         display: 'flex',
                         width: '100%',
                         height: '100%',
-                        background: 'rgba(210, 75, 78, 0.4)',
+                        background: 'rgba(var(--dnd-indicator-rgb), 0.4)',
                         justifyContent: 'center',
                         alignItems: 'center',
                         zIndex: 1001,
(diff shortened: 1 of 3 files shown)


webapp/src/components/call_widget/component.tsx
```diff
@@ -187,7 +187,7 @@ export default class CallWidget extends React.PureComponent<Props, State> {
                 height: '28px',
                 borderRadius: '4px',
                 color: '#D24B4E',
-                background: 'rgba(210, 75, 78, 0.04)',
+                background: 'rgba(210, 75, 78, 0.08)',
```
streamer45, line 190: This (leave button background) is unrelated but it wasn't matching the design although I could find differing versions. One with 8% (barely visible) and one with 24%. @abhijit-singh Could you point me to the most up to date value?

image

webapp/src/components/call_widget/component.tsx
```diff
@@ -588,6 +588,13 @@ export default class CallWidget extends React.PureComponent<Props, State> {
         const currentID = this.props.currentUserID;
         const isSharing = sharingID === currentID;
 
+        let fill = '';
+        if (isSharing) {
+            fill = 'rgba(210, 75, 78, 1)';
```
cpoile, line 593: Probably we should change this to `var(--center-channel-color)` while we're at it?  (or whatever var this rgba is)

cpoile (approved):
Nice!

webapp/src/components/call_widget/component.tsx
streamer45: Looks like that would be the `--dnd-indicator`. Let me update :+1:

webapp/src/components/call_widget/component.tsx
abhijit-singh: Ideally, it should be 8% in light themes and 24% in dark themes. But afaik that's not simple to implement so I'd recommend going with 8% right now, and then we'll work on a solution for opacities like this under a broader initiative around updating colour variables across different themes.

abhijit-singh (approved):
Updated disabled state looks good 👍

Issue #134, opened 2022-06-28
Can't listen participants


joserocha:
I can successfully start or join to a call, but none of the participants listen to each other. I tried v0.6.0 and v0.6.1

developer tool log
```
com.mattermost.calls_1e7361aa1e5244af_bundle.js:2 com.mattermost.calls: found default audio input device to use: Default - Grupo de microfones (Realtek Audio)
com.mattermost.calls_1e7361aa1e5244af_bundle.js:2 com.mattermost.calls: found default audio output device to use: Fones de ouvido / Alto falantes (Realtek Audio)
com.mattermost.calls_1e7361aa1e5244af_bundle.js:2 
        
[Deprecation] The ScriptProcessorNode is deprecated. Use AudioWorkletNode instead. (https://bit.ly/audio-worklet)
No @ com.mattermost.calls_1e7361aa1e5244af_bundle.js:2
initVAD @ com.mattermost.calls_1e7361aa1e5244af_bundle.js:2
init @ com.mattermost.calls_1e7361aa1e5244af_bundle.js:2
await in init (async)
o @ com.mattermost.calls_1e7361aa1e5244af_bundle.js:2
(anonymous) @ com.mattermost.calls_1e7361aa1e5244af_bundle.js:2
a @ call_button.tsx:56
s @ react-dom.production.min.js:14
c @ react-dom.production.min.js:14
(anonymous) @ react-dom.production.min.js:14
F @ react-dom.production.min.js:15
ot @ react-dom.production.min.js:52
at @ react-dom.production.min.js:51
rt @ react-dom.production.min.js:52
mt @ react-dom.production.min.js:56
x @ react-dom.production.min.js:287
G @ react-dom.production.min.js:19
Jt @ react-dom.production.min.js:70
Zt @ react-dom.production.min.js:69
t.unstable_runWithPriority @ scheduler.production.min.js:19
Wi @ react-dom.production.min.js:122
v @ react-dom.production.min.js:287
Vt @ react-dom.production.min.js:68
com.mattermost.calls_1e7361aa1e5244af_bundle.js:2 com.mattermost.calls: ws open, sending join msg
com.mattermost.calls_1e7361aa1e5244af_bundle.js:2 com.mattermost.calls: vad: noise avg 24 72.12962962962963
com.mattermost.calls_1e7361aa1e5244af_bundle.js:2 com.mattermost.calls: join ack received, initializing connection
[...]

PR #137, opened 2022-07-04
[MM-45499] Fix ordering to avoid race condition on session initialization

streamer45:
#### Summary

We were sending the join acknowledge event before initializing the actual call session. This could lead to a potential race condition where client signaling data gets discarded because it's received prior to initializing the rtc session.

PR reworks the ordering slightly to make sure we send the join event after creating the session.

Originally reported in https://github.com/mattermost/mattermost-plugin-calls/issues/134

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-45499

### DIFF
--- a/server/websocket.go
+++ b/server/websocket.go
@@ -382,22 +382,24 @@ func (p *Plugin) handleJoin(userID, connID, channelID, title string) error {
 		}, &model.WebsocketBroadcast{ChannelId: channelID, ReliableClusterSend: true})
 	}
 
-	// send successful join response
-	p.metrics.IncWebSocketEvent("out", "join")
-	p.API.PublishWebSocketEvent(wsEventJoin, map[string]interface{}{
-		"connID": connID,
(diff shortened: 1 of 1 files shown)


server/websocket.go
```diff
@@ -419,36 +421,49 @@ func (p *Plugin) handleJoin(userID, connID, channelID, title string) error {
...
 		}
 		if err := p.rtcdManager.Send(msg, channelID); err != nil {
-			p.LogError(fmt.Errorf("failed to send client message: %w", err).Error())
+			return fmt.Errorf("failed to send client join message: %w", err)
 		}
 	} else {
 		if handlerID == p.nodeID {
-			wg.Add(1)
-			go func() {
-				defer wg.Done()
```
streamer45, line 428: This was pure technical debt at this point since we made the initialization non-blocking a while ago so no need for a goroutine anymore. We can now easily return errors to the caller.

PR #143, opened 2022-07-12
e2e: Gitlab integration

Phrynobatrachus:
Adds `.gitlab-ci.yml` hook to internal definitions and updates setup logic to enforce `defaultenabled` plugin configuration instead of individual enabling for channels.


streamer45 (approved):
Awesome :+1:

PR #145, opened 2022-07-21
[MM-43838] WebSocket reconnection support

streamer45:
#### Summary

PR adds WebSocket reconnection support to the signaling channel by decoupling as much as possible the two connections (RTC and WS). What this means in practice is that a call can survive websocket disconnects and even server restarts (if using `rtcd`).

The main complexity involved here was state handling and consistency in all the various modes we support, namely:

- Single instance with plugin hosting calls (e.g. team edition).
- Single instance with `rtcd` hosting calls.
- HA cluster with plugin hosting calls, load balanced across the nodes.
- HA cluster with single node hosting the calls.
- HA cluster with `rtcd` hosting calls.

Before this PR, upon websocket disconnect we would cleanup everything related to a session, closing the RTC channel, and cleaning up the state. To support reconnects we need to wait instead and potentially deal with clients never coming back (through timeouts). Furthermore a client could come back on a different node than the one it was originally connected to, or the server could have restarted completely and having to setup a session from (almost) scratch.

We are adding two client events:

- `leave` This is to explicitly signal that a client is leaving a call. Upon receiving this we can cleanup everything just like we did before. Of course we still need to support the case that the event never gets in and cleanup resources appropriately.
- `reconnect` This is a new event sent by a client looking to reconnect to an existing session. Client will include the original connection id along with the previous connection id for proper state tracking.

Changes have been stress tested for a couple of weeks now to spot any potential leak, race condition or regression. I've updated the [load-test client](https://github.com/mattermost/mattermost-plugin-calls/pull/44) acconrdingly.

**Note**
This work does not yet address the potential case of sending/receiving new tracks while disconnected. To support that we either need to implement some extra queuing on the server side or rely on the existing data channel ([MM-45894](https://mattermost.atlassian.net/browse/MM-45894)).

#### Related PRs

https://github.com/mattermost/rtcd/pull/65
https://github.com/mattermost/mattermost-mobile/pull/6497

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-43838

### DIFF
--- a/server/plugin.go
+++ b/server/plugin.go
@@ -133,16 +133,12 @@ func (p *Plugin) handleEvent(ev model.PluginClusterEvent) error {
 		go p.startSession(us, msg.SenderID)
 		return nil
 	case clusterMessageTypeReconnect:
-		p.LogDebug("reconnect event", "UserID", msg.UserID, "ConnID", msg.ConnID, "PrevConnID", msg.PrevConnID)
+		p.LogDebug("reconnect event", "UserID", msg.UserID, "ConnID", msg.ConnID)
 
 		p.mut.Lock()
 		defer p.mut.Unlock()
 
 		us := p.sessions[msg.ConnID]
-		if us == nil && msg.PrevConnID != "" {
-			us = p.sessions[msg.PrevConnID]
-		}
-
 		if us == nil {
 			return nil
 		}
--- a/server/plugin.go
+++ b/server/plugin.go
@@ -133,7 +133,7 @@ func (p *Plugin) handleEvent(ev model.PluginClusterEvent) error {
 		go p.startSession(us, msg.SenderID)
 		return nil
 	case clusterMessageTypeReconnect:
-		p.LogDebug("reconnect event", "ChannelID", msg.ChannelID, "UserID", msg.UserID, "ConnID", msg.ConnID, "PrevConnID", msg.PrevConnID)
+		p.LogDebug("reconnect event", "UserID", msg.UserID, "ConnID", msg.ConnID, "PrevConnID", msg.PrevConnID)
 
 		p.mut.Lock()
 		defer p.mut.Unlock()
@@ -159,7 +159,7 @@ func (p *Plugin) handleEvent(ev model.PluginClusterEvent) error {
 
 		return nil
 	case clusterMessageTypeLeave:
-		p.LogDebug("leave event", "ChannelID", msg.ChannelID, "UserID", msg.UserID, "ConnID", msg.ConnID)
+		p.LogDebug("leave event", "UserID", msg.UserID, "ConnID", msg.ConnID)
 
 		p.mut.RLock()
 		us := p.sessions[msg.ConnID]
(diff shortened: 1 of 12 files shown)


webapp/src/websocket.ts
```diff
@@ -35,22 +64,37 @@ export default class WebSocketClient extends EventEmitter {
...
+                    this.emit('open', this.originalConnID, this.connID, false);
+                }
                 return;
             } else if (!this.connID) {
                 logWarn('ws message received while waiting for hello');
                 return;
             }
 
-            if (msg.data.connID !== this.connID) {
+            if (msg.data.connID !== this.connID && msg.data.connID !== this.originalConnID) {
```
cpoile, line 97: If I understand correctly, this means we're allowing the case where there are two valid connection ids for this client?

server/plugin.go
```diff
@@ -113,36 +118,76 @@ func (p *Plugin) handleEvent(ev model.PluginClusterEvent) error {
...
+		defer p.mut.Unlock()
+
+		us := p.sessions[msg.ConnID]
+		if us == nil {
+			return nil
+		}
+
+		if atomic.CompareAndSwapInt32(&us.wsReconnected, 0, 1) {
+			p.LogDebug("closing reconnectCh", "connID", msg.ConnID)
+			close(us.wsReconnectCh)
```
cpoile, line 148: cool -- had to follow this through, makes sense now.

cpoile (approved):
Wow, nice work! Just minor comments, nothing blocking. Thanks!

webapp/src/client.ts
```diff
@@ -48,6 +50,11 @@ export default class CallsClient extends EventEmitter {
         this.onDeviceChange = () => {
             this.updateDevices();
         };
+        this.onBeforeUnload = () => {
+            logDebug('unload');
+            this.disconnect();
+        };
+        window.addEventListener('beforeunload', this.onBeforeUnload);
```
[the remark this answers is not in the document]
streamer45: Technically if you refresh the page (or any other similar action) the client would completely disconnect anyway. This is really about cleaning up before that happens (sending the leave message) to avoid having to timeout connections on the server.

webapp/src/websocket.ts
streamer45: This should really go away soon as we now support a way to send per connection ws events but for now, depending on which node a client is connected to and from where messages are coming from it may be either the original connection ID or the current one.

webapp/src/websocket.ts
```diff
@@ -35,22 +64,37 @@ export default class WebSocketClient extends EventEmitter {
                 msg = JSON.parse(data);
             } catch (err) {
                 logErr('ws msg parse error', err);
+                return;
+            }
+
+            if (msg) {
```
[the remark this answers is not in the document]
streamer45: I am thinking we would potentially lose some messages that way as we need to keep the counter up to date even if we don't process them.

server/plugin.go
streamer45: Unfortunately overall logic is not very simple but there's a lot to consider there and had to cover for potential misuses as well (e.g. clients sending multiple/out of order events). Happy to answer anything that's not clear.

PR #148, opened 2022-07-26
e2e: Gitlab -> Github status reporting

Phrynobatrachus:
Minor config update to support reporting statuses back to Github. Examples can be seen in the recent commit checks on this branch.

Screenshot from 2022-07-26 12-07-52

Screenshot from 2022-07-26 12-04-40

Also a note for the moment, mirroring from Github -> Gitlab currently happens every 30 minutes so the new check won't always appear immediately. This will change after [DOPS-1131](https://mattermost.atlassian.net/browse/DOPS-1131) is handled.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,4 +1,4 @@
 include:
   - project: qa/calls-e2e-testing
-    ref: main
+    ref: gh-report-status
     file: private.yml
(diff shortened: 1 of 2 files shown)


streamer45 (approved):
Great stuff :+1: Love that the check has your avatar :grinning:

Phrynobatrachus:
Ha yeah I'm going to ask release team or someone about using a Mattermod token instead.

PR #153, opened 2022-08-01
Replace `simple-peer` with `RTCPeer`

streamer45:
#### Summary

Some recent issues (e.g. [MM-45977](https://mattermost.atlassian.net/browse/MM-45977)) surfaced a need to have more control on lower level APIs of WebRTC connections. Moreover, future work (e.g. implementing simulcast) will pose similar needs as well .

I decided to spend part of last Open Source Friday to replace `simple-peer` with our own version of a WebRTC connection wrapper (which I should have probably named simpler-peer :p). Pun aside, `simple-peer` let us focus on delivering calls in a very efficient way without worrying too much about anything low level on the client side of which I am very grateful :raised_hands: . 

That said, I feel this is the right time to move on, especially with mobile v2 support going live soon. What this means, in practice:

- Lighter implementation (~150 [SLOC](https://en.wikipedia.org/wiki/Source_lines_of_code)) with only the strictly necessary to retain functionality and no unnecessary hack.
- Direct, low level control over the WebRTC API.
- Unified implementation across Web and Mobile. This is mostly doable today with a couple of *injections*. We may consider extracting this into a standalone module/npm package (tbd).
- 100% Typescript.

### DIFF
--- a/webapp/package-lock.json
+++ b/webapp/package-lock.json
@@ -14,7 +14,6 @@
         "react-redux": "7.2.4",
         "redux": "4.1.0",
         "semver-parser": "4.0.0",
-        "simple-peer": "9.11.0",
         "styled-components": "5.3.5",
         "typescript": "4.3.4"
       },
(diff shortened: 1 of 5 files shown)


cpoile (approved):
Awesome, thank you!

streamer45:
Going to merge this now so we can get some good soaking time on community.

PR #154, opened 2022-08-01
[MM-46107] Fix audio resources leaks

streamer45:
#### Summary

This was an interesting one. It started with just some hard to explain high CPU usage in Firefox (I usually test stuff locally on a Chrome + Firefox setup). Then in the last few days I was getting random issues with my audio service (`pulseaudio`) not being able to playback. Closing the Chrome tab I was testing Calls on temporarily fixed this odd behaviour. 

But it was only today, after failing once again to hear people due to a playback failure that I realized I was leaking dozens of audio resources that were still in the playback state even long after ending calls. 

PR reviews and fixes any outstanding leak of audio resources.

@lieut-data I just couldn't believe that it was simply a Linux quirk, so dug further, thank you :grin: 

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-46107

### DIFF
--- a/webapp/src/components/call_widget/component.tsx
+++ b/webapp/src/components/call_widget/component.tsx
@@ -281,6 +281,7 @@ export default class CallWidget extends React.PureComponent<Props, State> {
 
             document.body.appendChild(audioEl);
             voiceTrack.onended = () => {
+                audioEl.srcObject = null;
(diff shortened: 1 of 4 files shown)


cpoile (approved):
Wow, nice!

PR #155, opened 2022-08-03
[MM-46124] Improve session deinitialization

streamer45:
#### Summary

Two main changes here to avoid leaking the `handleJoin` goroutine:

- Handling the new `ClientMessageClose` event that gets sent by `rtcd`.
- Exiting from `wsReader`/`handleLeave` `select` blocks when the `rtcCloseCh` gets closed.

It's important to notice that this PR does not fix the race but simply makes it much more unlikely to cause a leak. This is because the logic is non trivial and serializing everything using a mutex feels like an excessive constraint just to cover a rare edge case so we allow this to happen but react accordingly.

#### Related PRs

https://github.com/mattermost/rtcd/pull/68

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-46124

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -15,7 +15,7 @@ require (
 require (
 	github.com/mattermost/logr/v2 v2.0.15
 	github.com/mattermost/mattermost-plugin-api v0.0.27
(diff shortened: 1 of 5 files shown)


cpoile (approved):
👍

PR #164, opened 2022-08-15
[MM-42395] Extract call duration to avoid unnecessary parent component re-render

streamer45:
#### Summary

This was a low hanging fruit especially as we are trying to optimize the widget to be as performance as possible.

**Before**
before png
**After**
after png

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-42395


cpoile (approved):
Awesome!

PR #166, opened 2022-08-17
Relay `devicechange` event to properly sync devices

streamer45:
#### Summary

This is probably the way it should have been implemented since the start. We relay the `devicechange` event to the component to keep everything in sync and avoid having to constantly poll. This has the added bonus that the menu will update even when fully open and visible.

### DIFF
--- a/webapp/src/client.ts
+++ b/webapp/src/client.ts
@@ -3,7 +3,7 @@ import {EventEmitter} from 'events';
 // @ts-ignore
 import {deflate} from 'pako/lib/deflate.js';
 
-import {CallsClientConfig, RTCStats} from 'src/types/types';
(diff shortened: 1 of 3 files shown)


webapp/src/client.ts
```diff
@@ -27,7 +27,7 @@ export default class CallsClient extends EventEmitter {
     private voiceTrackAdded: boolean;
     private streams: MediaStream[];
     private stream: MediaStream | null;
-    private audioDevices: { inputs: MediaDeviceInfo[]; outputs: MediaDeviceInfo[]; };
+    private audioDevices: AudioDevices;
```
cpoile, line 30: 👍

cpoile (approved):
Nicer, thanks!

PR #168, opened 2022-08-18
[MM-45025] Keyboard shortcuts

streamer45:
#### Summary

PR adds initial support for Calls specific keyboard shortcuts. Implementation wise I tried to keep things as simple as possible while allowing for a future where the mappings could be coming from an API and be potentially user defined through preferences.

I tested these on Linux both from web and desktop and partially on macOS as well. Given key codes may be affected by platform/os/browser/layout/language I'd encourage everyone to give it a try and report in case you encounter any issue.

A couple of high level considerations from testing this:

1. I know this has been discussed briefly but should we plan ways to advertise these shortcuts to our users? (either through UI, documentation or else).
2. It feels like a shortcut for opening the pop-out window would be particularly beneficial as well.
3. Do we want both shortcuts for toggling the participants list or only one of them?

/cc @abhijit-singh @itao for visibility

#### Design Link

https://www.figma.com/file/M4hDB8ndPo9muH8t75rpcy/Icons-%26-Shortcuts

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-45025


cpoile (approved):
Awesome! Tested and works for me on mac in browser and on desktop, in widget and expanded views. 

Thoughts on your questions:
1. At the least we should probably have the shortcut shown in the hover text. What do you think?
2. I second this -- when testing I thought "wow, I want a shortcut to expand and collapse the view." Especially since I toggle between those pretty frequently to read threads during a call.

Phrynobatrachus (approved):
Linux + Firefox quirks:

`Ctrl + Shift + p` - private browsing session hotkey appears to take precedence, does not open or close participants list. `Alt + p` works though. 

`Ctrl + Shift + e` - works for ending screen share, but doesn't prompt to start.

1. +1 for hover text, also might be nice to have in the `Ctrl + /` shortcut modal.
2. +1 would use
3. See above.

I'll definitely be happy to have mute/unmute, thanks for adding these!

streamer45:
Thanks @Phrynobatrachus!

A couple of comments:

- <kbd>CTRL</kbd> + <kbd>Shift</kbd> + <kbd>P</kbd> not working seems like a Firefox known thing as they don't allow native hotkeys to be overridden from JS (https://bugzilla.mozilla.org/show_bug.cgi?id=380637 and https://bugzilla.mozilla.org/show_bug.cgi?id=1052569 for reference). This means <kbd>Alt</kbd>+<kbd>P</kbd> is likely the more supported option.
- <kbd>CTRK</kbd> + <kbd>Shift</kbd> + <kbd>E</kbd> I think you may have hit a security error that prevents the browser from sharing the screen until an explicit user gesture has been performed on the page. It should work once you have interacted with the page (e.g. clicked). Unfortunately this is not something we can change but something to keep in mind.

tanmay-des (commented):
The shortcut seems to work well. Few things:

1. For the mute/unmute toggle by space. I was wondering if we can make it _hold to speak_ instead of toggle (just like in zoom) thoughts ? cc @itao @abhijit-singh 

2. I definitely agree with claudio, we should put up tooltips for showcasing the shortcut. I'll quickly add a figma link for that. In addition I was also thinking if we should prompt the tooltips/toast when a person uses a shortcut (In fullscreen mode) I'll add an example for this as well

streamer45:
Thanks @tanmay-des, I am happy to proceed with implementation as soon as the updated designs are ready.

webapp/src/components/shortcut.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+    const renderKeys = () => {
+        return keys.map((key) => {
+            let ch = '';
+            if (isMac()) {
+                ch = macKeyToCharMap[key];
+            }
+            if (!ch) {
+                ch = key;
+            }
```
cpoile, line 26: minor nit: would it be cleaner to do this?
```suggestion
        return keys.map((key) => {
            let ch = key;
            if (isMac()) {
                ch = macKeyToCharMap[key];
            }
```

webapp/src/shortcuts.ts
```diff
@@ -29,6 +31,29 @@ export const keyMappings: {[key: string]: {[key: string]: string}} = {
     popout: popoutMappings,
 };
 
+type reverseKeyMap = {
+    [key: string]: {
+        [key: string]: string[],
+    },
+};
+
+export const reverseKeyMappings = (() => {
```
cpoile, line 40: nice

cpoile (approved):
Awesome!

webapp/src/components/shortcut.tsx
streamer45: It would be cleaner but semantic would be different as `macKeyToCharMap` is expected to potentially return undefined if the char is not one of those mapped (only those two atm) in which case we want to fallback. So we would have to at least check that in addition.

itao (approved):

PR #169, opened 2022-08-18
loadtest: Simulate user interactions

cpoile:
#### Summary
- First cut of a simulation of some user interactions
- There are only two config options at the moment. Can be improved probably.
- There's more simulation I'd love to add (like users changing their pictures and names and teammate displayname setting), but cutting it short here because it does what we need it to: makes a good UI load tester for all our frontends. 
- A couple interesting gotchas that took some time:
  - the lt/client doesn't work if you are using the plugin only. It seems like you need to use the rtcd
  - the command line flags are really wonky. If you try `go run ./lt/*.go -admin-username chris -admin-password Testtest1! -team 4ch15noyx3grud6jwuz3of7p9h -channel 3obhsuzyi7nu8ez3qji5r93tca -join-duration 4s -duration 600s -sim true -sim-interval 5s -users-per-call 20` it won't read the last flag. But if you use: `go run ./lt/*.go -admin-username chris -admin-password Testtest1! -team 4ch15noyx3grud6jwuz3of7p9h -channel 3obhsuzyi7nu8ez3qji5r93tca -join-duration 4s -duration 600s -sim-interval 5s -users-per-call 20 -sim true` it does. This took a lot of head scratching and debugging to figure out what was going on. Maybe it's just mac? Maybe it's just my machine?

#### Ticket Link
- none


cpoile:
Ahhh... bool flags need to use the `=` sign. 🤕

lt/client.go
```diff
@@ -626,6 +639,13 @@ func (u *user) Connect(stopCh chan struct{}, channelType model.ChannelType) erro
 	}
 
 	log.Printf("%s: disconnecting...", u.cfg.username)
+
+	defer func() {
+		if err := recover(); err != nil {
+			log.Printf("%s: caught a panic, probably closing an already closed channel. That's expected bc of the sim. Err: %v\n",
+				u.cfg.username, err)
+		}
+	}()
```
streamer45, line 648: Not a huge fan of `recover` unless it's unfeasible to avoid it. I think it may be time to deepen a bit the abstraction here and consider decoupling the highest level, load-test user state from the calls client which can become its own type/package.

streamer45:
> the lt/client doesn't work if you are using the plugin only. It seems like you need to use the rtcd

It should work with the plugin just fine. I've hit a couple of connection failures in the past which required to set the ICE host override (or disabling STUN) but otherwise there's nothing preventing this from working in that mode. Let me know if you need help debugging but I am thinking it's a simple RTC connection failure (likely STUN involved and clients trying to go through your public IP).

streamer45 (commented):
Thanks @cpoile , looks great.

Just left a couple of comments which can be addressed at any time really as nothing is urgent here.

Also wanted to share a bit more of my vision for this work: my goal with this branch and the main reason it was never merged back is that I was seriously considering plugging this into [`mattermost-load-test-ng`](https://github.com/mattermost/mattermost-load-test-ng) at some point so that it could be added on top of the "suite" simulation.

Now that we are starting to add more advanced logic we are hitting the point where it would make sense to do this transition, mostly to avoid diverging too much. If you agree we could open a ticket to plan this work.

agnivade:
> Now that we are starting to add more advanced logic we are hitting the point where it would make sense to do this transition, mostly to avoid diverging too much. If you agree we could open a ticket to plan this work.

Please involve @agarciamontoro as well in such discussions. Our aim is to let different products keep their load testing code in their own repo to keep related things close, and make the load test repo pull in and run the activities. But we still haven't discussed the feasibility of that plan.

Alejandro is currently working on adding playbooks coverage, so this might be a good time to discuss calls integration as well.

streamer45:
@agnivade I don't mind where the code lives as long as it's easily usable by the main framework either through a common interface or else. Happy to coordinate if/when the time comes, just thinking out loud at this point.

agarciamontoro:
Also thinking out loud at this point:

> I don't mind where the code lives as long as it's easily usable by the main framework either through a common interface or else

The reason behind moving the load-test code specific to products to their corresponding repos (playbooks, focalboard, calls) is to encourage product teams to own the load-testing efforts. It's more a cultural measure (each team knows they own the code in their repo) than a technical one. Of course, that alone won't be enough, but I think it can help.

In any case, happy to listen to any comments or ideas about that :)

cpoile:
Closing for now, will pick up later when needed.

PR #170, opened 2022-08-22
[MM-46501] Fix connect event not firing

streamer45:
#### Summary

Yet another browser inconsistency. Today I learned that Firefox does not implement the [`RTCPeerConnection.onconnectionstatechange`](https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/connectionstatechange_event#browser_compatibility) so the event was never firing. Switching to use the more widely adopted `oniceconnectionstatechange` instead.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-46501

### DIFF
--- a/webapp/src/rtcpeer/index.ts
+++ b/webapp/src/rtcpeer/index.ts
@@ -18,7 +18,7 @@ export default class RTCPeer extends EventEmitter {
         this.pc = new RTCPeerConnection(config);
         this.pc.onnegotiationneeded = () => this.onNegotiationNeeded();
         this.pc.onicecandidate = (ev) => this.onICECandidate(ev);
-        this.pc.onconnectionstatechange = () => this.onConnectionStateChange();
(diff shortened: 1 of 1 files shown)


webapp/src/rtcpeer/index.ts
```diff
@@ -33,14 +33,11 @@ export default class RTCPeer extends EventEmitter {
...
-    private onConnectionStateChange() {
-        switch (this.pc?.connectionState) {
+    private onICEConnectionStateChange() {
+        switch (this.pc?.iceConnectionState) {
         case 'connected':
             this.emit('connect');
             break;
-        case 'disconnected':
-            this.emit('disconnect');
-            break;
```
streamer45, line 43: Removing this for the time being since we are not making use of it. Should come back if/when we start to support ICE reconnects.

cpoile (approved):

PR #171, opened 2022-08-22
[MM-46412] Disable calls in archived channels

streamer45:
#### Summary

PR disables calls in archived channels. While technically nothing was preventing them up until now, the client would error out.
We are now disabling the Start/Join button and providing some information on hover.

What's potentially left to handle is the automatically joining through URL case. With the error fixed nothing would happen upon visiting the URL. If we'd like to be more informative we may consider opening up a modal or give some other kind of feedback to the user. Let me know your thoughts.

/cc @itao 

#### Screenshots

image

image

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-46412

### DIFF
--- a/server/websocket.go
+++ b/server/websocket.go
@@ -393,6 +393,9 @@ func (p *Plugin) handleJoin(userID, connID, channelID, title string) error {
 	if appErr != nil {
 		return appErr
 	}
+	if channel.DeleteAt > 0 {
+		return fmt.Errorf("cannot join call in archived channel")
+	}
 
 	state, err := p.addUserSession(userID, connID, channel)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


cpoile (approved):
Looks good, but we could also not show the join button at all in archived channels. Maybe a UX call? @tanmay-des

streamer45:
> Looks good, but we could also not show the join button at all in archived channels. Maybe a UX call? @tanmay-des

Indeed. I implemented it this way to be as informative as possible, especially if the user is coming from the join URL, not seeing the button may be confusing. At least we are giving them a way to understand why calling is possible. But open to suggestions from UX of course.

tanmay-des (approved):
Looks good to me! 
Sorry for the delay, this one kindoff slipped under my radar.

PR #173, opened 2022-08-22
[MM-46099] Return error on joining call through slash command if calls are disabled

streamer45:
#### Summary

PR adds an error message in case `/call start` or `/call join` slash commands are sent in a channel where calls are not enabled. As an added bonus this PR led to some small refactor on the redux side.

#### Screenshot

image

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-46099

### DIFF
--- a/webapp/src/components/channel_header_dropdown_button/index.ts
+++ b/webapp/src/components/channel_header_dropdown_button/index.ts
@@ -6,7 +6,7 @@ import {getCurrentChannelId} from 'mattermost-redux/selectors/entities/channels'
 import {
     voiceConnectedUsers,
     connectedChannelID,
-    channelState,
+    callsEnabled,
     isCloudFeatureRestricted,
     isCloudProfessionalOrEnterprise,
     isLimitRestricted,
@@ -16,7 +16,7 @@ import {
 import ChannelHeaderDropdownButton from './component';
 
 const mapStateToProps = (state: GlobalState) => ({
-    show: Boolean(channelState(state, getCurrentChannelId(state))?.enabled),
+    show: callsEnabled(state, getCurrentChannelId(state)),
     inCall: Boolean(connectedChannelID(state) && connectedChannelID(state) === getCurrentChannelId(state)),
     hasCall: voiceConnectedUsers(state).length > 0,
     isCloudFeatureRestricted: isCloudFeatureRestricted(state),
--- a/webapp/src/components/channel_header_dropdown_button/index.ts
+++ b/webapp/src/components/channel_header_dropdown_button/index.ts
@@ -6,7 +6,7 @@ import {getCurrentChannelId} from 'mattermost-redux/selectors/entities/channels'
 import {
     voiceConnectedUsers,
     connectedChannelID,
-    isVoiceEnabled,
+    channelState,
     isCloudFeatureRestricted,
     isCloudProfessionalOrEnterprise,
     isLimitRestricted,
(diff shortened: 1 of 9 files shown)


Phrynobatrachus:
I fixed another issue with the status report, the previous two commits should've been red. Looking at the failed E2E tests themselves now.

streamer45:
Nice, thanks @Phrynobatrachus for taking care of this :raised_hands:

webapp/src/reducers.ts
```diff
@@ -34,16 +32,25 @@ import {
...
+    id: string,
+    enabled: boolean,
+}
+
+interface channelStateAction {
+    type: string,
+    data: channelState,
+}
+
+const channelState = (state: {[channelID: string]: channelState} = {}, action: channelStateAction) => {
```
cpoile, line 48: An alternative would be:
```suggestion
const channelState = (state: Record<string, channelState> = {}, action: channelStateAction) => {
```
But I kind of like that we're saying the key is a `channelID`... 👍

webapp/src/reducers.ts
```diff
@@ -34,16 +32,25 @@ import {
...
     SHOW_END_CALL_MODAL,
     HIDE_END_CALL_MODAL,
+    RECEIVED_CHANNEL_STATE,
 } from './action_types';
 
-const isVoiceEnabled = (state = false, action: { type: string }) => {
+interface channelState {
+    id: string,
+    enabled: boolean,
+}
```
cpoile, line 41: I wonder why we're making this a channelState now, is it in anticipation for future details?
Because all we really need to store is `enabled` since the channelID is the index...

webapp/src/selectors.ts
```diff
@@ -205,3 +204,5 @@ export const isCloudTrialNeverStarted: (state: GlobalState) => boolean = createS
         return subscription?.trial_end_at === 0;
     },
 );
+
+export const channelState = (state: GlobalState, channelID: string) => getPluginState(state).channelState[channelID];
```
cpoile, line 208: I wonder if for this we should still offer a `isVoiceEnabled(state: GlobalState, channelID: string)` selector, so that it would simplify where you've used `channelState` above. (I'll make a comment there in a sec)

webapp/src/components/channel_header_dropdown_button/index.ts
```diff
@@ -16,7 +16,7 @@ import {
 import ChannelHeaderDropdownButton from './component';
 
 const mapStateToProps = (state: GlobalState) => ({
-    show: isVoiceEnabled(state),
+    show: Boolean(channelState(state, getCurrentChannelId(state))?.enabled),
```
cpoile, line 19: re: the comment I just left about making a `isVoiceEnabled` selector. That would simplify this to:
```suggestion
    show: isVoiceEnabled(state, getCurrentChannelId(state)),
```
And in the other places you've used it.

cpoile (approved):
Cool! Just some thoughts about simplifying the selectors and reducer

webapp/src/reducers.ts
streamer45: I am just trying to bring back some consistency with the server-side API and also start moving into the direction of receiving the whole state at once to avoid races. At some point this could hold the call state as well. Of course we are still planning a proper refactor around all of this.

As far as the ID, it's totally redundant but in my experience it's worth having it because the index is metadata that could go away as soon as we change how the actual data is stored but the data/type has an id. If we were worried about memory I would remove it but I don't think that's a concern at this time.

webapp/src/selectors.ts
streamer45: Sure, I don't mind a further simplification as long as we pass the channelID.

webapp/src/reducers.ts
streamer45: What do you think of deferring this type of change to a dedicated ticket where we consistently make use of it?

cpoile (approved):

webapp/src/reducers.ts
cpoile: Sounds good. But to be clear: I think using `{[channelID: string]: channelState}` is fine, and maybe even better than `Record` because it tells us the index is a channelID.

PR #174, opened 2022-08-25
[MM-45128] Disable join sounds after number of participants threshold

streamer45:
#### Summary

PR disables the "user has joined the call" sound after reaching a threshold of 8 participants. 

Since we are looking to make this configurable at some point in the future, I opted not to check against a simple constant but add the necessary boilerplate to support the more advanced use case.

/cc @tanmay-des @itao 

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-45128

### DIFF
--- a/webapp/src/selectors.ts
+++ b/webapp/src/selectors.ts
@@ -8,7 +8,7 @@ import {LicenseSkus} from '@mattermost/types/general';
 
 import {isDMChannel} from 'src/utils';
 
-import {CallsConfig} from 'src/types/types';
+import {CallsConfig, CallsUserPreferences} from 'src/types/types';
 
 import {pluginId} from './manifest';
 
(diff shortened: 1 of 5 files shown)


webapp/src/selectors.ts
```diff
@@ -42,6 +42,13 @@ export const voiceConnectedUsersInChannel = (state: GlobalState, channelID: stri
 
 export const connectedChannelID = (state: GlobalState) => getPluginState(state).connectedChannelID;
 
+const numUsersInConnectedChannel: (state: GlobalState) => number = createSelector(
+    'numUsersInConnectedChannel',
+    connectedChannelID,
+    voiceConnectedChannels,
+    (channelID, channels) => channels ? channels[channelID]?.length || 0 : 0,
```
cpoile, line 49: since `voiceConnectedChannels` is initialized as an empty object, I think we can simplify this to:
```suggestion
    (channelID, channels) => channels[channelID]?.length || 0,
```
If this is correct, at some point we could probably clean up the rest of `selectors.ts`.  (or a HW ticket)

cpoile (approved):
Nice!

webapp/src/selectors.ts
streamer45: Yeah, just being extra careful I guess :)

PR #178, opened 2022-09-01
Fix `/call stats` payload

streamer45:
#### Summary

Not exactly sure what's going on since command is working fine locally but can't get it to work on Community. I believe it's due to the quotes.


cpoile (approved):
Huh, interesting. Let's try it.

PR #182, opened 2022-09-06
MM-46898 - Fix for: Cloud trial does not unlock calls in channels

cpoile:
#### Summary
- Seems like we weren't treating a cloud trial like a fully paid version of the product. (Ticket shows what was happening.)
- Need to test this on a CWS cloud test server

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-46898

### DIFF
--- a/server/cloud_limits.go
+++ b/server/cloud_limits.go
@@ -124,3 +124,7 @@ func isCloud(license *model.License) bool {
 func isCloudStarter(license *model.License) bool {
 	return license != nil && license.SkuShortName == "starter"
 }
+
+func isTrial(license *model.License) bool {
+	return license != nil && license.IsTrial
+}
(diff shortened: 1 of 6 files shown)


server/cloud_limits.go
```diff
@@ -124,3 +124,7 @@ func isCloud(license *model.License) bool {
 func isCloudStarter(license *model.License) bool {
 	return license != nil && license.SkuShortName == "starter"
 }
+
+func isTrial(license *model.License) bool {
+	return license != nil && license.IsTrial
```
streamer45, line 129: Is there a way to distinguish what type of trial it is or is it always enterprise somehow?

streamer45 (approved):
👍

server/cloud_limits.go
cpoile: Not that I can tell. It's is an assumption I'm making -- that being a trial means it's treated as enterprise. That's seems to be what they're doing in the webapp, but it's not 1:1 because they're using webapp data, and here we're using server data.

cpoile:
Was able to confirm that it works in https://github.com/mattermost/mattermost-server/pull/20971

Phrynobatrachus (approved):

cpoile:
Hah, oops, should have waited for your approval, sorry @Phrynobatrachus :)

Phrynobatrachus:
S'all gooood. :call_me_hand:

PR #185, opened 2022-09-08
[MM-46807] Implement client-side telemetry

streamer45:
#### Summary

PR implements client side telemetry through a new `/telemetry/track` API endpoint. This doesn't instrument any path yet but merely provides the generic functionality to track events. Will send a PR that builds on top of this one for that.

To be used as such:

```js
trackEvent('user_toggle_participants_list', 'widget');
```

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-46807

### DIFF
--- a/server/api.go
+++ b/server/api.go
@@ -6,7 +6,6 @@ package main
 import (
 	"encoding/json"
 	"fmt"
-	"io/ioutil"
 	"net/http"
 	"net/http/pprof"
 	"regexp"
@@ -24,6 +23,8 @@ import (
 var chRE = regexp.MustCompile(`^\/([a-z0-9]+)$`)
 var callEndRE = regexp.MustCompile(`^\/calls\/([a-z0-9]+)\/end$`)
 
+const requestBodyMaxSizeBytes = 1024 * 1024 // 1MB
+
 type Call struct {
 	ID              string      `json:"id"`
 	StartAt         int64       `json:"start_at"`
@@ -35,7 +36,7 @@ type Call struct {
 }
 
 type ChannelState struct {
-	ChannelID string `json:"channel_id"`
+	ChannelID string `json:"channel_id,omitempty"`
 	Enabled   bool   `json:"enabled"`
 	Call      *Call  `json:"call,omitempty"`
 }
@@ -317,15 +318,8 @@ func (p *Plugin) handlePostChannel(w http.ResponseWriter, r *http.Request, chann
 		}
 	}
 
-	data, err := ioutil.ReadAll(r.Body)
-	if err != nil {
-		res.Err = err.Error()
-		res.Code = http.StatusInternalServerError
-		return
-	}
-
 	var info ChannelState
-	if err := json.Unmarshal(data, &info); err != nil {
+	if err := json.NewDecoder(http.MaxBytesReader(w, r.Body, requestBodyMaxSizeBytes)).Decode(&info); err != nil {
 		res.Err = err.Error()
 		res.Code = http.StatusBadRequest
 		return
@@ -353,7 +347,7 @@ func (p *Plugin) handlePostChannel(w http.ResponseWriter, r *http.Request, chann
 
 	p.API.PublishWebSocketEvent(evType, nil, &model.WebsocketBroadcast{ChannelId: channelID, ReliableClusterSend: true})
 
-	if _, err := w.Write(data); err != nil {
+	if err := json.NewEncoder(w).Encode(info); err != nil {
 		p.LogError(err.Error())
 	}
 }
@@ -531,6 +525,11 @@ func (p *Plugin) ServeHTTP(c *plugin.Context, w http.ResponseWriter, r *http.Req
 
 		if matches := callEndRE.FindStringSubmatch(r.URL.Path); len(matches) == 2 {
 			p.handleEndCall(w, r, matches[1])
+			return
+		}
+
+		if r.URL.Path == "/telemetry/track" {
+			p.handleTrackEvent(w, r)
 			return
 		}
 	}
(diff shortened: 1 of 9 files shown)


server/api.go
```diff
@@ -317,15 +318,8 @@ func (p *Plugin) handlePostChannel(w http.ResponseWriter, r *http.Request, chann
...
-	data, err := ioutil.ReadAll(r.Body)
-	if err != nil {
-		res.Err = err.Error()
-		res.Code = http.StatusInternalServerError
-		return
-	}
-
 	var info ChannelState
-	if err := json.Unmarshal(data, &info); err != nil {
+	if err := json.NewDecoder(http.MaxBytesReader(w, r.Body, requestBodyMaxSizeBytes)).Decode(&info); err != nil {
```
streamer45, line 322: This is unrelated to the PR but something we should definitely fix so doing it here.

server/telemetry.go
```diff
@@ -68,3 +94,53 @@ func (p *Plugin) initTelemetry(enableDiagnostics *bool) error {
...
+
+	if data.Props == nil {
+		data.Props = map[string]interface{}{}
+	}
+
+	if data.Source != "" {
+		data.Props["Source"] = data.Source
+	}
+
+	data.Props["ActualUserID"] = r.Header.Get("Mattermost-User-Id")
```
cpoile, line 139: Just for my own interest, is this what we're supposed to do? (In terms of privacy)

server/telemetry.go
```diff
@@ -68,3 +94,53 @@ func (p *Plugin) initTelemetry(enableDiagnostics *bool) error {
...
+		res.Err = "invalid telemetry event"
+		res.Code = http.StatusBadRequest
+		return
+	}
+
+	if _, ok := telemetryClientTypes[data.ClientType]; !ok {
+		res.Err = "invalid client type"
+		res.Code = http.StatusBadRequest
+		return
+	}
```
cpoile, line 129: I can maybe understand making sure the client type is correct, but I wonder if the event types needs to be strictly checked like this? Why not just send it through? Then there's one less place we need to update when we add an event.

cpoile (approved):
Awesome. Non-blocking comment and question.

server/telemetry.go
streamer45: Cannot speak in terms of privacy but we've been adding user ids since forever. Every time you switch a team (or do any other tracked action) you are sending a `user_actual_id` in the payload. Generally we are treating ids as fully opaque values, in fact they are not considered a concern from a purely security perspective.

In this specific case I find it important to track since it's the only piece of authenticated information. Everything else is coming from the client and potentially untrustable.

server/telemetry.go
streamer45: Funnily I've even considered strong typing the whole props set but it was just too hard to lose the flexibility as I couldn't possibly predict what could be useful there.

As for event types I still think the trade-off is worth it as I prefer having a single source of truth on the server that explicitly enumerates what's possible to send through the API. I've actually refactored a bit to make it slightly easier and less error prone.

streamer45:
@tanmay-des Just waiting for the list of events we'd like to track. Once you have that ready I'll instrument the code and merge, thanks.

tanmay-des (commented):
Here is the list of the actions I think we need some data against. What do you think @streamer45 

1. Opening expanded mode
2. Toggling participant list (In both expanded and widget mode)
3. Starting share screen (In both expanded vs widget)
4. Ending share screen (widget vs expanded)
5. Raise hand 
6. Channel link

streamer45:
Thanks @tanmay-des, looks good, just a couple of questions:

- At which level would you like to track these actions? For example, should we track key-board initiated actions as well or buttons only?
- What's the interest in knowing when someone stops sharing? I assume if a user starts sharing they eventually will stop sharing or leave the call before doing it. Same logic applies to any toggling actions really, like opening/closing participants list or raising/lowering hand.

tanmay-des:
1. Yes if possible we'd like to register both key-bind as well as UI button initiated actions. 
2. As we have two ways to stop screen sharing i.e both in widget mode and expanded mode with different treatment. I want to know if users do switch out the call mode after sharing screen. 

cc @streamer45

streamer45:
@cpoile Asking for a second pass, specifically for https://github.com/mattermost/mattermost-plugin-calls/pull/185/commits/11bad430f0f71b6058851cc75f45f91b7b2d0180

server/telemetry.go
```diff
@@ -23,7 +23,13 @@ var (
...
 	telemetryClientEvents = []string{
 		"user_open_expanded_view",
-		"user_toggle_participants_list",
+		"user_open_participants_list",
+		"user_close_participants_list",
+		"user_share_screen",
+		"user_unshare_screen",
+		"user_raise_hand",
+		"user_lower_hand",
+		"user_open_channel_link",
```
cpoile, line 32: would we also want `user_close_expanded_view`, just to be complete?

webapp/.eslintrc.json
```diff
@@ -333,12 +333,8 @@
...
     "no-self-compare": 2,
     "no-sequences": 2,
-    "no-shadow": [
-      2,
-      {
-        "hoist": "functions"
-      }
-    ],
+    "no-shadow": "off",
+    "@typescript-eslint/no-shadow": ["error"],
```
cpoile, line 337: was this intentional?

cpoile (approved):

webapp/.eslintrc.json
streamer45: Yes, it's [known bug](https://github.com/typescript-eslint/tslint-to-eslint-config/issues/856) with linting typescript enums. Ideally we should update since it's been fixed but I am inclined to defer that to another time.

PR #186, opened 2022-09-08
[MM-46918] Fallback to using telemetry ID as rtcd's client ID

streamer45:
#### Summary

This is one bit of confusing syntax I wanted to get rid of. Using the telemetry ID as a fallback seems to be a good choice as it's saved in DB already and unique across the cluster.

This means the `RTCD service URL` setting can be now configured simply using the `http://hostname:port` form.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-46918


cpoile (approved):
Nice!

PR #189, opened 2022-09-12
[MM-46672] Implement non-blocking client errors

streamer45:
#### Summary

There's quite a lot here but mainly:

- **UI changes**: handling client-side errors through banners/tooltips/button icons.
- **Functionality changes**: calls can now be joined without audio input devices and/or permissions to them.

I tried to keep everything as consistent as possible. I am happy to change any string, just trying to make sense of all the possible interactions (tooltip text/subtext/banner/menu label).

#### Screenshots

image

image

image

image

image

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-46672
https://mattermost.atlassian.net/browse/MM-45469


webapp/src/constants.ts
```diff
@@ -0,0 +1,29 @@
...
+        icon: 'microphone',
+        bannerText: 'Allow microphone access to Mattermost.',
+        tooltipText: 'No audio input permissions',
+        tooltipSubtext: 'Allow microphone access to Mattermost',
+    },
+    missingScreenPermissions: {
+        type: CallAlertType.Error,
+        icon: 'monitor',
+        bannerText: 'Allow screen recording access to Mattermost in your system preferences.',
+        tooltipText: 'No screen-sharing permissions',
```
justinegeffen, line 25:
```suggestion
        tooltipText: 'No screen sharing permissions',
```

justinegeffen (approved):
This is great! I really like the consistency, thank you! :) Just added some punctuation suggestions.

webapp/src/components/expanded_view/global_banner.tsx
```diff
@@ -0,0 +1,77 @@
...
+  a, a:hover, a:visited {
+    color: ${({color}) => color};
+  }
+
+  i {
+    font-size: 18px;
+  }
+`;
+
+const Body = styled.span<{}>`
```
cpoile, line 64: minor nit: same here, no need for the empty props

webapp/src/components/expanded_view/controls_button.tsx
```diff
@@ -0,0 +1,131 @@
...
+const ButtonContainer = styled.div<{margin?: string}>`
+   display: flex;
+   flex-direction: column;
+   align-items: center;
+   justify-content: center;
+   margin: ${({margin}) => margin || '0 16px'};
+`;
+
+const Button = styled.button<{bgColor: string, isDisabled?: boolean, isUnavailable?: boolean}>`
+  &&& {
```
cpoile, line 82: curious why we need the extra specificity (&&&) here?

cpoile (approved):
This is great, thanks Claudio.
Super minor nits, nothing big (other than the feeling that we need to refactor these components and make them more manageable 😂 ).

tanmay-des:
This looks great! Thanks @streamer45 
Two that I had in mind

image
Can we make it so the close icon on the right sticks to the end of the toast

image
Can we change icons in the global header? according to the type of error?

streamer45:
> Can we make it so the close icon on the right sticks to the end of the toast

Absolutely, I totally didn't notice this, thanks :+1:

> Can we change icons in the global header? according to the type of error?

Indeed, that would make it more consistent. Will update.

webapp/src/components/expanded_view/controls_button.tsx
streamer45: Some definitions when implementing `WidgetButton` were definitely conflicting with some other in-product (coming from webapp that is) stylings of buttons and likely tried to prevent the same when porting the same to the expanded view although it doesn't seem to be needed there. I removed them, thanks :+1:

streamer45:
Initial reviews have been addressed. Still waiting on some UX work to implement animations on the error banners.

streamer45:
@tanmay-des Could you give this a try and let me know how it's looking now?

tanmay-des (approved):
Just checked it out! Looks amazing @streamer45

PR #190, opened 2022-09-13
[MM-46972] Implement standalone widget

streamer45:
#### Summary

The first of hopefully not to many steps to implement the global widget on desktop. This PR extracts the widget into a standalone react app so that it gets decoupled from the webapp and load only what's strictly necessary.

The widget can be accessed at:

`http://MM_SITRE_URL/static/plugins/com.mattermost.calls/widget/widget.html?call_id=XXX`

`call_id` would equal the channel id at the moment but since it could be something more in the future I decided to name it in advance to what's likely going to become.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-46972


webapp/src/websocket_handlers.ts
```diff
@@ -0,0 +1,201 @@
...
+    followThread,
+} from './utils';
+
+import {
+    connectedChannelID,
+} from './selectors';
+
+import {logErr, logDebug} from './log';
+
+export function handleCallEnd(store: Store, ev: any) {
```
cpoile, line 36: I like this refactor, nice.

webapp/src/components/call_widget/component.tsx
```diff
@@ -1208,7 +1220,7 @@ export default class CallWidget extends React.PureComponent<Props, State> {
         const MuteIcon = window.callsClient.isMuted() ? MutedIcon : UnmutedIcon;
         const muteTooltipText = window.callsClient.isMuted() ? 'Click to unmute' : 'Click to mute';
 
-        const hasTeamSidebar = Boolean(document.querySelector('.team-sidebar'));
+        const hasTeamSidebar = Boolean(document.querySelector('.team-sidebar')) || Boolean(this.props.global);
```
cpoile, line 1223: Is this correct? Has team sidebar if global?  Or, maybe it should be renamed to: `widerWidget` or something, to be clearer?

cpoile (approved):
I can't see anything blocking, you're a React and webpack pro now!

webapp/src/components/admin_console_settings/rtcd_service_url/index.tsx
```diff
@@ -34,7 +34,7 @@ const RTCDServiceUrl = (props: CustomComponentProps) => {
     const rightCol = 'col-sm-8';
 
     // Webapp doesn't pass the placeholder setting.
-    const placeholder = manifest.settings_schema?.settings.find((e) => e.key === 'RTCDServiceURL')?.placeholder || '';
+    const placeholder = manifest.settings_schema?.settings.find((e: Record<string, any>) => e.key === 'RTCDServiceURL')?.placeholder || '';
```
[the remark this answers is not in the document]
streamer45: Oh that's why I was constantly getting the error. Thanks :)

webapp/src/components/call_widget/component.tsx
streamer45: Yes, it's a naming/semantic mismatch. I like your suggestion, thanks!

streamer45:
> I can't see anything blocking, you're a React and webpack pro now!

:bowing_man: 

By the way, a couple of things I forgot to mention/do:

- Changed the base branch to `MM-46971-global-widget` which I'll be using for subsequent PRs belonging to this work.
- While I think this is good to merge to the feature branch, I'll soon be changing the directory structure a bit since it's becoming clear now that this standalone logic will be useful for more things than just the widget (e.g. separate window view for external calls or even recordings).

PR #191, opened 2022-09-15
[MM-47001] Fix accessibility conflict

streamer45:
#### Summary

Keyboard shortcut to mute/unmute is causing a conflict with the accessibility controller since the [recommended behaviour](https://www.w3.org/WAI/ARIA/apg/patterns/button/) is that `Space` toggles a button in focus.

Given we are providing specific shortcuts I am thinking accessibility is retained hence preventing this behaviour by unfocusing the button element on click and let the in-product shortcut do its thing. But I am happy to revisit, just let me know your thoughts.

More context can be found in the original discussion at https://community.mattermost.com/core/pl/r9rktsydz3ndmmdqyjkicb8i6o

#### Ticket Link

https://mattermost.atlassian.net/browse/

### DIFF
--- a/webapp/src/components/call_widget/component.tsx
+++ b/webapp/src/components/call_widget/component.tsx
@@ -452,6 +452,11 @@ export default class CallWidget extends React.PureComponent<Props, State> {
             return;
         }
 
+        // This is needed to prevent a conflict with the accessibility controller on buttons.
+        if (document.activeElement instanceof HTMLElement) {
+            document.activeElement.blur();
+        }
+
         const isMuted = window.callsClient.isMuted();
         if (isMuted) {
             window.callsClient.unmute();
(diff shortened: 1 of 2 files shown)


webapp/src/components/call_widget/component.tsx
```diff
@@ -452,6 +452,11 @@ export default class CallWidget extends React.PureComponent<Props, State> {
             return;
         }
 
+        // This is needed to prevent a conflict with the accessibility controller on buttons.
+        if (document.activeElement instanceof HTMLElement) {
+            document.activeElement.blur();
+        }
```
cpoile, line 458: Apologies, I don't fully understand why we're blurring the activeElement.. If you've already clicked it, or used the shortcut to 'click' it, then what does blurring it do?

streamer45: @cpoile When you click to unmute (or any button) the browser implicitly focuses on the button after which <kbd>Space</kbd> can be used to toggle it on/off. 

To reproduce:

- Start call.
- Unmute by clicking button.
- Try to mute back with shortcut. Pressing a single <kbd>Space</kbd> should work instead.

This is because it triggers the accessibility controller before we get a chance to do anything on our end. Unfocusing on click prevents this behaviour.

cpoile: Ah, I get it now. Yep, I can see it. Thanks.

cpoile (approved):

streamer45:
@tanmay-des Let me know if anything is unclear here.

streamer45:
Discussed internally with @tanmay-des . Merging this as a fix while we think of a proper way to customize shortcuts.

PR #196, opened 2022-09-21
initial redux setup

Willyfrog:
#### Summary
initial setup for redux, don't expect it to work right away as it is not tested, but should provide the places to modify if needed without breaking each others work.


EricssonLiu (approved):
lgtm 👍

PR #198, opened 2022-09-22
Reaction websocket broadcast

agarciamontoro:
#### Summary
This PR adds the minimum logic to listen for reaction events sent by the client and broadcast them to the users in the call.

I also added a second commit (sorry for the force-push, I had to fix some typing there :nerd_face:) with an endless loop of emoji reactions sent by every client that joins the call. This will be removed as soon as the emoji picker is in place.

#### Ticket Link
--

### DIFF
--- a/webapp/src/client.ts
+++ b/webapp/src/client.ts
@@ -271,7 +271,8 @@ export default class CallsClient extends EventEmitter {
             this.intervalID = window.setInterval(() => {
                 this.ws?.send('reaction', {
                     data: JSON.stringify({
-                        emoji: emojis[count],
+                        name: emojis[count],
+                        skin: count,
                     }),
                 });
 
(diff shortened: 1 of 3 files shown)


Willyfrog (approved):
Lgtm

webapp/src/client.ts
```diff
@@ -271,7 +271,8 @@ export default class CallsClient extends EventEmitter {
             this.intervalID = window.setInterval(() => {
                 this.ws?.send('reaction', {
                     data: JSON.stringify({
-                        emoji: emojis[count],
```
AGMETEOR: We can remove this now since https://github.com/mattermost/mattermost-plugin-calls/pull/199 is open

AGMETEOR (approved):
LGTM pending removal of webapp change

webapp/src/client.ts
agarciamontoro: Done in https://github.com/mattermost/mattermost-plugin-calls/pull/198/commits/e13e727f1f9063d553bb6691c65938815104638a

cyrilzhang-mm (approved):

PR #199, opened 2022-09-22
Add emoji picker to calls expanded view

AGMETEOR:
#### Summary
Add an emoji picker to calls expanded view

#### Ticket Link
None

#### Screenshots
Screenshot 2022-09-22 at 14 03 09


Willyfrog (changes requested):

Willyfrog:
Looking good other than the shortcut choice :)

webapp/src/shortcuts.ts
```diff
@@ -13,6 +14,7 @@ const globalMappings = {
 const widgetMappings = {
     [isMac() ? 'meta+shift+space' : 'ctrl+shift+space']: MUTE_UNMUTE,
     [isMac() ? 'meta+shift+y' : 'ctrl+shift+y']: RAISE_LOWER_HAND,
+    [isMac() ? 'meta+shift+r' : 'ctrl+shift+r']: MAKE_REACTION,
```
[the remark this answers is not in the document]
AGMETEOR: Refreshing should be `meta+r`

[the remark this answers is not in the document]
AGMETEOR: @Willyfrog which shortcut do you recommend?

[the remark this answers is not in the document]
AGMETEOR: Changed it to `ctrl+shift+x`

cyrilzhang-mm (approved):

webapp/src/types/types.ts
```diff
@@ -111,3 +111,11 @@ export type CallsUserPreferences = {
 export const CallsUserPreferencesDefault = {
     joinSoundParticipantsThreshold: 8,
 };
+
+export type EmojiData = {
+    id: string;
+    skin?: number;
```
cyrilzhang-mm, line 117: Can we make the skin a string instead of number? Like how the skins are represented in https://github.com/mattermost/mattermost-webapp/blob/master/utils/emoji.json
e.g. a skin looks like "1F3FB"
If not that's fine I'll make a map from number to string when rendering the emoji.

AGMETEOR: That’s a good suggestion. However, the third party library we’re using presents it as a number

Willyfrog: when an emoji has a skin code, it is usually present on the unicode (that 1f3fb)
how does the library return the skin when there are multiple? like family, or holding hands ones? those usually have a skin combination as each person is represented.

Willyfrog (approved):
other than the multiskin code on some cases, looks good to me, if the type matches for emojis like 👩🏾‍🤝‍👨🏼 it is good to go for me.

webapp/src/types/types.ts
AGMETEOR: Not really sure because I haven’t read much in the documentation.

AGMETEOR: @cyrilzhang-mm maybe we need the server to send more data about the emoji. That’s if you’ll be using this method from the library https://github.com/missive/emoji-mart/blob/2054b155cac965bb432ca6b6cd70b51304f8ed7f/packages/emoji-mart/src/utils.js#L18

cyrilzhang-mm: For rendering the emoji, all that is needed is the emoji name, the skin, and the unified string. Technically only the unified string is needed, note that the unified string always contains the skin as well. So it's not a problem if the skin is a number. This includes for emoji with multiple people in them, the unified string should also contain the skin for all people in the emoji. If you can check that emoji-mart does this that would be good, but I'm almost certain it will.

If you want to get an idea for how I will be rendering the emoji check my commit here https://github.com/mattermost/mattermost-plugin-calls/commit/e1cbfcb337c13727e78c08d096a38a3e0025eb0b I'm using the `EmojiIndicesByAlias` map but there is also `EmojiIndicesByUnicode` which uses the unified string instead of the emoji name. 

Also @Willyfrog let me know your thoughts as you will be rendering the emoji in the same way I think?

EricssonLiu (approved):
Looks good to me :)

webapp/src/types/types.ts
```diff
@@ -111,3 +111,9 @@ export type CallsUserPreferences = {
 export const CallsUserPreferencesDefault = {
     joinSoundParticipantsThreshold: 8,
 };
+
+export type EmojiData = {
+    id: string;
```
agarciamontoro, line 116: Should this be `name`? I can change it in the server instead, 0/5

webapp/src/components/expanded_view/component.tsx
```diff
@@ -561,6 +604,34 @@ export default class ExpandedView extends React.PureComponent<Props, State> {
...
+                            <div style={{position: 'relative'}}>
+                                {this.renderEmojiPicker()}
+                                <OverlayTrigger
+                                    key='tooltip-emoji-picker'
+                                    placement='top'
+                                    overlay={
+                                        <Tooltip
+                                            id='tooltip-emoji-picker'
+                                        >
+                                            <span>{'Make a reaction'}</span>
```
agarciamontoro, line 616: In Channels this is `Add Reaction`, but 0/5
image

agarciamontoro (approved):
Awesome! I left some questions and comments below, but all non-blocking.

webapp/src/types/types.ts
cyrilzhang-mm: Oops yes, let me fix this.

webapp/src/types/types.ts
Willyfrog: I'll happily adapt to whatever we use. My only concern on they way that it is being rendered is that we need to import all the emoji data from webview and that needs updating once it updates on webapp. But other than that the system works, so I'm ok with that.

PR #200, opened 2022-09-22
Fix send event 

cyrilzhang-mm:
#### Summary

We need to stringify the data before sending it to the server.


EricssonLiu (approved):
Good catch 🎣

PR #202, opened 2022-09-23
Close on click outside

AGMETEOR:
#### Summary
Close on click outside

### DIFF
--- a/webapp/src/components/expanded_view/component.tsx
+++ b/webapp/src/components/expanded_view/component.tsx
@@ -229,6 +229,7 @@ export default class ExpandedView extends React.PureComponent<Props, State> {
                     emojiVersion={EMOJI_VERSION}
                     skinTonePosition='search'
                     onEmojiSelect={this.handleUserPicksEmoji}
+                    onClickOutside={this.toggleEmojiPicker}
                 />
             </div>
         ) : null;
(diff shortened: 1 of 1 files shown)


webapp/src/components/expanded_view/component.tsx
```diff
@@ -229,6 +229,7 @@ export default class ExpandedView extends React.PureComponent<Props, State> {
                     emojiVersion={EMOJI_VERSION}
                     skinTonePosition='search'
                     onEmojiSelect={this.handleUserPicksEmoji}
+                    onClickOutside={this.toggleEmojiPicker}
```
agarciamontoro, line 232: Does this handler fire only when the user clicks outside *and* the emoji picker is open? Or also when it's closed?

agarciamontoro: Nevermind, I tested it and it works great!

AGMETEOR: Only when it's open since the picker is already wrapped in an if check for `showEmojiPicker`

agarciamontoro (approved):
Neat! :tada:

cyrilzhang-mm (approved):

AGMETEOR:
We got 2 dev reviews. Going to merge

PR #204, opened 2022-09-23
Add a timeout for reactions

agarciamontoro:
#### Summary
Add a reaction timeout so that the emojis don't stay forever on top of the avatars. This applies to the notification stream as well, but I have not been able to test it properly.

#### Ticket Link
--

### DIFF
--- a/webapp/src/index.tsx
+++ b/webapp/src/index.tsx
@@ -95,6 +95,7 @@ import {
     SHOW_SWITCH_CALL_MODAL,
     SHOW_END_CALL_MODAL,
     VOICE_CHANNEL_USER_REACTION,
+    VOICE_CHANNEL_USER_REACTION_TIMEOUT,
 } from './action_types';
 
 import {PluginRegistry, Store} from './types/mattermost-webapp';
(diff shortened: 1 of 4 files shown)


AGMETEOR (approved):
LGTM

cyrilzhang-mm (approved):

agarciamontoro:
Sorry for the re-request for review, but I changed the way reactions are identified: we don't actually need an external identifier as `seq`, we can do with just the `timestamp`.

cyrilzhang-mm (approved):

webapp/src/index.tsx
```diff
@@ -293,21 +294,32 @@ export default class Plugin {
...
-                            name: ev.data.emoji_name,
-                            skin: ev.data.emoji_skin,
-                            unified: ev.data.emoji_unified,
-                        },
-                        timestamp: ev.data.timestamp,
-                    },
+                    reaction,
                 },
             });
+            setTimeout(() => {
```
Willyfrog, line 313: I'm wondering if we should store this in a queue the same size as the reactions queue, so if there are people reacting like crazy we don't eat up a lot of memory. but that might be me doing premature performance 0/5
Anyway, we can add it later, as my changes are not in the branch yet and I'll have to update the code to match what it is approved already :)

EricssonLiu (approved):
lgtm 👁

webapp/src/reducers.ts
```diff
@@ -452,6 +462,26 @@ const voiceUsersStatuses = (state: usersStatusesState = {}, action: usersStatuse
...
             },
         };
+    case VOICE_CHANNEL_USER_REACTION_TIMEOUT:
+        if (!state[action.data.channelID]?.[action.data.userID]?.reaction) {
+            return state;
+        }
+        if (!action.data.reaction) {
+            return state;
+        }
+        if (state[action.data.channelID][action.data.userID].reaction?.timestamp !== action.data.reaction.timestamp) {
```
Willyfrog, line 472: I'm wondering if a `<=` wouldn't be safe in case something goes wrong. Again, not really needed, but if for some reason a reaction fails to time out this would cover it.

Willyfrog (approved):
Added a couple of suggestions, but neither of them are blocking or required. looks good 🚀

webapp/src/index.tsx
agarciamontoro: Wait. Why don't we simply cancel the previous timeout when a new reaction arrives? We can store the timeout ID in memory along with the reaction and, if there's one already, we just cancel it. We can still check for the timestamp when handling the timeout event to make sure no race condition happens anywhere, but I think that's doable.
Let me try.

webapp/src/reducers.ts
agarciamontoro: Ah, yeah, that makes sense.

agarciamontoro: Done in https://github.com/mattermost/mattermost-plugin-calls/pull/204/commits/2a967201b342204c3641a92539466e1bb195a014

webapp/src/index.tsx
agarciamontoro: I'll merge this PR and open a new one with this change.

PR #205, opened 2022-09-23
Minor bug fixes

cyrilzhang-mm:
#### Summary

Fix the grinning emoji and fix emoji version for emoji picker. The data prop is actually not needed and it overrides the emojiVersion prop.

Grinning emoji:
Screen Shot 2022-09-23 at 7 45 59 AM

Emoji picker (no more v14 emojis)
Screen Shot 2022-09-23 at 7 46 18 AM

### DIFF
--- a/webapp/src/components/expanded_view/component.tsx
+++ b/webapp/src/components/expanded_view/component.tsx
@@ -5,7 +5,6 @@
 import React, {CSSProperties} from 'react';
 import {compareSemVer} from 'semver-parser';
 import {OverlayTrigger, Tooltip} from 'react-bootstrap';
-import data from '@emoji-mart/data';
 import Picker from '@emoji-mart/react';
 
 import {getEmojiImageUrl} from 'mattermost-redux/utils/emoji_utils';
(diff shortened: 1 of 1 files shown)


webapp/src/components/expanded_view/component.tsx
```diff
@@ -47,7 +46,7 @@ import {
 
 import './component.scss';
 
-const EMOJI_VERSION = 13;
+const EMOJI_VERSION = '13';
```
agarciamontoro, line 49: :facepalm:

agarciamontoro (approved):
Great! Thank you, Cyril :tada:

webapp/src/components/expanded_view/component.tsx
agarciamontoro: Ah, but the actual problem was the `data` prop, right?

cyrilzhang-mm: Ah yes. I just made this explicitly a string since we can pass versions such as 13.1 to it as well.

Willyfrog (approved):

PR #206, opened 2022-09-23
Stream reactions

Willyfrog:
#### Summary

Stream on participant list

#### Pending

- [x] raising hands. Done in https://github.com/mattermost/mattermost-plugin-calls/pull/208
- [ ] display on top of screen sharing
- [ ] bonus: option to hide the stream if someone finds it distracting.
- [x] reuse emoji component


webapp/src/components/reaction_stream/reaction_stream.tsx
```diff
@@ -0,0 +1,88 @@
...
+    margin: 2px
+`;
+
+const getEmojiURL = (emoji: EmojiData) => {
+    const index = EmojiIndicesByUnicode.get(emoji.unified.toLowerCase());
+    if (typeof index === 'undefined') {
+        return '';
+    }
+    return getEmojiImageUrl(Emojis[index]);
+};
```
agarciamontoro, line 50: Isn't this copied somewhere else? I'm not opposed to it, just curious because I'd swear I've seen this these days.

webapp/src/components/reaction_stream/reaction_stream.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+type Props = {
+    reactions: ReactionWithUser[],
+    currentUserID: string,
+    profiles: {[key: string]: UserProfile;},
+};
+
+const ReactionStreamList = styled.div`
+    position='absolute';
+    left=0;
```
cyrilzhang-mm, line 22: `:`? 
Also a suggestion: We can probably add a fixed height to this of about 75% of the viewport height as a way to artificially make the reactions "disappear" when there are many of them.

webapp/src/components/reaction_stream/reaction_stream.tsx
```diff
@@ -0,0 +1,88 @@
...
+    }
+    return getEmojiImageUrl(Emojis[index]);
+};
+
+// add a list of reactions, on top of that add the hands up as the top element
+export const ReactionStream = (props: Props) => {
+    const reversed = [...props.reactions];
+
+    // temporary push items to mimic a raised hands
+    //reversed.push({emoji: {name: 'raised-hand', unified: 's1', skin: ''}, timestamp: 1, user_id: 'ut3kxffbi7fxzfbz9igmc3ejte'});
```
agarciamontoro, line 57: Remove?
```suggestion
```

webapp/src/components/reaction_stream/reaction_stream.tsx
```diff
@@ -0,0 +1,88 @@
...
+                    backgroundImage: `url(${getEmojiURL(reaction.emoji)})`,
+                    width: '18px',
+                    minWidth: '18px',
+                    height: '18px',
+                    minHeight: '18px',
+                }}
+            >
+                {reaction.emoji.name}
+            </span>);
+        const user = reaction.user_id === props.currentUserID ? 'you' : getUserDisplayName(props.profiles[reaction.user_id]) || 'unknown';
```
agarciamontoro, line 78: Nit: Maybe 'someone' instead of unknown?
```suggestion
        const user = reaction.user_id === props.currentUserID ? 'you' : getUserDisplayName(props.profiles[reaction.user_id]) || 'someone';
```

webapp/src/components/reaction_stream/reaction_stream.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+const ReactionStreamList = styled.div`
+    position='absolute';
+    left=0;
+    bottom=100px;
+    display: flex;
+    flex-direction: column-reverse;
+    margin-left: 10px;
+    -webkit-mask: linear-gradient(#0000, #000);
+    mask: linear-gradient(#0000, #0003, #000f);
```
EricssonLiu, line 28: Oh this is a super smart way to do it 🤯

webapp/src/reducers.ts
```diff
@@ -226,9 +226,9 @@ interface userReactionsState {
 }
 
 const queueReactions = (state: ReactionWithUser[], reaction: ReactionWithUser) => {
-    const result = [...state];
+    const result = state?.length ? [...state] : [];
     result.push(reaction);
-    if (result.length > 8) { // TODO: random size, this should probably be configurable
+    if (result.length > 50) { // TODO: random size, this should probably be configurable
```
agarciamontoro, line 231: Is this for debugging purposes? Or do we want 50?

agarciamontoro (approved):
Looking good!!!! :rocket: I left some comments below, but nothing blocking.

cyrilzhang-mm (commented):
I checked out your branch and it looks good, works really well. Just a few comments, mainly on styling.

webapp/src/components/reaction_stream/reaction_stream.tsx
cyrilzhang-mm: Yes it is. We can move it to another file I guess?

webapp/src/components/reaction_stream/reaction_stream.tsx
```diff
@@ -0,0 +1,88 @@
...
+                    minWidth: '18px',
+                    height: '18px',
+                    minHeight: '18px',
+                }}
+            >
+                {reaction.emoji.name}
+            </span>);
+        const user = reaction.user_id === props.currentUserID ? 'you' : getUserDisplayName(props.profiles[reaction.user_id]) || 'unknown';
+        return (
+            <ReactionChip key={reaction.timestamp + reaction.user_id}><span>{emoji}</span>&nbsp;<span>{user}</span></ReactionChip>
```
EricssonLiu, line 80: Probably better to multiline this?

EricssonLiu (approved):
looks good :)

webapp/src/components/reaction_stream/reaction_stream.tsx
Willyfrog: yep, I removed my code and used the same. the plan is to move the emoji code to its own component and reuse from there. But I wanted to make sure everythign looked good before pushing around everything. I'll add it to the list of todos in the PR description

webapp/src/reducers.ts
Willyfrog: I wanted a big number enought to cover most screens, originally we agreed on 8 but later on decided on 70% of the screen, but I don't want to hold the entire history of reactions, as a two hour call might be able to crash lower end computers, so either 50 or whatever people feel it is a good number, but not limitless :)

webapp/src/components/reaction_stream/reaction_stream.tsx
```diff
@@ -0,0 +1,88 @@
...
+    mask: linear-gradient(#0000, #0003, #000f);
+`;
+
+const ReactionChip = styled.div`
+    display: flex;
+    flex-direction: row;
+    align-items: flex-end;
+    padding: 4px 6px;
+    gap: 4px;
+    width: 94px;
```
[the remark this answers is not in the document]
Willyfrog: no, that's copied from the figma, but can be removed, I'll test it

webapp/src/components/expanded_view/component.tsx
```diff
@@ -537,13 +538,19 @@ export default class ExpandedView extends React.PureComponent<Props, State> {
         const currentID = this.props.currentUserID;
         const isSharing = sharingID === currentID;
 
+        // building the list here causes a bug tht if a user leaves and recently reacted it will show as blank
```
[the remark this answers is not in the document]
Willyfrog: that's another approach, we could just keep the list of profiles so we don't end up having this problem, but I didn't dig too much into it

webapp/src/components/reaction_stream/reaction_stream.tsx
Willyfrog: yes, I was just testing how it would look if I pushed an element and forgot about it, thanks!

webapp/src/components/reaction_stream/reaction_stream.tsx
Willyfrog: sure, I've changed through 3 names already but I don't find any that I like, so if you prefer that one, I'm ok with that :)

webapp/src/components/reaction_stream/reaction_stream.tsx
Willyfrog: ok tested this and doesn't look good, but that's my faulty css, we can iterate on this later

webapp/src/components/reaction_stream/reaction_stream.tsx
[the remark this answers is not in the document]
Willyfrog: switched to fit-content

webapp/src/components/reaction_stream/reaction_stream.tsx
cyrilzhang-mm: These lines of CSS are also invalid, they should say `:` instead of `=`.

cyrilzhang-mm: Actually let me make a suggestion here. Try replacing these 
```
    position='absolute';
    left=0;
    bottom=100px;
```
with
```
    align-self: flex-end;
    height: 75vh;
 ```
 I think this will have the effect we want.

File: webapp/src/components/reaction_stream/reaction_stream.tsx [hunk 10]
```diff
@@ -0,0 +1,68 @@
...
+const ReactionChip = styled.div`
+    display: flex;
+    flex-direction: row;
+    align-items: flex-end;
+    padding: 2px 10px;
+    gap: 4px;
+    height: 23px;
+    background: rgba(221, 223, 228, 0.08);
+    border-radius: 12px;
+    margin: 2px
```
cyrilzhang-mm, line 40: Missing a semicolon here. Also a margin of `4px 0` matches the Figma better I think.

webapp/src/components/reaction_stream/reaction_stream.tsx
```diff
@@ -0,0 +1,68 @@
...
+    -webkit-mask: linear-gradient(#0000, #000);
+    mask: linear-gradient(#0000, #0003, #000f);
+`;
+
+const ReactionChip = styled.div`
+    display: flex;
+    flex-direction: row;
+    align-items: flex-end;
+    padding: 2px 10px;
+    gap: 4px;
```
cyrilzhang-mm, line 36: This gap is too large, I think `2px` is better.

webapp/src/components/reaction_stream/reaction_stream.tsx
Willyfrog: oh! good catch

webapp/src/components/reaction_stream/reaction_stream.tsx
```diff
@@ -0,0 +1,68 @@
...
+export const ReactionStream = (props: Props) => {
+    const reversed = [...props.reactions];
+
+    // add hands up into reversed
+    reversed.reverse();
+    const elements = reversed.map((reaction) => {
+        // emojis should be a separate component that is reused both here and in the extended view
+        // getEmojiURL should be memoized as people tend to react similarly and this would speed up the process.
+        const emoji = (<Emoji emoji={reaction.emoji}/>);
+        const user = reaction.user_id === props.currentUserID ? 'you' : getUserDisplayName(props.profiles[reaction.user_id]) || 'someone';
```
cyrilzhang-mm, line 54: s/you/You ? Just to match the Figma.

cyrilzhang-mm (approved):
I made a couple of style comments. Mostly nitpicks and two places where the CSS is invalid so they weren't being applied.

webapp/src/components/reaction_stream/reaction_stream.tsx
Willyfrog: yep, that seems to work better, thanks!

EricssonLiu (approved):

PR #207, opened 2022-09-23
Clear reaction timeout as soon as possible

agarciamontoro:
#### Summary
This is a purely performance-focused PR. No behaviour is modified, it only focuses on clearing old timers so that they don't accumulate in the case of a user reacting like crazy.

#### Ticket Link
--

### DIFF
--- a/webapp/src/index.tsx
+++ b/webapp/src/index.tsx
@@ -14,6 +14,7 @@ import {getConfig} from 'mattermost-redux/selectors/entities/general';
 
 import {displayFreeTrial, getCallsConfig} from 'src/actions';
 import {PostTypeCloudTrialRequest} from 'src/components/custom_post_types/post_type_cloud_trial_request';
+import {voiceUserStatus} from 'src/selectors';
 
 import RTCDServiceUrl from 'src/components/admin_console_settings/rtcd_service_url';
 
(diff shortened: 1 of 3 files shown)


cyrilzhang-mm (approved):

webapp/src/index.tsx
```diff
@@ -302,15 +303,13 @@ export default class Plugin {
...
-                    userID: ev.data.userID,
-                    reaction,
-                },
-            });
-            setTimeout(() => {
+
+            // Clear the previous timeout of the users's reaction, if any
+            const userState = voiceUserStatus(store.getState(), ev.broadcast.channel_id, ev.data.userID);
+            clearTimeout(userState?.reaction?.timeoutID);
+
```
Willyfrog, line 310: I'm not sure if I'm getting this, is this removing the previous timout and replacing it with a new one per user?
so, if a user clicks like crazy, it will keep the reactions in the stream for longer as the timeouts are cleared, right? probably wont even clear unless they go outside of the queue/view

Willyfrog (commented):
Left a comment on it, not sure if I got the behaviour right.

webapp/src/index.tsx
cyrilzhang-mm: I think the idea is when the final timeout occurs, it will remove all reactions with timestamp less than the final one, so all of them will be removed. But I think yes, while the user is clicking, the old reactions are kept in the queue. Only after the user stops clicking + 5s will all the reactions be removed.

cyrilzhang-mm: Hmm, maybe there is a problem if the user is reacting every 4 seconds assuming the timeout is for 5 seconds. In that case the user is not clicking like crazy but it will still clear the timeout every time, making the reactions stay in the queue...

agarciamontoro: Thanks for your questions and comments, @Willyfrog and @cyrilzhang-mm. Thanks to that I realized this does not work at all :facepalm: It only removes the notifications of a user when the last timeout fires, of course.

PR #212, opened 2022-09-26
Fix horizontal shift when adding reactions

agarciamontoro:
#### Summary

Absolutely position the reaction list inside the parent div so that it does not shift the placement of any other siblings. The `flex-direction` is no longer needed in the parent, so I removed it.

https://user-images.githubusercontent.com/3924815/192263536-f1017d92-a58f-4a66-9ec3-e09aebd1385a.mp4

#### Ticket Link
--

### DIFF
--- a/webapp/src/components/expanded_view/component.tsx
+++ b/webapp/src/components/expanded_view/component.tsx
@@ -567,7 +567,7 @@ export default class ExpandedView extends React.PureComponent<Props, State> {
                         }
                     </div>
                     { !this.props.screenSharingID &&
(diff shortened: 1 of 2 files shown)


cyrilzhang-mm (approved):
Makes sense to me! Although there can be a problem where if someone's name is too long, it can block some of the UI. However as the reactions will disappear shortly, I think it's fine.

agarciamontoro:
> if someone's name is too long, it can block some of the UI

Yup, you're right. If we see that's a problem, we can always set a maximum width for the chips and ellipsize if the content is larger.

AGMETEOR (approved):

PR #213, opened 2022-09-28
[MM-46993] Global widget desktop communication

streamer45:
#### Summary

PR implements the logic to communicate with the Desktop side. This includes any necessary dynamic resizing of the widget.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-46993

#### Related PRs

https://github.com/mattermost/desktop/pull/2265

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -195,7 +195,7 @@ ifneq ($(HAS_WEBAPP),)
 	mkdir -p dist/$(PLUGIN_ID)/webapp
 	cp -r webapp/dist dist/$(PLUGIN_ID)/webapp/
 	## We remove the assets directory since files are redundant and loaded from webapp.
-	rm -fr widget/dist/files
+	rm -fr widget/dist/files/*.png
 	cp -r widget/dist dist/$(PLUGIN_ID)/webapp/dist/widget
 endif
 	cd dist && tar -cvzf $(BUNDLE_NAME) $(PLUGIN_ID)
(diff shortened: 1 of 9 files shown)


Makefile
```diff
@@ -195,7 +195,7 @@ ifneq ($(HAS_WEBAPP),)
 	mkdir -p dist/$(PLUGIN_ID)/webapp
 	cp -r webapp/dist dist/$(PLUGIN_ID)/webapp/
 	## We remove the assets directory since files are redundant and loaded from webapp.
-	rm -fr widget/dist/files
+	rm -fr widget/dist/files/*.png
```
streamer45, line 198: Using the existing assets from webapp became problematic but going completely standalone means that webpack will include tons of stuff that's not really needed, some of which are 20MB pngs with the whole emoji set. I need to find a proper solution to this but for now just removing the heavy unused stuff to keep the build to a reasonable size.

webapp/src/utils.ts
```diff
@@ -325,11 +325,13 @@ export function playSound(name: string) {
...
     }
 
-    // A hack to make the standalone widget re-use the sound files included
-    // from here.
-    src = src.replace('/static', '');
+    if (src.indexOf('/static') === 0) {
+        src = `${window.basename || ''}${src}`;
+    } else {
+        src = getPluginStaticPath() + src;
+    }
```
streamer45, line 332: I still need to test the Mattermost under a subpath case because I have a feeling it will break some stuff.

cpoile (approved):
Looks good!

PR #214, opened 2022-09-30
Qol contribute changes

Willyfrog:
#### Summary
- increased timer for reaction disappearance to 10 secs
- emoji picker focus on search bar on opening
- emojis per line 9 -> 12
- put feature behind experimental flag

it is now required to use the command `/calls experimental on`within your mattermost client to access the feature

### DIFF
--- a/webapp/src/components/expanded_view/component.tsx
+++ b/webapp/src/components/expanded_view/component.tsx
@@ -419,7 +419,7 @@ export default class ExpandedView extends React.PureComponent<Props, State> {
                             </div>
                         </>
                         }
-                        {!hasExperimentalFlag() && !isHandRaised && hasReaction && status.reaction &&
+                        {hasExperimentalFlag() && !isHandRaised && hasReaction && status.reaction &&
                         <>
                             <div style={style.reactionBackground as CSSProperties}/>
                             <div style={style.reactionContainer as CSSProperties}>
--- a/webapp/src/components/expanded_view/component.tsx
+++ b/webapp/src/components/expanded_view/component.tsx
@@ -419,7 +419,7 @@ export default class ExpandedView extends React.PureComponent<Props, State> {
                             </div>
                         </>
                         }
-                        {!isHandRaised && hasReaction && status.reaction &&
+                        {!hasExperimentalFlag() && !isHandRaised && hasReaction && status.reaction &&
                         <>
                             <div style={style.reactionBackground as CSSProperties}/>
                             <div style={style.reactionContainer as CSSProperties}>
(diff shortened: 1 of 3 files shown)


webapp/src/components/expanded_view/component.tsx
```diff
@@ -417,7 +419,7 @@ export default class ExpandedView extends React.PureComponent<Props, State> {
                             </div>
                         </>
                         }
-                        {!isHandRaised && hasReaction && status.reaction &&
+                        {!hasExperimentalFlag() && !isHandRaised && hasReaction && status.reaction &&
```
agarciamontoro, line 422: Shouldn't this be `hasExperimentalFlag`?
```suggestion
                        {hasExperimentalFlag() && !isHandRaised && hasReaction && status.reaction &&
```

webapp/src/components/expanded_view/component.tsx
```diff
@@ -242,6 +242,8 @@ export default class ExpandedView extends React.PureComponent<Props, State> {
                     skinTonePosition='search'
                     onEmojiSelect={this.handleUserPicksEmoji}
                     onClickOutside={this.toggleEmojiPicker}
+                    autoFocus={true}
+                    perLine={12}
```
agarciamontoro, line 246: Does this make the picker shorter?

agarciamontoro (changes requested):
Thank you!!!

webapp/src/components/expanded_view/component.tsx
Willyfrog: yes, I meant to change it after seeing I got it reversed, but since I was testing multiple things I forgot about this one

webapp/src/components/expanded_view/component.tsx
Willyfrog: it is wider

cyrilzhang-mm: I think we also wanted to make the picker shorter, so that it only shows about 4 lines at a time. It should be doable via CSS, like in this example: https://github.com/missive/emoji-mart/blob/main/packages/emoji-mart-website/example-custom-styles.html

Willyfrog: we could change the size of the emoji within the emoji picker to make it shorter if that's what we want.

@itao @tanmay-des any thoughts?

cyrilzhang-mm: We can also directly set the height to be shorter as well. But yeah, maybe we can wait until the design review to decide what is the best approach.

cyrilzhang-mm (approved):

AGMETEOR (approved):

EricssonLiu (approved):

PR #215, opened 2022-09-30
 Make consistent order of hands up 

cyrilzhang-mm:
#### Summary

Create the hands up array in the same order as the participants list RHS, so hands up are shown in the order that they were raised. Also changed the background colour to white. 

Screen Shot 2022-09-30 at 10 57 24 AM

### DIFF
--- a/webapp/src/components/expanded_view/component.tsx
+++ b/webapp/src/components/expanded_view/component.tsx
@@ -527,17 +527,16 @@ export default class ExpandedView extends React.PureComponent<Props, State> {
         const currentID = this.props.currentUserID;
         const isSharing = sharingID === currentID;
 
+        const handsup: string[] = [];
+
         // building the list here causes a bug tht if a user leaves and recently reacted it will show as blank
         const profileMap: {[key: string]: UserProfile;} = {};
         this.props.profiles.forEach((profile) => {
             profileMap[profile.id] = profile;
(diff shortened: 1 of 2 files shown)


Willyfrog (changes requested):
This is having the same problem as before as it is using the object keys in whatever ordering is handled, so it might work sometimes but can't be trusted. We need to sort by the value of hands up as I believe it is a timestamp

cyrilzhang-mm:
@Willyfrog I think it's okay because it's using the `profiles` prop, which is an array so it is ordered, and the participants list is ordered in the same way using the `profiles` prop
https://github.com/mattermost/mattermost-plugin-calls/blob/e617565854023842049ee6a27f1ba8e84658bfbc/webapp/src/components/expanded_view/component.tsx#L488-L497
the `profiles` prop is sorted before being passed into the component
https://github.com/mattermost/mattermost-plugin-calls/blob/e617565854023842049ee6a27f1ba8e84658bfbc/webapp/src/components/expanded_view/index.ts#L23-L28

Willyfrog:
> the `profiles` prop is sorted before being passed into the component
> 

Oh, that's right! Never mind my comment then

Willyfrog (approved):

AGMETEOR (approved):
LGTM

PR #217, opened 2022-10-05
[MM-47364] Disable crash on hovering disabled button

streamer45:
#### Summary

There doesn't seem to be a clean way to prevent the `OverlayTrigger` from rendering so I am using css to avoid doing some weird code duplication or more complex stuff (wrapping the element).

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-47364

### DIFF
--- a/webapp/src/components/expanded_view/component.tsx
+++ b/webapp/src/components/expanded_view/component.tsx
@@ -684,6 +684,7 @@ export default class ExpandedView extends React.PureComponent<Props, State> {
                                     />
                                 }
                                 unavailable={noScreenPermissions}
+                                disabled={sharingID !== '' && !isSharing}
                             />
 
                             <ControlsButton
--- a/webapp/src/components/expanded_view/controls_button.tsx
+++ b/webapp/src/components/expanded_view/controls_button.tsx
@@ -26,8 +26,10 @@ export default function ControlsButton(props: Props) {
             key={props.id}
             placement='top'
             overlay={
-                props.disabled ||
-                <Tooltip id={`tooltip-${props.id}`}>
+                <StyledTooltip
+                    id={`tooltip-${props.id}`}
+                    isDisabled={props.disabled}
+                >
                     <div>{props.tooltipText}</div>
                     {props.tooltipSubtext &&
                     <TooltipSubtext>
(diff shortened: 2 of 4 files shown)


webapp/src/components/expanded_view/component.tsx
```diff
@@ -684,6 +684,7 @@ export default class ExpandedView extends React.PureComponent<Props, State> {
                                     />
                                 }
                                 unavailable={noScreenPermissions}
+                                disabled={sharingID !== '' && !isSharing}
```
streamer45, line 687: We were actually forgetting to disable it here.

webapp/src/components/expanded_view/controls_button.tsx
```diff
@@ -125,3 +127,9 @@ const UnavailableIcon = styled.div`
 const TooltipSubtext = styled.div`
   opacity: 0.56;
 `;
+
+const StyledTooltip = styled(Tooltip)<{isDisabled?: boolean}>`
+  ${({isDisabled}) => isDisabled && css`
+      display: none;
+  `}
+`;
```
cpoile, line 135: We can put this is the `shared.tsx`

webapp/src/components/expanded_view/component.tsx
cpoile: I tested, and it looks like it hides it only when you haven't given permissions (not when you press cancel), so that could be a way to distinguish between the two error states? (not 100% sure, only a quick test)

webapp/src/components/expanded_view/controls_button.tsx
streamer45: Oh nice, I didn't realized we had a common file. Thanks :+1:

webapp/src/components/expanded_view/component.tsx
streamer45: Mmm, not sure I understand. What this does is it disables the button when someone else is screen sharing.

cpoile: Hahah... ahh -- I misunderstood. It thought it hides when you don't have permissions to share, because I used Opera to test it and I was logged into a server with one team (smaller LHS), which hides the screensharing button. 🤕  I'm going to fade back into the bushes now.

streamer45: Opera? Good Lord :see_no_evil:

PR #218, opened 2022-10-06
[MM-46532] Implement AllowScreenSharing config setting

streamer45:
#### Summary

PR adds a new config setting to control whether screen sharing is available to calls participants. This comes from a customer request that for security reasons wants to prevent any visual content to be accidentally shared in calls.

The current implementation simply disables the buttons in case screen sharing is not allowed. A couple of open UX questions would be:

1. Does it make sense to show a tooltip with a message explaining to the user why the functionality is not available?
2. Or would it be perhaps better to simply avoid rendering the screen share button altogether?

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-46532


server/websocket.go
```diff
@@ -36,6 +36,10 @@ const (
 )
 
 func (p *Plugin) handleClientMessageTypeScreen(us *session, msg clientMessage, handlerID string) error {
+	if cfg := p.getConfiguration(); cfg == nil || cfg.AllowScreenSharing == nil || !*cfg.AllowScreenSharing {
```
cpoile, line 39: Does this basically mean the default is to not allow screensharing? (although probably the first two checks are not expected to ever happen)

webapp/src/selectors.ts
```diff
@@ -124,7 +124,7 @@ export const maxParticipants: (state: GlobalState) => number = createSelector(
 );
 
 export const needsTURNCredentials: (state: GlobalState) => boolean = createSelector(
-    'maxParticipants',
+    'needsTURNCredentials',
```
cpoile, line 127: hah, thanks

webapp/src/selectors.ts
```diff
@@ -136,6 +136,12 @@ export const isLimitRestricted: (state: GlobalState) => boolean = createSelector
     (numCurrentUsers, max) => max > 0 && numCurrentUsers >= max,
 );
 
+export const allowScreenSharing: (state: GlobalState) => boolean = createSelector(
+    'maxParticipants',
```
cpoile, line 140:
```suggestion
    'allowScreenSharing',
```

cpoile (approved):

webapp/src/selectors.ts
streamer45: Oh well :man_facepalming:

server/websocket.go
cpoile: Sounds good to me.

PR #225, opened 2022-10-20
[MM-47659] Properly fix element stacking

streamer45:
#### Summary

There's a reason I consider myself a backend engineer and today I was reminded of that.

PR fixes an issue in Safari where the notification bar (along with any menu showed on top of the widget) was hidden behind the LHS due to a workaround I used in https://github.com/mattermost/mattermost-plugin-calls/pull/189 to display the notifications coming from below the widget during animation.

Turns out this problem was solvable with some smart use of `z-index` alone plus some minor element rearrangement but it required deeper understanding of how the [stacking context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context) works.

Fortunately I was able to acquire some knowledge mostly thanks to the following resources which I highly recommend: 

https://www.joshwcomeau.com/css/stacking-contexts/
https://philipwalton.com/articles/what-no-one-told-you-about-z-index/

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-47659


streamer45:
I am thinking this issue may be impacting enough to consider a dot release to be published on marketplace and cloud.

webapp/src/components/call_widget/component.tsx
```diff
@@ -1382,15 +1380,16 @@ export default class CallWidget extends React.PureComponent<Props, State> {
                 }}
                 ref={this.node}
             >
-                <div style={this.style.status as CSSProperties}>
-                    <div style={{position: 'absolute', bottom: 'calc(100% + 4px)', width: '100%', transform: 'translateZ(-10px)'}}>
```
cpoile, line 1386: I think all you need to do is remove the transform, that should prevent the div from being unclickable -- not stacking context, per se... Kind of curious, why were you using it?

streamer45: The transform was added as a workaround to make the notification appear from behind the main widget body. I was mostly following https://fjolt.com/article/css-make-child-appear-behind-parent

What am I missing?

cpoile (approved):
Thanks! The new way looks good too, so whichever you prefer.

The articles were good, but I think the JoshComeau one was definitely better, imo.

webapp/src/components/call_widget/component.tsx
cpoile: Oh, I didn't realize that was for notifications. I'm not sure what you're missing ( :) ) but I was missing that info. But I guess the transform causes a stacking context, I'm not sure why that alone would cause this (and in safari alone).

streamer45: The stacking context was already there since we set `z-index` on the root element (`#calls-widget`) to make it show in front of anything else in Mattermost. And the notifications were always showing above the widget until I implemented non-blocking errors and refined the animation which looks much nicer if they come from below. At that point I couldn't find a proper way to make the child element display above its parent so resorted to the transform hack which doesn't work in Safari for some reason (maybe https://stackoverflow.com/questions/25055505/css-transition-transform-z-index-conflict-in-safari-works-on-chrome-ff). Either way, now it's much better as it only relies on z-index.

cpoile: That is pretty cool -- good work here, thanks!

PR #226, opened 2022-10-20
Fix e2e snapshots

streamer45:
#### Summary

Looks like something in boards may have slightly changed the overall size of the menu. Since the change is practically unnoticeable (2px) I am leaning towards just regenerating these.


cpoile (approved):
Sounds good to me. 👍

streamer45:
mmm, they should not fail though, that was the whole point of this PR 😬 , I'll have another look

streamer45:
Okay so e2e are rendering yet another image height than the one I can produce locally. So we either have CI update the snapshots or we need to find a different solution. @Phrynobatrachus thoughts? I'd be also okay with increasing slightly the tolerance here since we are talking about a couple of pixels.

Phrynobatrachus:
I reverted the Boards update in the CI pipeline that started causing failures, my bad for not considering these particular snapshots when I originally did that. 

https://git.internal.mattermost.com/qa/calls-e2e-testing/-/commit/0d4b80c544155d89e401027c98f0f133a927e926

Phrynobatrachus:
Ah it looks like these snapshots are still an issue even without the latest Boards, pushing updates for them in a moment.

Phrynobatrachus:
Alright I definitely am not understanding something here, it seems like the widget was different sizes in different runs. Will keep looking.

streamer45:
I think the two unrelated issues may have fooled me. The size change was likely a result of https://github.com/mattermost/mattermost-plugin-calls/pull/225. Updating the snapshots there seems to fix the pipeline so I think we should be good now.

PR #229, opened 2022-10-21
Fix e2e tests

cpoile:
#### Summary
- This isn't doing anything to fix the e2e tests, just updating the package lock after running npm i locally.
- ~~But looks like e2e tests are passing now? Or, maybe the gitlab e2e process isn't running on PR branches... So, is there something we should do to fix this? It's not great if we get gitlab e2e failures only after PRs are pushed to main, right?~~ nvm, looks like gitlab e2e just took a while to appear. And the focalboard one is failing...
- @Phrynobatrachus can you take a look at Saturn's reply https://community.mattermost.com/private-core/pl/bffdzhzkafd5b88qdqctfnr3dr and let me know if there are any changes in gitlab we need to make?

#### Ticket Link
- none


streamer45 (approved):

Phrynobatrachus (approved):

cpoile:
keep forgetting update-branch doesn't work, I think because we're calling master main

PR #233, opened 2022-10-31
[MM-48041] Add support for building standalone clients

streamer45:
#### Summary

PR is mostly about renaming the `widget` directory to `standalone` as I hinted at in https://github.com/mattermost/mattermost-plugin-calls/pull/190#issuecomment-1249097721. 

There are however a couple of additions that will help to apply the standalone approach to other instances of clients that don't require the whole webapp loaded in order to render, such as the upcoming call recorder's view.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-48041


standalone/src/init.ts
```diff
@@ -105,21 +96,12 @@ function connectCall(channelID: string, callTitle: string, wsURL: string, iceCon
         window.callsClient = new CallsClient({
             wsURL,
             iceServers: iceConfigs,
+            authToken: getToken(),
```
streamer45, line 99: This is not needed for the widget use case but it will be for the recording view and potentially more use cases in the future (e.g. external calls).

standalone/src/widget.tsx
```diff
@@ -0,0 +1,55 @@
...
+import {
+    logDebug,
+    logErr,
+} from 'plugin/log';
+
+import CallWidget from 'plugin/components/call_widget';
+
+import init from './init';
+
+init('widget', (store, theme) => {
```
cpoile, line 18: cool, the init and close cbs are a nice touch. Is there a functional reason for them, or is it mostly for organization?

cpoile (approved):
Cool! Just one question.

streamer45:
There is a reason since I needed to customize the logic further depending on what standalone client we are rendering. It will become clearer in the recordings PR which will make use of those in a slightly different way (plus more stuff as well). So the callbacks will actually move into a single `InitConfig` argument with various options to extend the logic.

standalone/src/widget.tsx
streamer45: There is a reason since I needed to customize the logic further depending on what standalone client we are rendering. It will become clearer in the recordings PR which will make use of those in a slightly different way (plus more stuff as well). So the callbacks will actually move into a single `InitConfig` argument with various options to extend the logic.

PR #235, opened 2022-11-03
[MM-48142] Fix remaining call state issues on parent window

streamer45:
#### Summary

PR fixes a couple of state related issues in the main window when using the global widget, namely:

- Channel toast should not show after joining the call.
- Leave button should render in the post card.
- Leaving the call from the post card should be allowed, as well as through slash command.
- Switching calls from the post card should work as in web.

#### Related PR

https://github.com/mattermost/desktop/pull/2349

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-48142


cpoile (approved):
Nice!

PR #236, opened 2022-11-04
Track total screen sharing duration per call

streamer45:
#### Summary

Just a little addition to help us figure out how much time is spent screen sharing while in calls.


cpoile (approved):
Good idea. 👍

PR #237, opened 2022-11-04
Replacing screenshare controls with media-chrome

crspeller:
#### Summary
Just a small PR to remove an annoyance I experience when viewing screen shares. 
I tend to click in the middle of the expanded view in order to focus the window. When a screen is being shared this stops the video component. It usually takes me a few seconds to remember and see what happened. 

We decided to use media-chrome to replace the old controls and keep the full-screen functionality, also leaving space for future controls. 

### Before
image

### After
image


cpoile:
We toyed with this in the past, but I think the consensus was having all the controls is annoying but we want to keep the optionfor maximizing the video. If there isn't a way to just have the maximize control (I don't think there is?), the other way we saw at the conference would be to use media chrome: https://www.mux.com/blog/building-the-next-generation-of-video-players-with-media-chrome

streamer45:
> Thanks! We toyed with this in the past, but I think the consensus was having all the controls is annoying but we want to keep the optionfor maximizing the video. If there isn't a way to just have the maximize control (I don't think there is?), the other way we saw at the conference would be to use media chrome: https://www.mux.com/blog/building-the-next-generation-of-video-players-with-media-chrome

Yes, there are a couple of other reasons that demand adding a wrapper like that:

- We'd like to expose a custom button for picture-in-picture (when supported by the browser) which is now hidden behind a mouse right click.
- We need the wrapping element to implement an overlay with the reactions stream when in full screen. https://community.mattermost.com/private-core/pl/3qp3ssufkid3tqfynnjtkpjwbo for context.

crspeller:
If I am understanding correctly, we want to preserve the ability to go full screen with only the video showing? (otherwise you can just use browser full screen). 

If that is the case I will take a look at removing the controls while preserving that functionality for OSF.

streamer45:
> If I am understanding correctly, we want to preserve the ability to go full screen with only the video showing? (otherwise you can just use browser full screen).

Yes, we'd like to retain the full screen button and potentially customize other parts of the controls (like adding a button to trigger picture in picture).

crspeller:
Couldn't resist trying out the media-chrome stuff. Seems to work:

Screenshot 2022-11-09 154442

Was able to easily preserve the full screen button. Tested on Firefox, still need to make sure it works everywhere else.

streamer45 (approved):
Much better now, thanks :+1:

crspeller:
Tested on FF and Desktop. Calls doesn't work for me at all on Chrome. 
On desktop the UI does not appear on hover but that is consistent with previous behavior (intentional?)

I took a look at adding e2e tests for this, but there is no existing pattern for testing screen share. It seems like we would have to add some way to create a fake incoming stream. Maybe a virtual client that connects to the call? Unless we have other ideas on how to test that?

streamer45:
> Calls doesn't work for me at all on Chrome. 

That doesn't sound great. What happens exactly?

> On desktop the UI does not appear on hover but that is consistent with previous behavior (intentional?)

I tried this on Desktop and it was working okay, hover was definitely working :thinking: 

> I took a look at adding e2e tests for this, but there is no existing pattern for testing screen share. It seems like we would have to add some way to create a fake incoming stream. Maybe a virtual client that connects to the call? Unless we have other ideas on how to test that?

Yeah, we are looking into adding more realistic interactions.

crspeller:
> That doesn't sound great. What happens exactly?

Taken to community: https://community.mattermost.com/private-core/pl/q4nzhftfet8q7n5ri3zfupmr1a

> I tried this on Desktop and it was working okay, hover was definitely working 

Strange wonder why my behavior is different.

streamer45:
@tanmay-des Any thought on this?

cpoile (approved):
Awesome, thanks Christopher!

tanmay-des (approved):
Great work @crspeller!
This is certainly something that we wanted to fix, the controls made the share screen experience a lit weird at times.

This looks good to me!

PR #238, opened 2022-11-07
[MM-48074] Implement standalone recording client

streamer45:
#### Summary

PR implements a standalone web client app that will be used to record calls through dedicated jobs running [`calls-recorder`](https://github.com/mattermost/calls-recorder) by opening the following URL:

`https://${SITE_URL}/plugins/com.mattermost.calls/standalone/recording.html?call_id=${CHANNEL_ID}&token=${BOT_AUTH_TOKEN}`

The most notable changes here are about allowing the newly added Calls bot to access calls metadata (e.g. participants) without necessarily being in the channels where such calls are happening as this is a requirement to support recording in DMs/GMs.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-48074


server/channel_state.go
```diff
@@ -37,11 +37,15 @@ type channelState struct {
...
-func (cs *callState) getUsersAndStates() ([]string, []userState) {
+func (cs *callState) getUsersAndStates(botID string) ([]string, []userState) {
 	var i int
 	users := make([]string, len(cs.Users))
 	states := make([]userState, len(cs.Users))
 	for id, state := range cs.Users {
+		// We don't want to expose to the client that the bot is in a call.
+		if id == botID {
+			continue
+		}
```
streamer45, line 48: I went back and forth on whether handle this implementation detail on the client but doing it here seems cleaner overall although we are effectively lying to the client about the state of things.

server/websocket.go
```diff
@@ -391,7 +417,7 @@ func (p *Plugin) handleLeave(us *session, userID, connID, channelID string) erro
 func (p *Plugin) handleJoin(userID, connID, channelID, title string) error {
 	p.LogDebug("handleJoin", "userID", userID, "connID", connID, "channelID", channelID)
 
-	if !p.API.HasPermissionToChannel(userID, channelID, model.PermissionCreatePost) {
+	if !p.isBot(userID) && !p.API.HasPermissionToChannel(userID, channelID, model.PermissionCreatePost) {
```
cpoile, line 420: here too, and below -- but it's totally up to you

server/websocket.go
```diff
@@ -35,6 +35,32 @@ const (
...
 
+func (p *Plugin) publishWebSocketEvent(ev string, data map[string]interface{}, broadcast *model.WebsocketBroadcast) {
+	botID := p.getBotID()
+	// We don't want to expose to the client that the bot is in a call.
+	if (ev == wsEventUserConnected || ev == wsEventUserDisconnected) && data["userID"] == botID {
+		return
+	}
+
+	// If broadcasting to a channel we need to also send to the bot since they
+	// won't be in the channel.
```
cpoile, line 46: really good comment, thank you

standalone/src/recording/components/recording_view/component.tsx
```diff
@@ -0,0 +1,299 @@
...
+
+    render() {
+        const callsClient = this.getCallsClient();
+        if (!callsClient) {
+            return null;
+        }
+
+        const sharingID = this.props.screenSharingID;
+
+        return (
```
cpoile, line 224: I think I've mentioned it before, but I can't help but wonder if there's a way to use the existing component, maybe by adding some flags to its props... Otherwise this is going to be another "platform" we have to support with any UI changes. Probably not for this PR, but maybe soon?

cpoile (approved):
Awesome! Non-blocking comments/questions.

standalone/src/recording/components/recording_view/component.tsx
streamer45: Yeah, I touched on this a bit before but I felt there was too much divergence to make it work through the existing component. The end result of that would have been worse in my mind as it required adding a layer of conditional rendering logic on top of an already overly complex component. 

What I am trying to do instead (in the upcoming PRs) is to share more sub-components between the two views so that changes on one side are reflected on the other side without having to duplicate code and work.

server/api.go
```diff
@@ -55,7 +55,7 @@ func (p *Plugin) handleGetVersion(w http.ResponseWriter, r *http.Request) {
 
 func (p *Plugin) handleGetChannel(w http.ResponseWriter, r *http.Request, channelID string) {
 	userID := r.Header.Get("Mattermost-User-Id")
-	if !p.API.HasPermissionToChannel(userID, channelID, model.PermissionReadChannel) {
+	if !p.isBotSession(r) && !p.API.HasPermissionToChannel(userID, channelID, model.PermissionReadChannel) {
```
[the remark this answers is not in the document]
streamer45: I always liked De Morgan so no objections here.

server/channel_state.go
streamer45: Agreed.

streamer45:
Once the CI passes I will merge this into the [feature branch](https://github.com/mattermost/mattermost-plugin-calls/tree/MM-47182).

PR #239, opened 2022-11-08
using notify v2 demo

AGMETEOR:


cpoile:
Hey @AGMETEOR, is this supposed to go into Calls, or is it something that we can close?

PR #240, opened 2022-11-08
MM-48285 - Emoji reactions in webapp

cpoile:
#### Summary
- This is the webapp portion of the emoji reactions epic. 
- Many many thanks to the contribute team!!!! @Willyfrog @itao @agarciamontoro  @AGMETEOR @cyrilzhang-mm @EricssonLiu (sorry if I missed anyone)
- Probably easiest to review by commit on and after [fb96dbd](https://github.com/mattermost/mattermost-plugin-calls/pull/240/commits/fb96dbdd609632bddddda2eeba0461a80969be86).
- Added the persistent bar.
- Todo: animations on the reaction stream. Allan is working on that. We can work on it together when he has the bandwidth. I would consider it a nice to have.
- UX question for @streamer45 and @tanmay-des -- do you want a click outside the persistent reaction bar to close it, or do you want it to be persistent until purposefully closed (by clicking on the reactions button again). I have code to do the first, but I stuck with the second because I think I prefer it. We might need to try it out to see which we like better.
- @tanmay-des I _might_ be able to style the emoji picker to give it the exact border and bg, but I will have to consult @calebroseland -- it's using a shadow dom element, and I can't figure out a way to style it from it's parent component (which is how it's supposed to be with shadow dom elements). But maybe Caleb knows a secret css ninja trick?
- @tanmay-des I changed the reaction stream bg and text a bit because I noticed that they disappeared if they appeared over a white bg (like when you're watching a screencast). Please take a look and give me revised colors if needed. 
- @streamer45 I can merge this after your upcoming PR, so you don't have to do the merge (which hopefully won't, but might be nasty).
- I have put this up on calls test so you can test it @tanmay-des to see if you like how it turned out. https://calls.test.mattermost.com/octo

image

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-48284
- Designs: https://www.figma.com/file/py15hDM6iSRaa3KgSpQQsC/Calls-Design-Resources?node-id=289%3A43274


agnivade:
Just checking, the emoji picker copied over from webapp right?

server/client_message.go
```diff
@@ -26,6 +26,7 @@ const (
 	clientMessageTypeScreenOff   = "screen_off"
 	clientMessageTypeRaiseHand   = "raise_hand"
 	clientMessageTypeUnraiseHand = "unraise_hand"
+	clientMessageTypeReaction    = "reaction"
```
streamer45, line 29: nit: we tend to prefer verbs so I'd suggest the simpler `react`.

server/websocket.go
```diff
@@ -30,6 +30,7 @@ const (
 	wsEventCallEnd          = "call_end"
 	wsEventUserRaiseHand    = "user_raise_hand"
 	wsEventUserUnraiseHand  = "user_unraise_hand"
+	wsEventUserReact        = "user_reaction"
```
streamer45, line 33:
```suggestion
	wsEventUserReacted        = "user_reacted"
```

server/websocket.go
```diff
@@ -261,6 +262,25 @@ func (p *Plugin) handleClientMsg(us *session, msg clientMessage, handlerID strin
...
+			p.LogError(err.Error())
+		}
+
+		p.API.PublishWebSocketEvent(evType, map[string]interface{}{
+			"userID":        us.userID,
+			"emoji_name":    emoji.Name,
+			"emoji_skin":    emoji.Skin,
+			"emoji_unified": emoji.Unified,
+			"timestamp":     time.Now().UnixMilli(),
+		}, &model.WebsocketBroadcast{ChannelId: us.channelID})
```
streamer45, line 283: I think it's worth making an effort to unify client and server on a single type.

webapp/src/client.ts
```diff
@@ -570,6 +570,14 @@ export default class CallsClient extends EventEmitter {
         this.isHandRaised = false;
     }
 
+    public sendUserReaction(data: EmojiData) {
+        if (this.ws) {
+            this.ws.send('reaction', {
+                data: JSON.stringify(data),
+            });
+        }
```
streamer45, line 578: Would `this.ws?.send(...)` work here to avoid the double nesting?

webapp/src/components/expanded_view/controls_button.tsx
```diff
@@ -22,8 +22,23 @@ export type Props = {
...
+
+    const onClick = () => {
+        if (props.disabled) {
+            return;
+        }
+
+        // @ts-ignore
+        overlayRef.current?.hide();
+        props.onToggle?.();
+    };
```
streamer45, line 35: Could you explain to a mere mortal what's happening here? :p

webapp/src/index.tsx
```diff
@@ -565,11 +573,12 @@ export default class Plugin {
                     });
                 }
 
+                // TODO: we should use types here, could cause trouble in the future.
```
streamer45, line 576: Good idea :100:

webapp/src/reducers.ts
```diff
@@ -215,11 +218,69 @@ interface usersStatusesAction {
...
+interface userReactionsState {
+    [channelID: string]: {
+        reactions: Reaction[],
+    }
+}
+
+const queueReactions = (state: Reaction[], reaction: Reaction) => {
+    const result = state?.length ? [...state] : [];
+    result.push(reaction);
+    if (result.length > 50) {
```
streamer45, line 235: I'd suggest making the hardcoded `50` a more visible constant.

webapp/src/types/types.ts
```diff
@@ -150,3 +152,16 @@ export const CallAlertStatesDefault = {
...
+    name: string;
+    skin?: string;
+    unified: string;
+}
+
+export type Reaction = {
+    emoji: EmojiData;
+    timestamp: number;
+    user_id: string;
+    displayName: string;
```
streamer45, line 166: I struggle to understand why we'd need to embed this vs calculating it from the `user_id` when necessary.

cpoile:
@agnivade Nope, it's a package called `emoji-mart`. I believe it's the same as the one used in focalboard.

agnivade:
Thanks, we really need to prioritize which components are the biggest impact to commonize.

webapp/src/components/expanded_view/reaction_button.tsx
```diff
@@ -0,0 +1,276 @@
...
+
+export const ReactionButton = forwardRef(({trackEvent}: Props, ref) => {
+    const barRef: RefObject<HTMLDivElement> = useRef<HTMLDivElement>(null);
+    const [showPicker, setShowPicker] = useState(false);
+    const [showBar, setShowBar] = useState(false);
+    useImperativeHandle(ref, () => ({
+        toggle() {
+            toggleReactions();
+        },
+    }));
```
streamer45, line 56: Is this actually needed? Mostly for my own understanding but couldn't we simply let the parent provide the handler?

streamer45 (commented):
Thanks! :muscle:  Gave this a first pass. Left a few minor comments here and there.

> do you want a click outside the persistent reaction bar to close it, or do you want it to be persistent until purposefully closed (by clicking on the reactions button again). I have code to do the first, but I stuck with the second because I think I prefer it. We might need to try it out to see which we like better.

Consistency wise I am thinking we should probably do the latter since users would likely expect that to happen.



Some random things I noticed:

I think this was pointed out before but just to confirm, it looks like the emoji set differs between the one displayed and the picker:

image

I'd expect the active state for the button to display slightly differently:

image

server/websocket.go
cpoile: Is it ok if we keep it consistent and use `wsEventUserReact = "user_react"`?

server/websocket.go
cpoile: Could you clarify what you mean?

webapp/src/components/expanded_view/controls_button.tsx
cpoile: Heh -- it's so that once we click on the control button, the tooltip overlay will go away. Before, when you clicked the reactions button, the reaction bar would appear but the tooltip was still there. I have to put ts-ignore because they (react-bootstrap) doesn't publicly expose what we're doing here, it's kind of internal.

webapp/src/reducers.ts
cpoile: Yes, good idea, done! (along with the other magic number we might want to adjust)

webapp/src/types/types.ts
cpoile: This was the bug mentioned in the comments from the contribute team: when a user leaves the call, their reactions are still in the stream, but their profile is no longer in the call => their reactions in the stream switch from "Claudio Costa" to "Somebody". I don't want to remove the reaction from the stream before it times out.
See the note here: https://github.com/mattermost/mattermost-plugin-calls/blob/b894245cef01939418db7c9d59151ddb8ee1c92c/webapp/src/websocket_handlers.ts#L207

webapp/src/components/expanded_view/reaction_button.tsx
cpoile: I have a branch where I've made `ReactionButton` a controlled component. The problem with that is that the expanded_view component is a class component, and any change to its state object forces a complete rerender. With this way, the only component that rerenders is the ReactionButton.
When we convert the expanded_view component to a functional hooks component we can get rid of this. :)

cpoile:
Thanks @streamer45 ! I've responded to the comments.
About the highlighted button -- that _should_ be highlighted after the bar is open. I've tested on Chrome, Firefox, Safari, and they all work. What are you seeing this on?

About the emoji -- the emoji-mart is using the native emojis, and then the emojis displayed on the stream are the ones provided by mattermost, that's why there's a mismatch on linux (and probably windows). Maybe @Willyfrog can help? Do you have any thoughts -- should we make the reaction stream native, or package the mattermost emojis and provide that as the emoji-mart `data`? (if that's even possible)

server/websocket.go
streamer45: `react` would be the client action and `reacted` the event that gets broadcast. But yeah, it's not like we're being consistent with the others so whatever makes sense is fine.

server/websocket.go
streamer45: We are defining an anonymous struct to parse the request and return something entirely different as ws payload. I think there's an opportunity to converge on a single type (with optional fields if needed) and use the same for request and response so that we minimize places where we need to change the casing of fields.

webapp/src/components/expanded_view/controls_button.tsx
streamer45: Interesting, I wonder if a similar approach could perhaps fix https://mattermost.atlassian.net/browse/MM-48336 :thinking:

webapp/src/types/types.ts
streamer45: Surely we could still access the profile through the main redux store? It's not like they left the team or something.

streamer45:
> Thanks @streamer45 ! I've responded to the comments. About the highlighted button -- that _should_ be highlighted after the bar is open. I've tested on Chrome, Firefox, Safari, and they all work. What are you seeing this on?
> 

Tested locally both on Chrome and Firefox. As soon as the mouse moves away (hover ends) the button is not clearly highlighted any longer.

cpoile:
> Tested locally both on Chrome and Firefox. As soon as the mouse moves away (hover ends) the button is not clearly highlighted any longer.

@streamer45 That's really weird, I don't know why that would happen. Can you try on calls.test ?

streamer45:
> > Tested locally both on Chrome and Firefox. As soon as the mouse moves away (hover ends) the button is not clearly highlighted any longer.
> 
> @streamer45 That's really weird, I don't know why that would happen. Can you try on calls.test ?

Indeed it works fine on the test server. I am guessing it could be the webapp version then. Are you using master branch locally?

calebroseland:
Re: styling & Custom Emoji support, how hard would it be to use the same emoji picker we have in Channels? 

CleanShot 2022-11-11 at 12 12 18@2x

cpoile:
@calebroseland Once it's common component-ized and easy to import, sure! Otherwise, it will be too much to try to copy paste that (there must be a reason we built our own from scratch).

cpoile:
> Indeed it works fine on the test server. I am guessing it could be the webapp version then. Are you using master branch locally?

@streamer45 I wasn't, but now I am, and it still works as it should. It's just a css property based on the state of the bar, so I don't understand why yours doesn't work. Can you re-pull and build again just to make sure?

agnivade:
> Re: styling & Custom Emoji support, how hard would it be to use the same emoji picker we have in Channels?

This is something we are tracking over at the MPA effort. Commonizing requires some refactoring to reduce the dependency on redux. I am collecting a list of components from different teams that we need to commonize, and it seems like the emoji picker is one such thing that likely has high impact with medium effort.

cc @hmhealey

Willyfrog:
The Channels emoji picker was built around channels, Boards would be using that one if it was easy to import. So, during contribute we didn't have a ton of time so we took the other way around, used the same library as Boards uses but since we were already thinking about the future, we used the display component from Channels, so once we move into reusing the Channels one the custom emojis would be free.

Regarding custom emoji support, it is possible to add custom emojis to emoji mart, we would just need to add them as one of the properties of the component. It needs conversion though. See https://github.com/missive/emoji-mart#custom-emojis for docs. Still I'd prefer to work on commonizing the channels emoji picker rather than add the custom ones into the library.

Regarding the differences, on Channels we default to the Apple ones, os if you are looking at one of the other sources (only macs are suppossed to display Apple emojis), this might render differently. Changing the `set` property and forcing to use Mac, should "fix" the issue

streamer45:
> @streamer45 I wasn't, but now I am, and it still works as it should. It's just a css property based on the state of the bar, so I don't understand why yours doesn't work. Can you re-pull and build again just to make sure?

Yeah, I am on the latest commit here but it still happens. I know all other buttons backgrounds were not working (unmuted/screen sharing). I think I fixed this in https://github.com/mattermost/mattermost-plugin-calls/pull/242 but no idea if it's related to this as well. Anyhow, let's wait till everything is merged nicely, it's not a big deal at this time.

cpoile:
@Willyfrog Awesome, that `set` prop is probably exactly what we needed. Thanks!
As for the custom emojis -- I think we'll be fine without them until we get the common component.

server/websocket.go
cpoile: Ah, yes I see what you mean. I use the `toMap` pattern from one of your PRs, and now sending/receiving the ws data is a lot cleaner, thanks.

webapp/src/components/expanded_view/controls_button.tsx
cpoile: Yep, probably. I'll take a look at that. 👍

webapp/src/types/types.ts
cpoile: This was an interesting one. It turns out we don't have access to all the profiles in the ExpandedView -- since it's a `NeedsTeamRoute` mounted component, we would have to manually request all the known user profiles from the backend ourselves (possibly hundreds), and dispatch them to the store, so that they're there when we call `getUsers` selector. Since we don't actually honour the user's teammate displayname setting (should I make a ticket to fix that?), and we don't actually use the store's up-to-date version of the user profiles, we don't display any changes to the user display names once they've joined a call. They can change their names, or we can change our teammate displayname preferences, and there won't be any change in the Calls components. 
So, given all of that -- I like this as a performant way of calculating and storing the displayname once, rather than recalculating it on every re-render (with no change).

cpoile:
@tanmay-des I've updated the plugin in calls.test.mattermost.com so you can review it. Thanks!

tanmay-des:
Firstly @cpoile Great work! This looks really good!

I have a few items in mind that might make the experience a little better: 

**1. Reaction Chip Design**
I think the initial chip design had 10px font and that might've looked really tiny. I have adjusted the chip [design here](https://www.figma.com/file/2hVzrN84rXRi7kbw2Wbr1h/Calls%3A-Reactions?node-id=141%3A35321) using 12px font and adjusted padding. 

- [ ] Adjusted size and padding
- [ ] Background color and text color (for reactions)
- [ ] Icon for the raise hand (using the compass one)
- [ ] The raise hand chip should have user's name in **semi-bold**

Screenshot 2022-11-16 at 12 38 31 PM_(Image for reference)_


**2. React menu**
Two very small improvements [here is the design file as well](https://www.figma.com/file/2hVzrN84rXRi7kbw2Wbr1h/Calls%3A-Reactions?node-id=131%3A31705):

- [ ] New outlined compass icons for the raise hand
- [ ] Hover state for the raise hand and other buttons

Screenshot 2022-11-16 at 12 43 15 PM    _(Image for reference)_ 

**3. Animations**
I am experimenting with adding subtle transitions for these [reactions here](https://www.figma.com/file/2hVzrN84rXRi7kbw2Wbr1h/Calls%3A-Reactions?node-id=141%3A35321)
reactions-animations-small

webapp/src/types/types.ts
streamer45: Alright. At some point we'll have to request the profiles of any user in a call if we ever migrate to use the standalone view for this, although not something we need to deal with today.

streamer45 (approved):
Looks good :+1:

cpoile:
@tanmay-des All done (except for animations -- v2):

image
image

Heart hovered:
image

streamer45:
@cpoile Is there anything outstanding here? I'd propose merging this asap so I can start working on any conflict with the recordings one which is not 100% ready yet.

cpoile:
@streamer45 Yep, ready now. I'll merge when tests are finished. I can do the conflict resolution for #242 -- that would save you a lot of work.

streamer45:
> @streamer45 Yep, ready now. I'll merge when tests are finished. I can do the conflict resolution for #242 -- that would save you a lot of work.

:+1:  Just so you know e2e tests may fail here, see https://github.com/mattermost/mattermost-plugin-calls/pull/242/commits/b5ae5a12ad8a9a3c452bbae6c0cd6a594ca6a31e

#242 needs to go into the feature branch and get reviewed by security still. I'll open a PR as soon as we merge everything and then we can work together if it gets complex.

PR #242, opened 2022-11-11
[MM-48219] Implement recordings web UX flow

streamer45:
#### Summary

This should account for most of the web implementation and a good portion of back-end as well.

The most notable missing parts here are the interaction with the recording service (to be implemented) and some minor work on the post recording flow (e.g. publishing recordings and updating the call thread). 

The PR is quite dense as I took this as a chance to refactor some things slightly when it made sense to do so. Please let me know if anything is unclear as I am happy to go through the logic which at times gets far from being trivial.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-48219


server/websocket.go
```diff
@@ -547,6 +549,13 @@ func (p *Plugin) handleJoin(userID, connID, channelID, title string) error {
 		}, &model.WebsocketBroadcast{ChannelId: channelID, ReliableClusterSend: true})
 	}
 
+	if userID == p.getBotID() && state.Call.Recording != nil {
+		p.publishWebSocketEvent(wsEventCallRecordingStart, map[string]interface{}{
+			"callID":  channelID,
+			"startAt": state.Call.Recording.StartAt,
+		}, &model.WebsocketBroadcast{ChannelId: channelID, ReliableClusterSend: true})
+	}
```
cpoile, line 556: I suppose this could cause interesting states, if we send this message but the recording actually fails to start. Would it be better to send this only after the recording system has successfully started? (It would let us also send "failed" messages (with error/debug info too?), and hosts could try to start the recording again)

webapp/src/components/call_widget/component.tsx
```diff
@@ -1183,7 +1184,8 @@ export default class CallWidget extends React.PureComponent<Props, State> {
     }
 
     renderRecordingDisclaimer = () => {
-        if (this.state.recDisclaimerDismissedAt) {
+        if (!this.props.callRecordingStartAt ||
+            this.state.recDisclaimerDismissedAt > this.props.callRecordingStartAt) {
```
cpoile: Should this check be used before calling the `renderRecordingDisclaimer` fn?

webapp/src/components/expanded_view/component.tsx
```diff
@@ -430,7 +431,8 @@ export default class ExpandedView extends React.PureComponent<Props, State> {
     }
 
     renderRecordingDisclaimer = () => {
-        if (this.state.recDisclaimerDismissedAt) {
+        if (!this.props.callRecordingStartAt ||
+            this.state.recDisclaimerDismissedAt > this.props.callRecordingStartAt) {
```
cpoile: likewise

webapp/src/components/badge.tsx
```diff
@@ -0,0 +1,79 @@
...
+            }
+            { !props.loading &&
+            <Icon fill={props.iconFill}>{props.icon}</Icon>
+            }
+            <Text gap={props.gap}>{props.text}</Text>
+        </Container>
+    );
+}
+
+const Container = styled.div<{bgColor?: string, size: number, margin?: string, padding?: string, color?: string}>`
```
cpoile, line 38: This would be cleaner if we had a `ContainerProps` interface, imo.

webapp/src/components/badge.tsx
```diff
@@ -0,0 +1,79 @@
...
+  font-weight: 600;
+  margin-left: ${({gap}) => gap}px;
+`;
+
+const Icon = styled.div<{fill?: string}>`
+  display: flex;
+  fill: ${({fill}) => fill || 'currentColor'};
+`;
+
+const Spinner = styled.span<{size: number}>`
```
cpoile, line 61: nice

webapp/src/components/expanded_view/component.tsx
```diff
@@ -632,6 +675,36 @@ export default class ExpandedView extends React.PureComponent<Props, State> {
...
+        return (
+            <Badge
+                text={'REC'}
+                textSize={12}
+                gap={6}
+                margin={'0 12px 0 0'}
+                padding={'8px 6px'}
+                icon={(<RecordCircleOutlineIcon size={12}/>)}
+                bgColor={hasRecStarted ? '#D24B4E' : 'rgba(221, 223, 228, 0.04)'}
+                loading={!hasRecStarted}
```
cpoile, line 703: ah, I see -- we want to send out the call recording ws event, and then send out the `start_at` after, so we have UX feedback while it's getting set up. Gotcha.

cpoile (commented):
Very cool! Just a few comments/questions, nothing major.

server/api.go
```diff
@@ -245,7 +211,6 @@ func (p *Plugin) handleEndCall(w http.ResponseWriter, r *http.Request, channelID
 		return
 	}
 
-	p.metrics.IncWebSocketEvent("out", "call_end")
```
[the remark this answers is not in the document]
streamer45: I believe so, since I refactored the call into a wrapper we now send those metrics in one place only.

https://github.com/mattermost/mattermost-plugin-calls/blob/e34a56cb03085e0e3318bb91fae93a4fd45f0816/server/websocket.go#L64

server/channel_state.go
```diff
@@ -39,18 +41,114 @@ type channelState struct {
...
+	Enabled   bool       `json:"enabled"`
+	Call      *CallState `json:"call,omitempty"`
+}
+
+func (cs *callState) Clone() *callState {
+	if cs == nil {
+		return nil
+	}
+
+	newState := *cs
```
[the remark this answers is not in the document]
streamer45: Fair enough, I can add the `Client` suffix to the public ones if it helps.

server/channel_state_test.go
```diff
@@ -0,0 +1,262 @@
...
+			},
+		}
+
+		cloned := cs.Clone()
+		require.Equal(t, cs, cloned)
+		require.NotSame(t, cs, cloned)
+
+		require.Condition(t, func() bool {
+			return cs.Call != cloned.Call
+		})
```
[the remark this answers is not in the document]
streamer45: Because those fields are pointers and I want to make sure they are not pointing to the original object which is a problem we had before with `channelState.Call` (it was a bug really but luckily didn't affect anything).

webapp/src/components/call_widget/component.tsx
streamer45: That's a question I asked myself as well. I don't know what's the best practice but I was trying to avoid making the code there even heavier so opted to keep everything inside.

webapp/src/components/badge.tsx
streamer45: Yeah, it was originally one thing, then it became two and here we are. I'll fix it :+1:

webapp/src/components/expanded_view/component.tsx
streamer45: Yeah, since it's clearly an async operation I wanted to get that kind of feedback and potentially error out as well. But I'll review all this once I have the service logic plugged in.

webapp/src/reducers.ts
```diff
@@ -395,28 +398,60 @@ const voiceUsersStatuses = (state: usersStatusesState = {}, action: usersStatuse
...
+        return state;
+    }
+};
+
 interface callState {
     channelID: string,
-    startAt: number,
-    ownerID: string,
+    startAt?: number,
+    ownerID?: string,
```
cpoile, line 426: Just wondering -- why were these changed to optional?

streamer45: Because we are using the same type but the server can send a partial state (e.g. only the hostID). See `handleCallHostChanged`.

cpoile (approved):

PR #243, opened 2022-11-11
[MM-48351] Fix screen stream flakyness

streamer45:
#### Summary

I remember spending some good time on implementing this logic but looking back at it now it seems overly complex and easily error prone. I think relying on track ids to figure out if something has changed is a better approach which seems to be working fine.

This PR should fix https://mattermost.atlassian.net/browse/MM-48351 as well.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-48351

### DIFF
--- a/webapp/src/components/call_widget/component.tsx
+++ b/webapp/src/components/call_widget/component.tsx
@@ -436,14 +436,14 @@ export default class CallWidget extends React.PureComponent<Props, State> {
             screenStream = window.callsClient.getLocalScreenStream();
         }
 
-        const wasRendering = Boolean(prevProps.screenSharingID && prevState.screenStream && prevProps.show);
-        const shouldRender = Boolean(this.props.screenSharingID && screenStream);
-
-        if (!this.state.screenStream && screenStream) {
+        const hasScreenTrackChanged = screenStream && this.state.screenStream && this.state.screenStream.getVideoTracks()[0].id !== screenStream.getVideoTracks()[0].id;
+        if ((screenStream && !this.state.screenStream) || hasScreenTrackChanged) {
             // eslint-disable-next-line react/no-did-update-set-state
             this.setState({screenStream});
-        } else if (!wasRendering && shouldRender && this.screenPlayer.current) {
-            this.screenPlayer.current.srcObject = screenStream;
+        }
+
+        if (this.state.screenStream && this.screenPlayer.current && this.screenPlayer?.current.srcObject !== this.state.screenStream) {
+            this.screenPlayer.current.srcObject = this.state.screenStream;
         }
 
         let profiles;
--- a/webapp/src/components/call_widget/component.tsx
+++ b/webapp/src/components/call_widget/component.tsx
@@ -436,13 +436,13 @@ export default class CallWidget extends React.PureComponent<Props, State> {
             screenStream = window.callsClient.getLocalScreenStream();
         }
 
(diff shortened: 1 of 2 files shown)


webapp/src/components/call_widget/component.tsx
```diff
@@ -436,14 +436,14 @@ export default class CallWidget extends React.PureComponent<Props, State> {
             screenStream = window.callsClient.getLocalScreenStream();
         }
 
-        const wasRendering = Boolean(prevProps.screenSharingID && prevState.screenStream && prevProps.show);
-        const shouldRender = Boolean(this.props.screenSharingID && screenStream);
-
-        if (!this.state.screenStream && screenStream) {
+        const hasScreenTrackChanged = screenStream && this.state.screenStream && this.state.screenStream.getVideoTracks()[0].id !== screenStream.getVideoTracks()[0].id;
```
cpoile, line 439: Could this be simplified?
```suggestion
        const hasScreenTrackChanged = screenStream && this.state.screenStream?.getVideoTracks()[0].id !== screenStream.getVideoTracks()[0].id;
```

webapp/src/components/call_widget/component.tsx
```diff
@@ -436,14 +436,14 @@ export default class CallWidget extends React.PureComponent<Props, State> {
...
+        if ((screenStream && !this.state.screenStream) || hasScreenTrackChanged) {
             // eslint-disable-next-line react/no-did-update-set-state
             this.setState({screenStream});
-        } else if (!wasRendering && shouldRender && this.screenPlayer.current) {
-            this.screenPlayer.current.srcObject = screenStream;
+        }
+
+        if (this.state.screenStream && this.screenPlayer.current && this.screenPlayer?.current.srcObject !== this.state.screenStream) {
```
cpoile, line 445: Since screenPlayer will always have a ref, the `?` operator should be after the current (which may or may not have been attached to its target):
```suggestion
        if (this.state.screenStream && this.screenPlayer.current?.srcObject !== this.state.screenStream) {
```

webapp/src/components/expanded_view/component.tsx
```diff
@@ -336,10 +336,10 @@ export default class ExpandedView extends React.PureComponent<Props, State> {
             this.screenPlayer.current.srcObject = this.state.screenStream;
         }
 
-        const callsClient = this.getCallsClient();
-        if (!this.state.screenStream && callsClient?.getLocalScreenStream()) {
+        const localScreenStream = this.getCallsClient()?.getLocalScreenStream();
+        if (localScreenStream && (!this.state.screenStream || this.state.screenStream.getVideoTracks()[0].id !== localScreenStream.getVideoTracks()[0].id)) {
```
cpoile, line 340: Same as the first comment, not sure if it changes the semantics, but can this be simplified to:
```suggestion
        if (localScreenStream && (this.state.screenStream?.getVideoTracks()[0].id !== localScreenStream.getVideoTracks()[0].id)) {
```

cpoile (approved):
👍

webapp/src/components/call_widget/component.tsx
streamer45: I think so although it'd be purely a syntactic simplification rather than a logical one.

webapp/src/components/call_widget/component.tsx
streamer45: Thanks :+1:

PR #247, opened 2022-11-25
[MM-48220] Job service

streamer45:
#### Summary

PR adds the glue code needed to interface with the offloader job service.

#### Related PR

https://github.com/mattermost/calls-offloader/pull/1

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-48220


cpoile (approved):
Nice!
I wonder if we should handle the "everyone has left the call, so stop recording" case?

streamer45:
> Nice!
> 
> I wonder if we should handle the "everyone has left the call, so stop recording" case?

Probably but first maybe it's worth adding some UI for ending the call (and recording as direct consequence). I'll look into these cases next week, thanks!

PR #249, opened 2022-11-29
[MM-48668] Recording state improvements

streamer45:
#### Summary

A couple of changes to the state handling. Given calls to the job service can take a while we anticipate sending the websocket events so that clients don't feel like "nothing is happening". We also reset the state in case of any error.

I am still thinking whether to implement an additional "stopping" state since that operation could also take a bit to complete but for now this will do.

Also updating to the latest version of `calls-offloader` which will load the increased client timeouts as defined in https://github.com/mattermost/calls-offloader/pull/2


streamer45:
Just pushed the additional change we discussed, automatically stopping the recording (and ending the call as a result) if everyone leaves the call.

streamer45:
Pushed a couple more fixes (including [MM-48293](https://mattermost.atlassian.net/browse/MM-48293) since I was touching the same code).

PR #250, opened 2022-11-29
[MM-48219] Recordings: outstanding UI changes

streamer45:
#### Summary

PR implements some outstanding UI additions for recordings (see commits for details).

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-48219


streamer45:
Remaining e2e tests are failing due to the in product notice appearing. I am taking care of this in https://git.internal.mattermost.com/qa/calls-e2e-testing/-/merge_requests/8

webapp/src/components/call_widget/widget_banner.tsx
```diff
@@ -165,8 +182,25 @@ const Footer = styled.div`
 `;
 
 const ConfirmButton = styled.button`
-  color: var(--button-bg);
-  font-weight: 600;
+  &&& {
```
cpoile, line 185: Do we actually need the extra specificity here?

webapp/src/components/icons/raised_hand.tsx
```diff
@@ -13,14 +13,18 @@ export default function RaisedHandIcon(props: Props) {
     return (
         <svg
             {...props}
-            width='25px'
-            height='27px'
-            viewBox='0 0 25 27'
+            width='24px'
```
cpoile, line 16: (this is a comment for the whole thing) -- can't we use the compass-icons (and remove these svgs completely)? (I think I did that in another PR, but I've forgotten which)

cpoile (approved):
Nice! Just a couple non-blocking comments

webapp/src/components/call_widget/widget_banner.tsx
streamer45: At the moment we do unless we refactor some stuff as it was conflicting with definitions in https://github.com/mattermost/mattermost-plugin-calls/blob/da019884e890fed6a2333287c176f3504c16c706/webapp/src/components/call_widget/component.scss

webapp/src/components/icons/raised_hand.tsx
streamer45: Maybe. To be honest I battled hard with certain compass icons when it came to precise sizing and alignment. They are adding layers of definitions that are making it extra difficult to customize. Having the svgs included keeps it simpler and as we have seen recently, doesn't depend on webapp versions, making the product self sufficient in a way. Overall I think this needs a broader discussion and some general review of the whole product as we are now in a rather inconsistent state.

Created https://mattermost.atlassian.net/browse/MM-48854

streamer45:
Will merge this to unblock other reviews.

PR #254, opened 2022-11-30
[MM-48746] Implement basic error surfacing

streamer45:
#### Summary

PR adds some basic error handling to cover for the following failure cases:

1. Recording fails to start immediately (e.g. job service returns an error).
2. Recording job times out (e.g. recording bot fails to join the call).
3. Recording fails to stop.

@tanmay-des I opted to actually relay the server side error to the client in the hope it will give them more information but I also acknowledge it's not really actionable on their part as they likely need a system/infrastructure admin to intervene. Please let me know if you'd like it reverted to the generic "something has failed".

/cc @itao 

#### Screenshots

image

image

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-48746

### DIFF
--- a/server/recording_api.go
+++ b/server/recording_api.go
@@ -15,6 +15,57 @@ import (
 
 var callRecordingActionRE = regexp.MustCompile(`^\/calls\/([a-z0-9]+)/recording/(start|stop|publish)$`)
 
+const recordingJobStartTimeout = 15 * time.Second
+
+func (p *Plugin) recJobTimeoutChecker(callID, jobID string) {
+	time.Sleep(recordingJobStartTimeout)
+
+	state, err := p.kvGetChannelState(callID)
+	if err != nil {
+		p.LogError("failed to get channel state", "error", err.Error())
+		return
+	}
+
+	recState, err := state.getRecording()
+	if err != nil {
+		p.LogError("failed to get recording state", "error", err.Error())
+		return
+	}
+
+	// If the recording hasn't started (bot hasn't joined yet) we notify the
+	// client.
+	var clientState *RecordingStateClient
+	if recState.JobID == jobID && recState.StartAt == 0 {
+		if err := p.kvSetAtomicChannelState(callID, func(state *channelState) (*channelState, error) {
+			recordingState, err := state.getRecording()
+			if err != nil {
+				return nil, err
+			}
+			if recordingState.JobID != jobID {
+				return nil, fmt.Errorf("invalid recording job")
+			}
+
+			clientState = recordingState.getClientState()
+			state.Call.Recording = nil
+
+			return state, nil
+		}); err != nil {
+			p.LogError("failed to set channel state", "error", err.Error())
+			return
+		}
+
+		p.LogError("timed out waiting for recorder bot to join", "callID", callID, "jobID", jobID)
+
+		clientState.Err = "failed to start recording job: timed out waiting for bot to join call"
+		clientState.EndAt = time.Now().UnixMilli()
+
+		p.publishWebSocketEvent(wsEventCallRecordingState, map[string]interface{}{
+			"callID":   callID,
+			"recState": clientState.toMap(),
+		}, &model.WebsocketBroadcast{ChannelId: callID, ReliableClusterSend: true})
+	}
+}
+
 func (p *Plugin) handleRecordingAction(w http.ResponseWriter, r *http.Request, callID, action string) {
 	var res httpResponse
 	defer p.httpAudit("handlePostRecording", &res, w, r)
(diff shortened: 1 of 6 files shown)


tanmay-des (approved):
I think it's always better to give some context instead of no-context.
Though I think maybe we can put the error at the end of the prompt, in italics or different font maybe? 

[Design link](https://www.figma.com/file/9HL2HADCQ0lLmn1RBOBGGV/Calls%3A-Recording?node-id=277%3A119853&t=Nlxp86NuLfxyLjue-1)
image

streamer45:
> I think it's always better to give some context instead of no-context. Though I think maybe we can put the error at the end of the prompt, in italics or different font maybe?
> 
> [Design link](https://www.figma.com/file/9HL2HADCQ0lLmn1RBOBGGV/Calls%3A-Recording?node-id=277%3A119853&t=Nlxp86NuLfxyLjue-1) image

Sounds like a great idea, thanks :+1:

webapp/src/components/expanded_view/component.tsx
```diff
@@ -506,22 +508,40 @@ export default class ExpandedView extends React.PureComponent<Props, State> {
...
+            />);
 
         if (hasRecEnded) {
             header = 'Recording has stopped. Processing...';
             body = 'You can find the recording in this call\'s chat thread once it\'s finished processing.';
         }
 
+        let error = '';
+        if (this.props.callRecording?.err) {
+            header = 'Something went wrong with the recording!';
```
justinegeffen, line 523:
```suggestion
            header = 'Something went wrong with the recording';
```

webapp/src/components/expanded_view/component.tsx
```diff
@@ -506,22 +508,40 @@ export default class ExpandedView extends React.PureComponent<Props, State> {
...
+            body = 'Please try to record again. If the error persists please ask your system administrator.';
```
justinegeffen, line 524:
```suggestion
            body = 'Please try to record again. You can also contact your system admin for troubleshooting help.';
```

webapp/src/components/expanded_view/component.tsx
justinegeffen: Do we know when this error would come up - is it possible to get this in the middle of a call that's being recorded? If so, what happens to the recording up until then? Is it saved or discarded?

justinegeffen (commented):
Some feedback questions for copy review. Thank you!

webapp/src/components/expanded_view/component.tsx
streamer45: The way it's implemented so far the error would only show upon starting/stopping the recording. But that's a limitation since technically the recording process could fail at any time (although it would be a rare event). In such cases the file would likely be discarded until we put in place a more advanced system (past beta phase).

server/recording_api.go
```diff
@@ -15,6 +15,57 @@ import (
...
+	var clientState *RecordingStateClient
+	if recState.JobID == jobID && recState.StartAt == 0 {
+		if err := p.kvSetAtomicChannelState(callID, func(state *channelState) (*channelState, error) {
+			recordingState, err := state.getRecording()
+			if err != nil {
+				return nil, err
+			}
+			if recordingState.JobID != jobID {
+				return nil, fmt.Errorf("invalid recording job")
+			}
```
cpoile, line 46: would this ever happen? (multiple jobs started I guess?) -- but if it does, what happens to that invalid job? Should we do cleanup?

cpoile (approved):
Nice, this will be a much better UX.

server/recording_api.go
streamer45: This is purely to prevent a potential edge case due to a race condition. We check above that the job matches but by the time we read from K/V store the recording could have stopped and a new one started. At this point we don't want to stop the new one as the request was meant for the previous one.

streamer45:
Going to merge this to facilitate final review and testing. I took notes of a couple of improvements based on reviews and will apply them separately.

server/recording_api.go
```diff
@@ -15,6 +15,57 @@ import (
 
 var callRecordingActionRE = regexp.MustCompile(`^\/calls\/([a-z0-9]+)/recording/(start|stop|publish)$`)
 
+const recordingJobStartTimeout = 15 * time.Second
+
```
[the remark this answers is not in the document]
streamer45: Actually, the timeout I was considering here would be the HTTP client one which was taken care of in https://github.com/mattermost/mattermost-plugin-calls/pull/249 through bumping dependencies since this checker goroutine only starts once the HTTP request returns so that means the recording job (docker container) has started and 15 seconds maps well to the [existing one](https://github.com/mattermost/calls-recorder/blob/dd30b8b78c6d704394dcfc46de584df01fe30afe/cmd/recorder/recorder.go#L23) in `calls-recorder`.

PR #262, opened 2022-12-06
Fix potential edge case

streamer45:
#### Summary

Fixing a potential edge case where in case of a timeout (or other unexpected error) while making the start job request, the recording job process could still run and the bot start recording without anyone noticing. So we add an explicit check to make sure the bot is only allowed in if a recording has been requested but not started yet.


cpoile (approved):
That's a good idea

PR #263, opened 2022-12-06
Fix e2e tests on MacOS

streamer45:
#### Summary

Screenshot 2022-12-06 at 11 37 52

It's still not perfect as I found the executions to be a little flaky but these changes should resolve most of the issues.

Some important points to note are:

- Boards should be enabled (as there's a test for this).
- Mattermost server and webapp should be latest version (there was a recent bug making the call thread to fail loading).
- I had some issues achieving RTC connectivity. I removed the STUN server from config and set the ice override to `localhost` which seems to have helped.


cpoile:
Phew, they work. They weren't, and I tried a lot of things, and then I finally deleted the db and started fresh.  LGTM!

cpoile (approved):

cpoile:
Gonna merge so I can use them in the GA PR without future conflicts.

PR #267, opened 2022-12-06
[MM-48898] Implement info prompt for approaching max call duration

streamer45:
#### Summary

PR implements an info prompt that's showed to the host when the current call recording duration is approaching (10 minutes threshold) the configured limit.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-48898

### DIFF
--- a/server/configuration.go
+++ b/server/configuration.go
@@ -73,6 +73,12 @@ type clientConfig struct {
 	MaxRecordingDuration *int
 }
 
+const (
+	defaultRecDurationMinutes = 60
+	minRecDurationMinutes     = 15
+	maxRecDurationMinutes     = 180
+)
+
 type ICEServers []string
 type ICEServersConfigs rtc.ICEServers
 
@@ -194,7 +200,7 @@ func (c *configuration) SetDefaults() {
 		c.EnableRecordings = new(bool)
 	}
 	if c.MaxRecordingDuration == nil {
-		c.MaxRecordingDuration = model.NewInt(60)
+		c.MaxRecordingDuration = model.NewInt(defaultRecDurationMinutes)
 	}
 }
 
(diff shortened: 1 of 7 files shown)


server/configuration.go
```diff
@@ -73,6 +73,12 @@ type clientConfig struct {
 	MaxRecordingDuration *int
 }
 
+const (
+	defaultRecDurationMinutes = 60
+	minRecDurationMinutes     = 15
+	maxRecDurationMinutes     = 180
```
streamer45, line 79: @justinegeffen I am temporarily setting these as min/max limits. We can discuss tomorrow in our call if they need tuning and then can be added to the documentation.

webapp/src/components/expanded_view/recording_info_prompt.tsx
```diff
@@ -0,0 +1,153 @@
...
+                        icon='alert-outline'
+                        style={{
+                            fontSize: 18,
+                        }}
+                    />
+                }
+                iconFill='rgb(var(--dnd-indicator-rgb))'
+                iconColor='rgb(var(--dnd-indicator-rgb))'
+                header={`Your recording will end in ${getMinutesLeftBeforeEnd()} minutes`}
+                body={`Calls can be recorded for up to ${props.recordingMaxDuration} minutes.`}
```
streamer45, line 77: @justinegeffen Please help with the right content to show here.

cpoile (approved):

streamer45:
Given Justine is out I am going to merge this and we can review it post release if necessary.

server/configuration.go
justinegeffen: Brilliant, that works for me. Thanks.
(i missed the meeting as I was in surgery, sorry!)

webapp/src/components/expanded_view/recording_info_prompt.tsx
justinegeffen: This reads well to me. I wonder if we should switch the body and header?

header={`Calls can be recorded for up to ${props.recordingMaxDuration} minutes.`}
body={`Your recording will end in ${getMinutesLeftBeforeEnd()} minutes`}

server/configuration.go
streamer45: No worries, hope everything went well :)

webapp/src/components/expanded_view/recording_info_prompt.tsx
streamer45: Yep, we can do that. Let me update on the main PR. Thanks!

PR #269, opened 2022-12-07
MM-48859 - Fix for: Emoji picker doesn't open after opened once

cpoile:
#### Summary
- This is a staggering hack. The alternative is to just not close emoji picker when clicking outside of it; that would remove all of this craziness. But I think closing it when clicking out is important UX, so we're left with this.
- Oh, the PR that broke it was webapp upgrading to react 17, so I went through and upgraded us as well (to see if that was the problem). It wasn't. But it does trigger this issue, so 🤷 

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-48859


streamer45:
Merging this to unblock final testing.

PR #270, opened 2022-12-07
[MM-48928] Outstanding UX for recordings

streamer45:
#### Summary

PR addresses some outstanding UX feedback (see https://www.figma.com/file/9HL2HADCQ0lLmn1RBOBGGV/Calls%3A-Recording?node-id=298%3A120956&t=0Aq13EORRxvt3YoH-0 for details).

In addition we are at least partially addressing the issue of a user not being aware of becoming the host. We are showing the "you are recording this call" banner again upon host change even if the banner was dismissed prior (as non-host user). Ideally we should consider a custom message for this but we didn't get time to design further so happy to have something rather than nothing.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-48928

### DIFF
--- a/webapp/src/components/call_widget/component.tsx
+++ b/webapp/src/components/call_widget/component.tsx
@@ -82,6 +82,7 @@ interface Props {
     },
     callStartAt: number,
     callHostID: string,
+    callHostChangeAt: number,
     callRecording?: CallRecordingState,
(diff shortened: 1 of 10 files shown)


streamer45:
> Also: appearing on top of emojis

Indeed :+1:

PR #271, opened 2022-12-07
Fix missing ws events

streamer45:
#### Summary

Almost randomly spotted this while testing recordings. Users were leaving calls but it was not reflected in the recording view and in the output file as a result. 

We need to make extra sure to always use the provided `publishWebSocketEvent` wrapping method so that the call bot receives the ws events.


cpoile (approved):
ah, gotcha

PR #273, opened 2022-12-08
Fix trailing slash in services URLs

streamer45:
#### Summary

PR fixes an issue causing the plugin to fail to start if the provided URL (either `rtcd` or `calls-offloader`) has a trailing slash.

Thanks @oh6hay for spotting this :100:

### DIFF
--- a/server/rtcd.go
+++ b/server/rtcd.go
@@ -10,6 +10,7 @@ import (
 	"net"
 	"net/url"
 	"os"
+	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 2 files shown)


server/rtcd.go
```diff
@@ -299,6 +300,9 @@ func resolveURL(u string, timeout time.Duration) ([]net.IP, string, error) {
 }
 
 func (m *rtcdClientManager) newRTCDClient(rtcdURL, host string, dialFn rtcd.DialContextFn) (*rtcd.Client, error) {
+	// Remove trailing slash if present.
+	rtcdURL = strings.TrimSuffix(rtcdURL, "/")
```
cpoile, line 304: Shouldn't this one place be enough, and keep the others as is? 
I mean, it doesn't matter in the end, just curious.

cpoile (approved):

server/rtcd.go
streamer45: Those are two completely different services and URLs (job service and rtcd). And we need to do it a third time since there we support an environment override for `MM_CALLS_JOB_SERVICE_URL`.

PR #276, opened 2022-12-13
[MM-46048] Implement glare-free negotiation

streamer45:
#### Summary

PR implements glare-free negotiation using the proposed scheme in https://developer.mozilla.org/en-US/docs/Web/API/WebRTC_API/Perfect_negotiation (polite peer case).

In practical terms this avoids a whole set of issues involving clients receiving tracks while trying to send theirs. The most glaring example we have in product would be joining calls unmuted, which led to the unfortunate "sleep a few seconds" hack.

I've load tested a bit to cover a few scenarios and it seems like working as expected. I am still not entirely sure whether this covers all potential racing issues with signaling but it definitely improves the most common ones.

#### Related PRs

https://github.com/mattermost/rtcd/pull/80

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-46048

### DIFF
--- a/webapp/src/components/call_widget/component.tsx
+++ b/webapp/src/components/call_widget/component.tsx
@@ -381,14 +381,9 @@ export default class CallWidget extends React.PureComponent<Props, State> {
             }
 
             if (isDirectChannel(this.props.channel) || isGroupChannel(this.props.channel)) {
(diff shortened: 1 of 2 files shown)


cpoile (approved):
> The most _glaring_ example...

Hehe

PR #277, opened 2022-12-14
MM-48657 - Add close button to pop-out window on desktop

cpoile:
#### Summary
- Can you double check this doesn't break anything? I've tested it and it works. But I'm a little worried it was so easy -- is there a reason it wasn't included in the first place? 

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-48657

### DIFF
--- a/e2e/tests/popout.spec.ts
+++ b/e2e/tests/popout.spec.ts
@@ -1,4 +1,4 @@
-import {test, expect, chromium} from '@playwright/test';
+import {test, expect} from '@playwright/test';
 
 import PlaywrightDevPage from '../page';
 import {userState} from '../constants';
--- a/webapp/src/components/call_widget/component.tsx
+++ b/webapp/src/components/call_widget/component.tsx
@@ -1,3 +1,4 @@
+/* eslint-disable max-lines */
 import React, {CSSProperties} from 'react';
 import {OverlayTrigger, Tooltip} from 'react-bootstrap';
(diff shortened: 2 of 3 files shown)


streamer45:
It simply wasn't implemented because originally the expanded window on Desktop (global widget) had a frame around it so you could close it like any other window but I later changed it in: 

https://github.com/mattermost/mattermost-plugin-calls/blob/0796721cf5ccb44232b3f1d02ec3a6bf0b00bace/webapp/src/components/call_widget/component.tsx#L1419-L1421 

since it showed the menu bar otherwise which wasn't ideal.

If we tackle this on the desktop side we may be able to retain the frame (with close button) but it's more complex than a simple flag there especially since we have plans to remove this new window entirely and converge everything on the widget window.

cpoile:
@streamer45 Hm, I think I prefer our in-windows close button, over the windowing system's decoration (too much variation leaving it to the system UI). Plus that's more complicated, and like you said we'll eventually move to the widget-only approach. So, proceed with this PR?

streamer45:
> @streamer45 Hm, I think I prefer our in-windows close button, over the windowing system's decoration (too much variation leaving it to the system UI). Plus that's more complicated, and like you said we'll eventually move to the widget-only approach. So, proceed with this PR?

Absolutely, I was merely offering some context on why we got here :grinning:

streamer45 (approved):

PR #280, opened 2022-12-14
[MM-49086] Implement authentication check and re-registration logic in case of failure

streamer45:
#### Summary

PR adds some logic to:

- Login to the job service on plugin activation. This makes sure that everything is configured correctly past the initial registration.
- Attempt to re-register if the first login attempt fails. This is particularly important to prevent getting locked out and requiring resetting some state in the k/v store.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-49086

### DIFF
--- a/server/job_service.go
+++ b/server/job_service.go
@@ -169,6 +169,39 @@ func (p *Plugin) newJobService(serviceURL string) (*jobService, error) {
 		return nil, fmt.Errorf("failed to create client: %w", err)
 	}
 
+	err = client.Login(cfg.ClientID, cfg.AuthKey)
+	if err == nil {
+		return &jobService{
+			ctx:    p,
+			client: client,
+		}, nil
+	}
+
+	// If login fails we attempt to re-register once as the jobs instance may
+	// have restarted, potentially losing stored credentials.
+	p.LogError("failed to login to job service", "err", err.Error())
+	p.LogDebug("attempting to re-register the job service client")
+
+	mutex, err := cluster.NewMutex(p.API, "job_service_registration")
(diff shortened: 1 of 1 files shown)


server/job_service.go
```diff
@@ -169,6 +169,39 @@ func (p *Plugin) newJobService(serviceURL string) (*jobService, error) {
...
+			client: client,
+		}, nil
+	}
+
+	// If login fails we attempt to re-register once as the jobs instance may
+	// have restarted, potentially losing stored credentials.
+	p.LogError("failed to login to job service", "err", err.Error())
+	p.LogDebug("attempting to re-register the job service client")
+
+	mutex, err := cluster.NewMutex(p.API, "job_service_registration")
```
cpoile, line 185: Just curious why we use a cluster mutex here, and not above?

cpoile (approved):
Approving assuming you know what you're doing, but I have one question. :)

server/job_service.go
streamer45: It's used above as well, just buried in `getJobServiceClientConfig`

cpoile: Ah. Would it be too much to consolidate both of these, so one of them doesn't get accidentally removed or whatnot? Up to you.

streamer45: Maybe we could call it just once at the beginning. The down side is that all other instances will be blocked on this for potentially some time as there are requests being made but I guess during init that would be acceptable. I'll take another look.

streamer45: Thanks @cpoile , looks cleaner now and works as expected :+1:

PR #281, opened 2022-12-14
MM-48076 MM-49085: Global expanded view fixes

cpoile:
#### Summary
- Hopefully fixed the linux problem
- Fixed a weird crashing bug for the emoji picker. Still not sure what's going on there.
- Removed the clicking-everywhere-drags issue I was seeing on 5.3
- Added a drag area to the top bar so that the global window can be moved. Decided to not add a cursor crosshairs, because I couldn't get the cursor to consistently show them in the entire area (they would disappear near the top of the window). Not sure why. This is good enough for now and we can revisit if design wants them added in the future.
- I've tested it on 5.3, 5.2, and webapp, but if you can please do the same for linux.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-49076
- https://mattermost.atlassian.net/browse/MM-49085

### DIFF
--- a/webapp/src/components/expanded_view/reaction_button.tsx
+++ b/webapp/src/components/expanded_view/reaction_button.tsx
@@ -66,9 +66,13 @@ export const ReactionButton = forwardRef(({trackEvent}: Props, ref) => {
     const addReactionText = showBar ? 'Close Reactions' : 'Add Reaction';
 
     const handleUserPicksEmoji = (ev: EmojiPickEvent) => {
+        let skin;
+        if (ev.skin) {
+            skin = EMOJI_SKINTONE_MAP.get(ev.skin);
+        }
         const emojiData = {
             name: ev.id,
-            skin: ev.skin ? EMOJI_SKINTONE_MAP.get(ev.skin) : undefined, // eslint-disable-line no-undefined
+            skin,
             unified: ev.unified.toUpperCase(),
         };
         callsClient.sendUserReaction(emojiData);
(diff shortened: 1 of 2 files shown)


webapp/src/components/expanded_view/reaction_button.tsx
```diff
@@ -66,9 +66,13 @@ export const ReactionButton = forwardRef(({trackEvent}: Props, ref) => {
...
 
     const handleUserPicksEmoji = (ev: EmojiPickEvent) => {
+        let skin;
+        if (ev.skin) {
+            skin = EMOJI_SKINTONE_MAP.get(ev.skin);
+        }
         const emojiData = {
             name: ev.id,
-            skin: ev.skin ? EMOJI_SKINTONE_MAP.get(ev.skin) : undefined, // eslint-disable-line no-undefined
+            skin,
```
streamer45, line 75: Out of curiosity, how is this semantically different than the previous version?

cpoile: @streamer45 It's not; it's technically exactly the same. That's why it's so weird. But I traced the crash to that. It only happened in the global popout (5.3), when you click the emoji icon and the emoji-mart tries to open. It must... I don't know, take the   `handleUserPicksEmoji` that's passed in, and what, run it? It's one of the stranger bugs I've seen, and this fixes it. I still want to know what's going on...

streamer45: Very weird, where you able to reproduce consistently or just randomly?

cpoile: At the time I could reproduce 100% consistently. Now, I can't reproduce at all. 😵‍💫

streamer45 (approved):

webapp/src/components/expanded_view/reaction_button.tsx
cpoile: I've tried in 5.2 and 5.3 multiple times, but it looks like whatever the issue was isn't happening anymore.

PR #282, opened 2022-12-15
[MM-49082] Bring back title bar for pop out window on Desktop (global)

streamer45:
#### Summary

PR mostly reverts a few definitions that allowed for the pop out window to be dragged around and closed on Desktop >= 5.3 (global widget).

#### Related PR

https://github.com/mattermost/desktop/pull/2475

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-49082


standalone/src/init.ts
```diff
@@ -106,12 +106,16 @@ function connectCall(channelID: string, callTitle: string, wsURL: string, iceCon
...
+            if (closeCb) {
+                closeCb();
+            }
         });
     } catch (err) {
-        delete window.callsClient;
         logErr(err);
+        if (closeCb) {
+            closeCb();
+        }
```
streamer45, line 118: Unrelated to the PR but fixing a bug since we weren't closing the widget window in case of initial failure (e.g. insecure context) which would make the widget disappear but the window would still be active (fully transparent) and stuck there until app reload.

cpoile (approved):
We could probably revert all of that portion of #281, maybe I should have just done that, sorry for the extra work.  If you want I can do that (there are some other portions that I added, which we can now remove). But this PR still works, even if we don't revert everything.

streamer45:
> We could probably revert all of that portion of #281, maybe I should have just done that, sorry for the extra work. If you want I can do that (there are some other portions that I added, which we can now remove). But this PR still works, even if we don't revert everything.

It's okay. We'll have chance to review more of this component soon I am sure.

PR #283, opened 2022-12-16
[MM-47006] Implement basic call quality degradation logic

streamer45:
#### Summary

PR implements some logic to detect degradation of call quality and inform users through a warning banner (see screenshots below).

The calculation are based on the concept of [Mean opinion score](https://en.wikipedia.org/wiki/Mean_opinion_score). For details on the magic behind it you can find lots of material on Google. I found https://netbeez.net/blog/impact-of-packet-loss-jitter-and-latency-on-voip/ to be a decent reference.

Please consider this a draft but ready for a first pass of reviews. I am still not entirely happy with a couple of details so not going to merge this anytime soon, specifically:

1. Doesn't work on Firefox. Several WebRTC statistics are lacking and the one we need is not getting generated as of version 96 (see https://github.com/pion/interceptor/issues/165 for details). I am investigating this but so far it looks like a bug on Mozilla's side.
2. The way we are handling voice tracks means that upon muting the client stops sending packets. This will result in huge jitter if a later unmute were to happen potentially skewing the logic and suddenly showing the banner. The easiest solution to this problem is to allow the empty (silence) audio packets to flow even when muted. This should also help getting better stats since network is being utilized but it comes at an added cost on bandwidth.
3. I am currently using 4 as the MOS score threshold to show the warning. This will likely need some proper tuning and lots of testing as we don't want it to become an annoyance if perceived quality is not that bad.

#### Screenshots

image

image

#### Related PR

https://github.com/mattermost/rtcd/pull/81

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-47006


cpoile (approved):
> The way we are handling voice tracks means that upon muting the client stops sending packets. This will result in huge jitter if a later unmute were to happen potentially skewing the logic and suddenly showing the banner. The easiest solution to this problem is to allow the empty (silence) audio packets to flow even when muted. This should also help getting better stats since network is being utilized but it comes at an added cost on bandwidth.

Would it be possible to generate "empty packets" (stats-wise) for the tracks that are set to mute? I'm not such a fan of sending around dozens of silent streams...

> I am currently using 4 as the MOS score threshold to show the warning. This will likely need some proper tuning and lots of testing as we don't want it to become an annoyance if perceived quality is not that bad.

Judging by the writeup, maybe 3 `Fair` or 2 `Poor` would be better?

streamer45:
> Would it be possible to generate "empty packets" (stats-wise) for the tracks that are set to mute? I'm not such a fan of sending around dozens of silent streams...

We can't really control packets sent by the browser so I don't think sending empty packets is a viable option. But a client would only be sending at most one silent audio track at a time if previously unmuted. 

What we could do to limit the waste is turning on [DTX](https://www.rfc-editor.org/rfc/rfc6716#section-2.1.9) although that will slightly lower audio quality but it's an option I am considering. 

> Judging by the writeup, maybe 3 `Fair` or 2 `Poor` would be better?

I tested this quite a bit and below 4 you could start noticing some issues hence the choice.

But overall I think that depends on how complex we'd like this to be from a UX perspective (e.g. showing different messages/banners based on the level of degradation). My first approach was to not overcomplicate it and try to show something as soon as we spot some potential problems.

cpoile:
> We can't really control packets sent by the browser so I don't think sending empty packets is a viable option. But a client would only be sending at most one silent audio track at a time if previously unmuted.

Sorry, I don't mean sending an empty packet, rather: inserting the empty packet's info into the stats manually, so that the stats are fixed.

streamer45:
> > We can't really control packets sent by the browser so I don't think sending empty packets is a viable option. But a client would only be sending at most one silent audio track at a time if previously unmuted.
> 
> Sorry, I don't mean sending an empty packet, rather: inserting the empty packet's info into the stats manually, so that the stats are fixed.

Stats on the client side are a black box really so that wouldn't be possible. What we could do is tamper with the reports on the server to account for this but I probably like that even less as an option than sending silence frames.

streamer45:
@tanmay-des This is ready for UX pass, let me know what you think.

tanmay-des (approved):
Looks good to me. 

In the design system for the global banners we don't have a warning color yet (Which I think is pretty important to add) like the combination you have used.

streamer45:
Update: I added a couple of improvements but mostly waiting to implement the mobile side before merging everything. I'll ask again for dev review once ready.

streamer45:
Merged everything and exported the common parts to https://github.com/mattermost/calls-common . This should be good for a final pass.

webapp/src/client.ts
```diff
@@ -455,6 +475,14 @@ export default class CallsClient extends EventEmitter {
             }
         }
 
+        // NOTE: we purposely clear the monitor's stats cache upon unmuting
+        // in order to skip some calculations since upon muting we actually
+        // stop sending packets which would result in stats to be skewed as
+        // soon as we resume sending.
+        // This is not perfect but it avoids having to constantly send
+        // silence frames when muted.
```
cpoile, line 483: Awesome 👍

cpoile (approved):
Thanks!

PR #284, opened 2022-12-19
MM-48681 - Add emojis to the recording view

cpoile:
#### Summary
- This is the strangeness I was talking about -- the `<Provider>`'s store is not seen if the component lives in the `webapp/` src tree. So bizarre. But after that, things fell into place.
- Also converted the recording view to react hooks, as a trial run for the rest of the webapp (if it makes sense). I think it simplifies things a bit and removes some boilerplate. Let me know what you think. (There's still some work to be done to simplify/clean this up, but it's a first step.)

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-48681


standalone/src/recording/components/reaction_stream.tsx
```diff
@@ -0,0 +1,145 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
```
streamer45, line 1: Is the need to duplicate this here purely due to the issue with `<Provider>`?

webapp/src/reducers.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable max-lines */
```
streamer45, line 1: Off topic but I am thinking your IDE is applying linting that's different from the one configured in repo. It would be nice to align at some point since otherwise I'd be constantly breaking more stuff for you.

webapp/src/components/expanded_view/component.tsx
```diff
@@ -137,8 +137,6 @@ const StyledMediaFullscreenButton = styled(MediaFullscreenButton)`
 	background-color: transparent;
 `;
 
-const MaxParticipantsPerRow = 4;
```
streamer45, line 140: Where did this go?

standalone/src/init.ts
```diff
@@ -314,6 +308,16 @@ export default async function init(cfg: InitConfig) {
 
     await cfg.initCb(store, theme, channelID);
 
+    // Sending this in the initCb doesn't work
```
streamer45, line 311: Why? :grinning:

streamer45 (commented):
You really like functional components :) Jokes aside, this is great :1st_place_medal: I just left a couple of questions.

standalone/src/recording/components/reaction_stream.tsx
cpoile: yep, only because of `<Provider>`.

webapp/src/reducers.ts
cpoile: The repo settings are 
```
    "max-lines": [
      1,
      {
        "max": 450,
        "skipBlankLines": true,
        "skipComments": false
      }
    ],
```
So if we want to change that, we can, otherwise we're going to be getting that warning. 
But wrt to this one -- it's not breaking anything for me , I disabled it in this file because it was causing a lot of visual noise. :) But there are a ton of eslint warnings we're ignoring at the moment (like no `any`s). Maybe it's worth it to go through and clean up the warnings?

webapp/src/components/expanded_view/component.tsx
cpoile: It wasn't being used, so I removed it.

standalone/src/init.ts
cpoile: I knew you'd ask that. I don't know actually, I didn't look into the why because I was stuck on the other issues. But I can go back and investigate.

cpoile:
The move to a functional component was just because that was one of the solutions I tried to fix the `<Provider>` issue, and I figured I may as well just clean it up a bit and keep it. I tried lots of things, even creating our own store, passing in the store explicitly, wrapping Providers everywhere in the hierarchy, crazy stuff that I was almost 100% sure wouldn't make a difference. :)

webapp/src/components/expanded_view/component.tsx
streamer45: That must be a mistake or a previous merge issue because I remember adding it to both components. I guess it doesn't hurt to bring it back here:

https://github.com/mattermost/mattermost-plugin-calls/blob/f8d4c0918f92fe72a34bad5b501e165c80f730f5/webapp/src/components/expanded_view/component.tsx#L816

webapp/src/reducers.ts
streamer45: I just think we should be consistent, not very opinionated about the details. If you are using different rules locally we should port them over to the repository so I can follow them as well. This is separate work of course, just bringing it up here since I've seen it before. A cleanup ticket would be nice.

cpoile: >  If you are using different rules locally we should port them over to the repository so I can follow them as well. 

I'm not using different rules locally, the repo settings I pasted are in the `.eslintrc.json` file in the repo.

streamer45: :thinking: Then we must be running a different version or something because neither the CI nor my local env are reporting an issue with that file.

streamer45: I guess it's the warning vs error maybe the real problem here?

cpoile: Yep, it's because it's a level 1 (warning).

streamer45: Okay, that makes sense now. Then let's plan to:

1. Review existing warnings, deciding if anything should be an error instead or not even a warning.
2. Fix remaining warnings and errors.

standalone/src/init.ts
cpoile: Ok, looked at it -- my mistake, the comment was wrong. It's a refactor, and this way the recording view gets set up correctly. Pre-PR, the recording view wasn't registering itself as a participant (in the store), so the logic for websockets and displaying reactions was breaking (some of the selectors use the current voice channel state, which was null).
I fixed the comment.

standalone/src/recording/components/reaction_stream.tsx
streamer45: Out of curiosity, could you try adding this:

```patch
diff --git a/standalone/webpack.config.js b/standalone/webpack.config.js
index bb603c7..ff181f7 100644
--- a/standalone/webpack.config.js
+++ b/standalone/webpack.config.js
@@ -91,6 +91,7 @@ module.exports = {
             'react-dom': path.resolve(__dirname, './node_modules/react-dom'),
             'react-bootstrap': path.resolve(__dirname, './node_modules/react-bootstrap'),
             bootstrap: path.resolve(__dirname, './node_modules/bootstrap'),
+            'react-redux': path.resolve(__dirname, './node_modules/react-redux'),
         },
         fallback: {
             src: path.resolve(__dirname, '../webapp/src'),
```

And import from the webapp side instead.

cpoile: Ok, that works... But why? What does resolving react-redux have to do with importing from a different directory?

streamer45: No idea, I just found https://github.com/reduxjs/react-redux/issues/1489 and thought it was worth a try.

standalone/src/init.ts
cpoile: And, now I've reverted that change. Because for some reason it breaks the widget.

webapp/src/reducers.ts
cpoile: https://mattermost.atlassian.net/browse/MM-49235

standalone/src/recording/components/reaction_stream.tsx
cpoile: What a find, thank you. That clarifies (at least a little) what was going wrong.

streamer45 (approved):
Looks great :raised_hands:

PR #286, opened 2022-12-20
[MM-49267] Update recording job runner asynchronously on activation

streamer45:
#### Summary

PR adds logic to asynchronously update the recording job runner on activation rather than waiting for the first recording job to trigger.

Also reworked the `startCallRecording` action so that it correctly surface to the caller any error during the request (e.g. job service not initialized).

#### Related PR

https://github.com/mattermost/calls-offloader/pull/11

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-49267


cpoile (approved):
Good idea to return and display the error. 👍

PR #291, opened 2023-01-10
[MM-49589] CircleCI -> Github Actions

streamer45:
#### Summary

PR converts CircleCI pipeline into Github Actions using the provided plugin actions.

A couple of things left to consider:

1. We should disable CircleCI on this repository to avoid getting it to error out.
2. I am yet to understand how this will impact the release pipeline. I'd like to test it but also don't want to mess things and release random stuff. @phoinixgrr Added you to review so you can chime in and provide some feedback.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-49589


Makefile
```diff
@@ -160,16 +160,6 @@ else
...
-## Builds the webapp on ci -- dependencies are handled by the npm-dependencies step in ci
-.PHONY: webapp-ci
-webapp-ci:
-	cd webapp && $(NPM) run build
-
-## Builds the standalone apps on ci -- dependencies are handled by the npm-dependencies step in ci
-.PHONY: standalone-ci
-standalone-ci:
-	cd standalone && $(NPM) run build
-
```
cpoile, line 172: Was this intentional? This means we'll be rerunning the webapp/node_modules for ci

streamer45: I assumed that would be cached and wouldn't really matter. It was actually failing as it was probably because the way we do caching is slightly changed now.

cpoile: Hm, I guess I'm more familiar with the circleci way (where we would pull in hashed caches). But if this works, then I guess the extra work isn't too bad (?).

cpoile (approved):

.github/workflows/cd.yml
```diff
@@ -0,0 +1,20 @@
...
+    types:
+      - completed
+  push:
+    tags:
+      - "v[0-9]+.[0-9]+.[0-9]+"
+
+permissions:
+  contents: read
+
+env:
```
spirosoik, line 14: This is not needed actually, it is always true in github actions. check for reference https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables

.github/workflows/ci.yml
```diff
@@ -0,0 +1,19 @@
...
+    branches:
+      - main
+    tags:
+      - "v[0-9]+.[0-9]+.[0-9]+"
+  pull_request:
+
+permissions:
+  contents: read
+
+env:
```
spirosoik, line 13: same here

spirosoik (commented):
one minor change

.github/workflows/cd.yml
streamer45: Thanks, I wasn't 100% sure so I wanted to be on the safe side but looks like it can be removed indeed. Will update.

streamer45:
> @streamer45 Spiros mentioned in channel that this won't affect Matterbuild's plugin cutting process.

Cool, then this should be ready for a spin, pending review and some admin to disable CircleCI.

toninis:
Hey @streamer45 can we merge this ? Isn't it ready ?

streamer45:
> Hey @streamer45 can we merge this ? Isn't it ready ?

@toninis This is waiting from an approval from someone on your team as far as I can tell. There is an outstanding question on how this change will affect the release process but I believe that shouldn't be impacted.

toninis:
It should not change it . We 've already used the same logic for the rest plugins

toninis:
cc: @mattermost/core-build-engineers

streamer45:
Thanks @toninis, would you be able to help disabling CircleCI once we merge to avoid the failures?

streamer45:
Going to merge this. If you have concerns please ping the team :)

PR #292, opened 2023-01-10
[MM-48900] Display available recording files in call thread root post

streamer45:
#### Summary

I found a couple of issues while implementing this:

1. Multiple recording files were not saved correctly in post props due to a type mismatch.
2. There was a potential race condition in a rapid sequence of stop recording -> start recording that would lead to an invalid state with the host and users thinking a seeing the new recording as ended while ongoing.
3. Fixed a potential state parsing bug due to improper key length check.

As far as UX changes so far I only implemented a simple indicator of available recordings, no extra behaviour as there are some inconsistencies with the design so anything more will require further input.

#### Screenshot

image

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-48900

### DIFF
--- a/server/session.go
+++ b/server/session.go
@@ -124,6 +124,7 @@ func (p *Plugin) addUserSession(userID, connID string, channel *model.Channel) (
 		if userID == botID {
 			if state.Call.Recording != nil && state.Call.Recording.StartAt == 0 {
 				state.Call.Recording.StartAt = time.Now().UnixMilli()
+				state.Call.Recording.BotConnID = connID
 			} else if state.Call.Recording == nil || state.Call.Recording.StartAt > 0 {
 				// In this case we should fail to prevent the bot from recording
 				// without consent.
@@ -185,8 +186,6 @@ func (p *Plugin) removeUserSession(userID, connID, channelID string) (channelSta
 	var prevState channelState
 	errNotFound := errors.New("not found")
 
-	botID := p.getBotID()
-
 	setChannelState := func(state *channelState) (*channelState, error) {
 		if state == nil {
 			return nil, fmt.Errorf("channel state is missing from store")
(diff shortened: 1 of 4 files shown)


server/session.go
```diff
@@ -124,6 +124,7 @@ func (p *Plugin) addUserSession(userID, connID string, channel *model.Channel) (
 		if userID == botID {
 			if state.Call.Recording != nil && state.Call.Recording.StartAt == 0 {
 				state.Call.Recording.StartAt = time.Now().UnixMilli()
+				state.Call.Recording.BotConnID = connID
```
cpoile, line 127: I like the change, but I'm not sure I totally understand what the race condition was.

cpoile (approved):
Nice! Just one question for my own understanding.

server/session.go
streamer45: @cpoile Apologies, let me try to explain it properly with a triggering sequence:

With a recording already ongoing:

1. Host stops recording
2. Recording state is saved as ended and websocket event sent out.
3. Client receives ws event and updates local state, recording has ended.
4. Host quickly starts a new recording before recording bot from previous recording leaves call.
6. Recording bot for previous recording job eventually leaves the call, improperly setting the current recording as ended.

cpoile: Were you able to reproduce this? If so, does the second recording start? (considering we're using the same channel state)

streamer45: Yes it's pretty easy to reproduce. Recording starts but the client gets the stop event right away and then the state is tainted so both client and server think there's no recording on-going while the job is actually running and recording until the call ends.

cpoile: Cool :)

streamer45:
@tanmay-des Heads up that we are going to merge this soon mostly because we have some important fixes in it and would like to prepare a dot release. Anything outstanding UX wise will be taken in a subsequent PR.

tanmay-des (approved):
Sounds good!

PR #293, opened 2023-01-11
[MM-49004] Allow for listening address to be specified in configuration

streamer45:
#### Summary

PR adds a new `UDPServerAddress` setting to potentially configure the address/interface to be used by the RTC server (receiving/sending UDP packets).

In addition we customize those settings that don't make sense when RTCD is used so that they don't get displayed and cause confusion.

#### Related PR

https://github.com/mattermost/rtcd/pull/85

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-49004


cpoile (approved):
👍

PR #294, opened 2023-01-11
MM-49613 - Emoji picker crash

cpoile:
#### Summary
- It appears that, because of all the fancy features (shadow-doming? custom elements? IntersectionObserver?),  emoji-mart does not work if you try to instantiate it twice on the same page (see: https://github.com/missive/emoji-mart/issues/755). Somehow we were running into that. Or maybe into something completely different, but with the same `illegal-constructor` effect.
- Because I couldn't figure it out, propose to swap it out with `emoji-picker-react`. It has a few fewer customizations, and seems to lag a bit on load (I'm seeing a few failed fetches to a cdn for apple emojis). But otherwise it works.
- We lose the "on click outside" closing behavior, too. 

image

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-49613

### DIFF
--- a/webapp/package-lock.json
+++ b/webapp/package-lock.json
@@ -5,8 +5,6 @@
   "packages": {
     "": {
       "dependencies": {
-        "@emoji-mart/data": "1.0.6",
-        "@emoji-mart/react": "1.0.1",
         "@mattermost/compass-icons": "0.1.32",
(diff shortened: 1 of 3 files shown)


streamer45 (approved):
Thanks :100:

PR #295, opened 2023-01-12
[MM-49588] Fix audio devices list overflow

streamer45:
#### Summary

PR fixes especially long audio devices lists from overflowing outside the viewport. To achieve this we are doing two things:

1. Setting a `max-height` to be the height of widget + height of menu.
2. Making the submenus always start from the top.

Originally I tried to make them grow upwards but that would break the global widget logic and required even more complex calculations. I think this approach is simple enough since it doesn't make sense to grow indefinitely anyway.

Since I was touching the code I opted to add a background to help the user identify what list they currently have open.

#### Screenshots

Before

image

image

After

image

image

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-49588

### DIFF
--- a/webapp/src/components/call_widget/component.scss
+++ b/webapp/src/components/call_widget/component.scss
@@ -135,6 +135,8 @@ a.calls-channel-link {
 
 #calls-widget .Menu {
   position: static;
+  border: none;
+  box-shadow: none;
 }
 
 #calls-widget button {
--- a/webapp/src/components/call_widget/component.tsx
+++ b/webapp/src/components/call_widget/component.tsx
@@ -235,7 +235,9 @@ export default class CallWidget extends React.PureComponent<Props, State> {
             },
             audioInputsOutputsMenu: {
                 left: 'calc(100% + 4px)',
-                top: 'auto',
+                overflow: 'auto',
+                top: 0,
+                maxHeight: 'calc(100% + 90px)',
             },
             expandButton: {
                 position: 'absolute',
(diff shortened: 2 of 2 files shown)


webapp/src/components/call_widget/component.scss
```diff
@@ -135,6 +135,8 @@ a.calls-channel-link {
 
 #calls-widget .Menu {
   position: static;
+  border: none;
+  box-shadow: none;
```
streamer45, line 139: Slightly unrelated but this was needed to fix some more issues when switching to Boards.

webapp/src/components/call_widget/component.tsx
```diff
@@ -905,13 +907,13 @@ export default class CallWidget extends React.PureComponent<Props, State> {
...
-                if (entries.length === 0) {
+                if (entries.length === 0 || entries[0].borderBoxSize.length === 0) {
                     return;
                 }
                 sendDesktopEvent('calls-widget-resize', {
                     element: 'calls-widget-audio-menu',
-                    width: Math.round(entries[0].contentRect.width),
-                    height: Math.round(entries[0].contentRect.height),
+                    height: Math.round(entries[0].borderBoxSize[0].blockSize),
+                    width: Math.round(entries[0].borderBoxSize[0].inlineSize),
```
streamer45, line 916: Changes here are needed to make sure the width passed to the desktop side accounts for the scrollbar if present.

tanmay-des (approved):
Looks good!

PR #298, opened 2023-01-12
[MM-46887] Improve rtcd client auth logic if URL changes

streamer45:
#### Summary

PR addresses a potential corner case that would lead to getting a registered client unable to login. If the URL to rtcd changed but the target instance was kept the same (e.g. only the hostname changed), the client would attempt to register again and get a conflict error, resulting in the plugin permanently failing to start.

We slightly alter this behaviour to allow the client to authenticate first and only attempt registration if that fails.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-46887


cpoile (approved):
Makes sense. 👍

PR #304, opened 2023-01-20
MM-49235 - Fix eslint warnings, better TS typing

cpoile:
#### Summary
- Mostly annoying little things, but some good TS typings to help us in the future.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-49235


standalone/src/init.ts
```diff
@@ -254,49 +265,51 @@ export default async function init(cfg: InitConfig) {
     connectCall(channelID, callTitle, getWSConnectionURL(getConfig(store.getState())), iceConfigs, (ev) => {
         switch (ev.event) {
         case 'hello':
-            setServerVersion(ev.data.server_version)(store.dispatch, store.getState);
+            // TODO: where is this server_version coming from? I think we can remove this case.
+            // @ts-ignore
+            store.dispatch(setServerVersion(ev.data.server_version));
```
cpoile, line 270: This is an open question. Would this event even reach here? I don't think so...?

streamer45: The initial `hello` websocket response [contains](https://github.com/mattermost/mattermost-server/blob/3c1f63d41b7ac57699ea7945db17f22fb19f7c8d/app/platform/web_conn.go#L664) the server version which as far as I remember was necessary to be in store to make things function properly. Mostly mimicking what happens in [webapp](https://github.com/mattermost/mattermost-webapp/blob/e66cfe689f501a919af5be1dc16f0a01d3da6fd7/actions/websocket_actions.jsx#L1244-L1247).

streamer45: As to why we get there, that would be through https://github.com/mattermost/mattermost-plugin-calls/blob/e2d24282ec8a3ed0292abc2328e6d742bdf00232/webapp/src/websocket.ts#L116

cpoile: That's funny, because I logged it and I never reached it... Shouldn't it fire on every new call?

streamer45: I'd expect it to since we are creating a brand new connection, yes.

cpoile: Just double checked -- it doesn't seem to be receiving it.

streamer45: Honestly I cannot reproduce this, the event handler is always registered before connection happens and the `hello` event emitted correctly. 

Could you verify whether in your case the event handler is registered before or not?

standalone/src/init.ts
```diff
@@ -174,11 +189,11 @@ async function fetchChannelData(store: Store, channelID: string) {
                 },
             });
 
-            const userStates = {} as any;
+            const userStates: Record<string, UserState> = {};
             const users = resp.call.users || [];
             const states = resp.call.states || [];
             for (let i = 0; i < users.length; i++) {
-                userStates[users[i]] = states[i];
+                userStates[users[i]] = {...states[i], id: users[i], voice: false};
```
cpoile, line 196: Added some fields here. Wanted to point it out.

streamer45 (approved):
Awesome, let's merge this before it causes more pain :)

cpoile:
@streamer45 Thanks!

PR #305, opened 2023-01-20
Fix e2e snapshots

streamer45:
#### Summary

I think (hope) we simply went back and forth between boards product and plugin on the CI and that led to snapshots differing a few times. With that said, if these fail again anytime soon I'll simply remove them since it's getting a bit ridiculous :)


cpoile (approved):
Agreed. :)

PR #310, opened 2023-01-26
[MM-49998] Add jitter on reconnections to rtcd

streamer45:
#### Summary

Adding some jitter on (re)connection in an attempt to avoid bringing `rtcd` instance to a halt.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-49998


cpoile (approved):
This is probably what will fix it, I bet.

PR #311, opened 2023-01-27
[MM-49766] Review rtcd connection logic

streamer45:
#### Summary

PR slightly refactors the `rtcd` client connection logic aimed at limiting the number of authentication requests (either due to registrations or connections) and at the same type simplifying the logic where possible.

Also reviewed a few log statements and _downgraded_ some errors.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-49766

### DIFF
--- a/server/rtcd.go
+++ b/server/rtcd.go
@@ -6,6 +6,7 @@ package main
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"math/rand"
 	"net"
@@ -33,6 +34,8 @@ const (
 	hostCheckInterval       = 10 * time.Second
 	baseReconnectIntervalMs = 5000
 )
+
+var errClientReplaced = errors.New("client replaced")
 
 type rtcdHost struct {
 	ip           string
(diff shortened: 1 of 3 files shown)


server/rtcd.go
```diff
@@ -348,77 +351,55 @@ func (m *rtcdClientManager) newRTCDClient(rtcdURL, host string, dialFn rtcd.Dial
...
-		if err := m.registerRTCDClient(clientCfg, dialFn); err != nil {
-			return fmt.Errorf("failed to register rtcd client: %w", err)
-		}
-
-		return nil
-	}
-
-	reconnectCb := func(c *rtcd.Client, attempt int) error {
+	var reconnectCb rtcd.ClientReconnectCb
+	reconnectCb = func(c *rtcd.Client, attempt int) error {
```
streamer45, line 360: With this PR we are essentially treating this as a simple disconnected callback and dealing with the actual reconnection on the plugin side. This simplifies things as we have more context to operate on but the naming here is very misleading so I am potentially looking to change this or remove it entirely from the `rtcd` side to avoid further confusion.

cpoile: I'm not sure what you mean by `the naming here is very misleading` -- at first I was going to suggest renaming it to `disconnectCb`, but looking at the code, this is the reconnect handler... so what's the misleading part?

cpoile (approved):

server/rtcd.go
streamer45: Mostly the fact that the reconnection would happen after the callback if we returned a nil error. But the way we are using it now is purely to detect disconnection since we create a brand new client and return a fake error just to exit the previous handler.

PR #312, opened 2023-01-27
Fix websocket handlers

streamer45:
#### Summary

We need to check for in payload channelID  when receiving an event since the calls bot may not be in a channel so `broadcast.channel_id` won't be useful.


cpoile (approved):
😵‍💫 Sorry.
😅 

Tested on arm64 recorder, works. Thank you!

PR #316, opened 2023-02-02
[MM-50288] Simulcast

streamer45:
#### Summary

PR adds simulcast support for screen sharing. Clients will now try to send  two encodings when screen sharing:

- Low quality: 500Kb/s
- High quality: 2500Kb/s

This allows us to increase the overall quality from the previously hard-coded 1Mbit/s and at the same time allow for low bandwidth clients to receive a feed without triggering a full blown congestion.

As expected most changes are on the backend. On this side we simply instruct the peer to send the expected encodings.
A couple of additional changes included here are:

- Simplified logic on rendering the screen player which was causing a lot of flakyness.
- Updated the load-test client to actually send two different encodings for screen (`simulcast` flag).

As you can notice from the code I originally planned to send 3 encodings to also account for a medium quality track. However this resulted in much more pressure on the sender side (CPU and bandwidth) without that much of a benefit as it also made the backend logic trickier.

As noted in a comment this is not working in Firefox due to a bug, in which case we fallback to sending a single encoding.

#### Related PR

https://github.com/mattermost/rtcd/pull/90

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-50288


webapp/src/components/call_widget/component.tsx
```diff
@@ -256,10 +256,17 @@ export default class CallWidget extends React.PureComponent<Props, State> {
...
         this.node = React.createRef();
         this.menuNode = React.createRef();
-        this.screenPlayer = React.createRef();
+    }
+
+    setScreenPlayerRef = (node: HTMLVideoElement) => {
+        if (node && this.state.screenStream) {
+            node.srcObject = this.state.screenStream;
+        }
+        this.screenPlayer = node;
```
cpoile, line 269: Interesting, this is nice. Cleaner.

cpoile (approved):
Very nice!

streamer45:
Note to self: should add a config option and make this opt-in since it's still in a semi-experimental state.

cpoile (approved):
Thanks for adding the config!

PR #317, opened 2023-02-02
[MM-49944] Dispatch missing call start event

streamer45:
#### Summary

We were failing to dispatch the call start event when fetching call (and channel) data. This bug was easily reproducible by logging out and back in with an ongoing call. I cannot guarantee that's what happened to all the reporting users but these changes should likely fix most instances of the issue.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-49944

### DIFF
--- a/webapp/src/index.tsx
+++ b/webapp/src/index.tsx
@@ -590,65 +590,83 @@ export default class Plugin {
                     type: RECEIVED_CHANNEL_STATE,
                     data: {id: channelID, enabled: resp.data.enabled},
                 });
+
+                const call = resp.data.call;
+                if (!call) {
+                    return;
+                }
+
+                store.dispatch({
+                    type: VOICE_CHANNEL_CALL_START,
+                    data: {
+                        channelID,
+                        startAt: call.start_at,
+                        ownerID: call.owner_id,
+                        hostID: call.host_id,
+                    },
+                });
+
                 store.dispatch({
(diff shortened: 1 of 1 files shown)


webapp/src/index.tsx
```diff
@@ -590,65 +590,75 @@ export default class Plugin {
...
                 });
+
+                const call = resp.data.call;
+                if (!call) {
+                    return;
+                }
+
+                store.dispatch({
+                    type: VOICE_CHANNEL_CALL_START,
+                    data: {
```
cpoile, line 601: This is interesting... In mobile we rely on the websocket start event and keep all calls state up to date that way (with an overall "give me all the calls states" on connect/reconnect).  I'm not sure which is better, but it's certainly fewer api calls. Maybe a thought for the future if we want to move towards that?

cpoile (approved):
Nice debugging!

webapp/src/index.tsx
streamer45: Overall I agree. We should rely more on websocket as it's the only way to avoid potential race conditions. That said there are a few things to consider:

- The ws start event won't be much useful if users never receive it because the call started before they connected to MM (this is the case this PR is addressing).
- This long lasting TODO should be deal with: https://github.com/mattermost/mattermost-plugin-calls/blob/2ab4b0eaf070ed7350fc4f614b446a4678237bd8/server/websocket.go#L508

PR #326, opened 2023-02-08
[MM-42233] Fix widget crash on joining calls

streamer45:
#### Summary

PR fixes a long lasting bug causing a failure when joining a call. After some inspection I realized the problem wasn't the missing channel but the fact that the client doesn't know if it's connected due to not receiving the ws event for the current user. This could happen when someone tries to connect to a call during a ws reconnection attempt (e.g. just coming back from a laptop sleep). The end result would be very confusing, with the widget not showing while actually being in a call (e.g. hearing everyone but no UI).

The simplest (although not perfect) fix here is to use the channelID from the calls client as source of truth and trust that we are going to connect, eventually.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-42233

### DIFF
--- a/webapp/src/client.ts
+++ b/webapp/src/client.ts
@@ -37,7 +37,7 @@ export default class CallsClient extends EventEmitter {
     private readonly onDeviceChange: () => void;
     private readonly onBeforeUnload: () => void;
     private closed = false;
-    private initTime = Date.now();
+    public initTime = Date.now();
 
     constructor(config: CallsClientConfig) {
         super();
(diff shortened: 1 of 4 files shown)


streamer45:
To note this has an important side effect which I still need to consider. The change will essentially make it so we render the widget almost instantly, even if not connected. I think this may be okay as it actually makes it fill a bit snappier but we should likely think of some UI to notify of a successful connection otherwise it could get confusing once more.

In the meantime I guess we could add some simple state variable and only display the widget when actually connected. I'll think a bit more about this and see what can be done.

cpoile (approved):
Wow, good debugging.

PR #328, opened 2023-02-08
[MM-50413] Fix addIceCandidate error due to potential race

streamer45:
#### Summary

PR fixes a potential issue when ice candidates are received before a remote description is set. The workaround is to queue them up until we set the description.

@cpoile I believe the mobile side is potentially affected as well but thought that we may wait to share `RTCPeer` so we can stop duplicating every little fix. But let me know if you think that's too far in the future and if so I will create a PR on the other side.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-50413

### DIFF
--- a/webapp/src/rtcpeer/index.ts
+++ b/webapp/src/rtcpeer/index.ts
@@ -1,6 +1,6 @@
 import {EventEmitter} from 'events';
 
-import {logDebug} from '../log';
+import {logDebug, logErr} from '../log';
 
 import {
     RTCPeerConfig,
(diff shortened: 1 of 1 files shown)


webapp/src/rtcpeer/index.ts
```diff
@@ -95,7 +96,17 @@ export default class RTCPeer extends EventEmitter {
 
         switch (msg.type) {
         case 'candidate':
-            await this.pc.addIceCandidate(msg.candidate);
```
streamer45, line 98: I don't think there's any need to make this synchronous so changed it.

cpoile (approved):
This probably affects mobile too. As for sharing RTCPeer, after a week of work I paused it -- it's not going to be easy, if it's possible at all. I'll catch you up in our next 1:1. 
Assign me a ticket if you want me to take care of it (easy enough).

streamer45:
> This probably affects mobile too. As for sharing RTCPeer, after a week of work I paused it -- it's not going to be easy, if it's possible at all. I'll catch you up in our next 1:1. Assign me a ticket if you want me to take care of it (easy enough).

@cpoile Thanks, https://mattermost.atlassian.net/browse/MM-50415

PR #329, opened 2023-02-09
[MM-50429] Surface slash commands errors through modal

streamer45:
#### Summary

As discussed in https://github.com/mattermost/mattermost-plugin-calls/pull/318 relying on the native error surfacing for slash commands was quite brittle and isn't working at all if such commands are sent in a reply box. So opting to move away from this and start using our generic modals (see screenshots below).

Happy to review the messages to something better once we work on translations.

/cc @itao @tanmay-des for visibility.

#### Screenshots

`/call start` while in a call already:

image

`/call join` while in a call already:

image

`/call leave` while not in a call:

image

`/call end` while not in a call:

image

`/call end` with no permissions:

image

`/call recording start` with no call on going:

image

`/call recording stop` with no call on going:

image

`/call recording stop` with no recording on going:

image

`/call recording start` with recording already on going:

image

`/call recording start` with no permissions:

image

`/call recording stop` with no permissions:

image

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-50429


webapp/src/index.tsx
```diff
@@ -250,7 +253,11 @@ export default class Plugin {
             case 'start':
                 if (subCmd === 'start') {
                     if (voiceConnectedUsersInChannel(store.getState(), args.channel_id).length > 0) {
-                        return {error: {message: 'A call is already ongoing in the channel.'}};
+                        store.dispatch(displayGenericErrorModal(
+                            defineMessage({defaultMessage: 'Unable to start call'}),
+                            defineMessage({defaultMessage: 'A call is already ongoing in the channel.'}),
```
cpoile, line 258: i18n? nice!

cpoile (approved):
Awesome, let's get it in.

webapp/src/index.tsx
streamer45: I can still surprise you :grinning:

PR #331, opened 2023-02-09
MM-42564 - Audio Output for Airpods shows two outputs

cpoile:
#### Summary
- Chrome is where this problem can happen: it returns default devices as a separate device. I don't think this is a problem, and is probably helpful (so the user knows that that particular device is their usual default). So, instead of deduping (they will have the same `groupId`), let's ensure that the default is labelled as default. In the ticket, this didn't happen, and that's where the confusion came from. But if it was labelled "Default - Airpods" and "Airpods", that would make sense. I think. What do you think?
  - Aside: while testing today, it always had the "Default - " prefix. Maybe that's a recent Chrome change, or I did something to cause the screenshot seen in the ticket. Either way, this will prevent that from happening again.
- Safari won't ever have this trouble, because it's unable to show outputs.
- Firefox requires a flag to be set to see outputs. Even If it's set, Firefox shouldn't have this problem because it doesn't distinguish between default and non-default in the enumeration, so there won't be a double counting.

- Added notes for our documentation: https://community.mattermost.com/private-core/pl/ekwjh8dmu3rtbg9ajtpmd79ygh

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-42564

### DIFF
--- a/webapp/src/components/call_widget/component.tsx
+++ b/webapp/src/components/call_widget/component.tsx
@@ -1004,6 +1004,13 @@ export default class CallWidget extends React.PureComponent<Props, State> {
 
         const currentDevice = deviceType === 'input' ? this.state.currentAudioInputDevice : this.state.currentAudioOutputDevice;
 
+        const makeDeviceLabel = (device: MediaDeviceInfo) => {
+            if (device.deviceId.startsWith('default') && !device.label.startsWith('Default')) {
+                return `Default - ${device.label}`;
+            }
+            return device.label;
+        };
+
         const deviceList = devices.map((device) => {
             return (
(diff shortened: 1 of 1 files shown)


streamer45 (approved):
Thanks. I think this fine, with the only caveat that there's still a difference between system defaults (device labeled as default) and whatever device is saved as default by the client (saved in local storage). I don't think this should cause confusion but just wanted to mention it.

cpoile:
Yah, that's a good point. I don't think it should be an issue, since that "default" concept is more like a code-level default (not exposed to the user). I added a comment in the code to clarify. Thanks.

PR #332, opened 2023-02-13
Fix load-test client from sending PLIs for voice tracks

streamer45:
#### Summary

As part of improving the simulation I added some code to send a picture-loss-indication when receiving a track, similarly to what a real client (e.g. browser) would do. However I forgot to add a check on the track type so now we are sending them out for audio tracks as well which doesn't make sense and also causes an error spam on the server side.


cpoile (approved):
Or, this was load testing the error logging all along? :)

PR #333, opened 2023-02-14
Mitigate some glare in load-test client

streamer45:
#### Summary

`pion` doesn't yet support the updated API that allows for perfect negotiation so we have to resort to some hack to mitigate the issue.

To fix the `SetRemoteDescription failed: InvalidModificationError: invalid proposed signaling state transition: have-local-offer->SetRemote(offer)->have-remote-offer` error we detect the conflict and queue the offer for later on.


cpoile (approved):
👍

PR #339, opened 2023-02-17
[MM-49089] Propagate blocking errors to desktop

streamer45:
#### Summary

PR adds some logic to propagate client blocking errors (e.g. connection failures) to the desktop side since the global widget window gets closed immediately but we want to show a modal. I also took the opportunity to simplify a bit. I was originally storing the client error because I was unaware we could pass props to the modal but this is much cleaner.

#### Related PR

https://github.com/mattermost/desktop/pull/2557

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-49089


cpoile (approved):
Nice!

PR #342, opened 2023-02-17
MM-50542: Improve performance on connect/reconnect

cpoile:
#### Summary
- Batch the actions on connect/reconnect
- This requires a little bit of awaiting on the client, but I don't think this is an issue, and the overall gain is likely far higher.
- Just on my local dev server this batches 47 actions on reconnect. Probably a much better improvement on a server like community with more channels using calls.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-50542


streamer45 (approved):
This is great, can't wait to load it on Community :tada:

PR #343, opened 2023-02-21
[MM-50623] Allow to run Calls slash commands from other plugins/products

streamer45:
#### Summary

PR refactors the slash commands handler in order to let other plugins (Playbooks in today's case) to easily start/stop calls from the client side.

#### Related PR

https://github.com/mattermost/mattermost-plugin-playbooks/pull/1816

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-50623

### DIFF
--- a/webapp/src/slash_commands.tsx
+++ b/webapp/src/slash_commands.tsx
@@ -1,0 +1,222 @@
+import {defineMessage} from 'react-intl';
+
+import {CommandArgs} from '@mattermost/types/integrations';
+
+import {getChannel} from 'mattermost-redux/selectors/entities/channels';
+import {getCurrentUserId, getUser, isCurrentUserSystemAdmin} from 'mattermost-redux/selectors/entities/users';
+import {getChannel as getChannelAction} from 'mattermost-redux/actions/channels';
+
+import {
+    startCallRecording,
+    stopCallRecording,
+    displayGenericErrorModal,
+} from 'src/actions';
+import {SHOW_END_CALL_MODAL} from 'src/action_types';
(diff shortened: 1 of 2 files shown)


webapp/src/slash_commands.tsx
```diff
@@ -0,0 +1,222 @@
...
+        if (!connectedID) {
+            let title = '';
+            if (fields.length > 2) {
+                title = fields.slice(2).join(' ');
+            }
+
+            let channel = getChannel(store.getState(), args.channel_id);
+            if (!channel) {
+                await store.dispatch(getChannelAction(args.channel_id));
+                channel = getChannel(store.getState(), args.channel_id);
```
cpoile, line 67: This can be simplified I think:
```suggestion
                const {data: channelData} = await store.dispatch(getChannelAction(args.channel_id));
                channel = channelData
```

webapp/src/slash_commands.tsx
```diff
@@ -0,0 +1,222 @@
...
+            }
+
+            const team_id = args?.team_id || channel.team_id;
```
cpoile, line 70: Also we might want to do this check before, and then maybe we don't need to get the channel?

cpoile (approved):
Cool!

webapp/src/slash_commands.tsx
streamer45: I didn't know that, nice!

streamer45:
@cpoile Could you have another quick look?

cpoile (approved):
Looks great!

PR #346, opened 2023-02-22
[MM-50751] Fix race condition attaching media tracks on join

streamer45:
#### Summary

This was an interesting one given I could never reproduce the issue on my end, neither locally nor when connecting on remote instances.

The bug has been there for a long time as I remember first encountering this on a Community recording a few months ago. Something in the recorder setup made it more likely to reproduce. In fact, I was finally able to do this earlier today through a quick load-test on Community.

The problem was that we are registering the media track events handler(s) when mounting the component (both widget and recording view) but any delay in doing so could mean events firing before we get a chance to listen and tracks getting "lost", or better, never attached to the document, hence never played. To fix this, before listening on the track events we also check for existing available tracks and attach them accordingly.

Thanks to @hanzei for raising this issue. I was almost convinced it was a Chromium bug after the previous failed attempts at reproducing this but indeed, as often is the case, it was an issue on our end.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-50751

### DIFF
--- a/standalone/src/recording/components/recording_view/index.tsx
+++ b/standalone/src/recording/components/recording_view/index.tsx
@@ -40,25 +40,36 @@ const RecordingView = () => {
     }
     const callHostID = useSelector((state: GlobalState) => voiceChannelCallHostID(state, channelID)) || '';
 
-    useEffect(() => {
-        window.callsClient?.on('remoteScreenStream', (stream: MediaStream) => {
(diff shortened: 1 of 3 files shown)


hanzei:
Wow, I was not expecting a fix that soon. Thank you @streamer45 ! 🚀

cpoile (approved):
Nice!

PR #347, opened 2023-02-22
Media e2e

streamer45:
#### Summary

PR adds some basic e2e tests to cover media functionality, namely screen sharing and sending/receiving audio. This means we are now transmitting real media content and verifying that at least it gets received and rendered at the DOM level.

Another notable change added here is that I slightly refactored the tests to enable full parallelism which should hopefully make the pipeline slightly more efficient.

#### Screenshot

image


e2e/tests/media.spec.ts
```diff
@@ -0,0 +1,119 @@
...
+    test('share screen button', async ({page}) => {
+        const userPage = await startCall(userState.users[3].storageStatePath);
+
+        const devPage = new PlaywrightDevPage(page);
+        await devPage.joinCall();
+
+        await page.locator('#calls-widget-toggle-menu-button').click();
+        await page.locator('#calls-widget-menu-screenshare').click();
+
+        await expect(page.locator('#screen-player')).toBeVisible();
```
cpoile, line 28: For my own interest -- I'm not sure how this is working, shouldn't there be the `calls-screen-source-modal`, instead of the `screen-player`?

cpoile: Maybe I'm getting confused by the way the `page` is used create a `devPage`, which seems to be a way to add our own methods to the page? But then we have to still access the original page for its methods? 
Maybe there's a better way to do this, through inheritance? I see that they call that pattern the `Page Object Model` but that's a little weird, imo... You'd end up wrapping each regular call you'd want to forward to the original? hmm...

cpoile (approved):
Very cool, I almost can't believe it works. I have a few other questions I'll ask synchronously if I haven't figured them out by then. :)

e2e/tests/media.spec.ts
streamer45: > For my own interest -- I'm not sure how this is working, shouldn't there be the calls-screen-source-modal, instead of the screen-player?

Well, we are not testing that at all here. The browser given the `` flag will spit out a valid source without any kind of user consent. And with `#screen-player` we are just making sure the stream gets rendered in the thumbnail (just like in the screenshot above).

> Maybe I'm getting confused by the way the page is used create a devPage, which seems to be a way to add our own methods to the page? But then we have to still access the original page for its methods?
Maybe there's a better way to do this, through inheritance? I see that they call that pattern the Page Object Model but that's a little weird, imo... You'd end up wrapping each regular call you'd want to forward to the original? hmm...

Yeah, it's confusing and could use some improvements but I couldn't figure out an easy way of extending the page class.

PR #353, opened 2023-03-06
MM-50010 - Share common calls code between web and mobile

cpoile:
#### Summary
- Moving the easy to share parts now, we can think about more to share later.
- Mobile is working with this now, just have to do more testing before I'm comfortable pushing a PR.
- Taking the opportunity to update our web dependencies to match webapp

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-50010
- https://mattermost.atlassian.net/browse/MM-50829


webapp/packages/common/package.json
```diff
@@ -0,0 +1,13 @@
+{
+    "name": "@calls/common",
```
cpoile, line 2: Naming is hard. Open to suggestions.

cpoile:
I've manually tested, but I would feel better if the e2e tests ran -- I'm not sure why they're not being triggered...

streamer45:
> I've manually tested, but I would feel better if the e2e tests ran -- I'm not sure why they're not being triggered...

https://community.mattermost.com/private-core/pl/p5hnjqpqr38ofmemr444qwwh1a

PR #357, opened 2023-03-09
MM-51226 - Mobile: Emojis not showing in reaction stream

cpoile:
#### Summary
- The change to `emoji-picker-react` changed the emoji names a bit, and some emojis are not in the emoji name map on the mobile side. So, replace spaces in the names with underscores, and send the emoji literal string in case the map doesn't have the emoji name.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-51226

### DIFF
--- a/webapp/src/components/expanded_view/reaction_button.tsx
+++ b/webapp/src/components/expanded_view/reaction_button.tsx
@@ -52,8 +52,9 @@ export const ReactionButton = forwardRef(({trackEvent}: Props, ref) => {
     const addReactionText = showBar ? 'Close Reactions' : 'Add Reaction';
 
     const handleUserPicksEmoji = (ecd: EmojiClickData) => {
+        const name = ecd.names.length > 0 ? ecd.names[0] : '';
         const emojiData: EmojiData = {
-            name: ecd.names[0].replaceAll(' ', '_'),
+            name: name.replaceAll(' ', '_'),
             skin: ecd.activeSkinTone,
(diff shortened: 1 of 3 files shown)


streamer45:
@cpoile Could you merge against main? E2E are failing cause of a calls-recorder version mismatch.

webapp/src/components/expanded_view/reaction_button.tsx
```diff
@@ -53,9 +53,10 @@ export const ReactionButton = forwardRef(({trackEvent}: Props, ref) => {
 
     const handleUserPicksEmoji = (ecd: EmojiClickData) => {
         const emojiData: EmojiData = {
-            name: ecd.emoji,
+            name: ecd.names[0].replaceAll(' ', '_'),
```
streamer45, line 56: Just confirming that `ecd.names` is guaranteed to be non-empty, right? It seems like it by judging on the code below, but double checking.

streamer45 (approved):
Looks good :+1:

webapp/src/components/expanded_view/reaction_button.tsx
cpoile: Hah, good eye. I did look at the docs and there was nothing stated either way, so I looked at the `emojis.json` and I searched for an empty names list and didn't find one. So I assumed no. But now that you also mention it, it made me worried. So just in case I'll do a test first. :)

PR #359, opened 2023-03-10
MM-51226 - Verify and send the webapp's emoji name, with literal as backup

cpoile:
#### Summary
- #357 assumed the emoji name with spaces should have underscores, but some have hyphens. So, verify that we're sending the short name expected by webapp/mobile, and send the emoji literal as backup.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-51226


webapp/src/components/expanded_view/reaction_button.tsx
```diff
@@ -177,6 +177,25 @@ export const ReactionButton = forwardRef(({trackEvent}: Props, ref) => {
     );
 });
 
+const findEmojiName = (names: string[]) => {
```
streamer45, line 180: Do we know how complex this is? (time/space wise) Assuming this is executed every time someone sends a reaction just want to make sure we are not introducing a performance bottleneck.

cpoile: Everytime they click an emoji, roughly O(N), where N is number of names in the list (from around 1 to 3) : ~1-3 names per list, mapping over those (string and array garbage) and then those lookups in a static map. I considered doing a two stage (like search in the map first, then if not found do the mapping, then lookup again), but I really don't think this is going to be an impact. 
I would maybe be concerned if it was an automated thing, but this is only per emoji click.

streamer45: Sounds good, thanks :+1:

streamer45 (approved):

PR #360, opened 2023-03-10
[MM-49662] UX overhaul

streamer45:
#### Summary

PR implement many UX improvements and fixes including theming support.

Given the amount of little changes touching more or less the whole interface this is likely to be an iterative process.

@tanmay-des I am going to need your help to verify that everything looks good. There are still many open questions and little things to fixed (see open comments on Figma).

Also I temporarily deferred the dynamic sizing work as it's a considerable change which I didn't want to add on top of so much already. I'll work on that once we are happy with this restyling.

#### Designs

https://www.figma.com/file/G1xHnuLDxW4YPacSY4BW08/Calls%3A-UX-Nits


streamer45:
@tanmay-des You can use https://calls.test.mattermost.cloud to test as it should have everything working, recordings included. I updated it to run a build from this PR.

streamer45:
Still far from a pixel perfect match but getting closer I suppose.

https://user-images.githubusercontent.com/1832946/226500826-596741a0-98c8-4713-bde9-0042fa958a0f.mp4

/cc @tanmay-des

tanmay-des (changes requested):
This looks great @streamer45, I have found out some nits `[here](https://www.figma.com/file/G1xHnuLDxW4YPacSY4BW08/Calls%3A-UX-Nits?node-id=847-110072&t=ziV1JLjsplzDcYtw-0)` I think we'll be good to go after that

streamer45:
I updated both PR and test server with the latest nits. @tanmay-des Let me know if there's more to add or if I missed anything and also ping me once you have the updates for the new widget height ready.

streamer45:
@tanmay-des Updated with the latest changes, padding fix on reaction chip and sizing changes on widget (pending some decision on font sizes, left a comment on Figma).

On a slightly related note do you think the widget dimensions should follow the sizing rules based on having multiple teams on Desktop (global widget)? Right now, we default to the wider version which I think it's okay since the widget can be moved outside of MM and could be more practical but let me know if you think we should stick to the web behaviour. 

image

streamer45:
Updated again, I didn't realize the participants list sizing changed once more but now I updated and it should look better.

tanmay-des:
@streamer45 the recent changes looks great. For your previous question, I am afraid users might think that is some sort of a bug. I would first like to know do we know the percentage of users who only have a single team (i.e no sidebar) and are being shown the compact version of the widget? if it a major chunk of our users then I think we might need to figure out a better way to solve the problem.

streamer45:
> @streamer45 the recent changes looks great. For your previous question, I am afraid users might think that is some sort of a bug. I would first like to know do we know the percentage of users who only have a single team (i.e no sidebar) and are being shown the compact version of the widget? if it a major chunk of our users then I think we might need to figure out a better way to solve the problem.

@tanmay-des According to our telemetry data most of our installations have less that 2 teams so it's a common case.

streamer45:
@tanmay-des To unblock this we can simply apply the same logic on desktop as well with the single team using the more compact format. Let me know if that works.

tanmay-des:
Yes @streamer45  I think that will be a good option for now. 
Now that we know it's a common case we can come up with a better solution later.

tanmay-des:
Also, @streamer45 I found out we got a contrast issue in our latest change, I tired out different solutions. Here's what I have in mind. Let's change the label color to center channel text so that the lower hand button have much more clarity and legibility.
image

streamer45:
@tanmay-des I updated as discussed, fixed the widget width on Desktop and changed the hand text color for better contrast.

streamer45:
Thanks @tanmay-des, it should be better now.

tanmay-des:
Screenshot 2023-04-15 at 12 16 04 PM

Small bug I found, if I raise hand while the emoji reactions are present on the screen, it somehow affects all the reactions

streamer45:
Thanks @tanmay-des, updated once more.

tanmay-des (approved):
Thanks @streamer45, I think we have covered everything. 
Looks good to go!

webapp/src/components/call_widget/component.tsx
```diff
@@ -1064,18 +1033,37 @@ export default class CallWidget extends React.PureComponent<Props, State> {
                 >
                     <button
                         className='style--none'
-                        style={{background: device.deviceId === currentDevice?.deviceId ? 'rgba(28, 88, 217, 0.12)' : ''}}
+                        style={{
+                            background: device.deviceId === currentDevice?.deviceId ? 'rgba(28, 88, 217, 0.08)' : '',
```
cpoile, line 1037: I hate to say, but should this be themed?

webapp/src/components/expanded_view/call_participant.tsx
```diff
@@ -56,40 +57,36 @@ export default function CallParticipant(props: Props) {
...
-                            <CompassIcon
-                                icon={'hand-right'}
-                                style={{marginRight: '2px'}}
-                            />
-                        </div>
-                    </>
+                <div style={styles.handRaisedContainer}>
+                    <HandEmoji
+                        style={{
+                            fill: 'rgb(255, 188, 66)',
```
cpoile, line 75: Should this be awayIndicator?

webapp/src/components/expanded_view/component.tsx
```diff
@@ -155,6 +159,128 @@ export default class ExpandedView extends React.PureComponent<Props, State> {
...
+                flex: 1,
+                overflow: 'auto',
+                background: 'rgba(var(--button-color-rgb), 0.08)',
+                borderRadius: '8px',
+                margin: '0 12px',
+            },
+        };
+    };
+
+    private style = this.genStyle();
```
cpoile, line 282: Eventually we should decide on a standard -- scss, style objects, styled components, etc. But not now.

webapp/src/components/expanded_view/component.tsx
```diff
@@ -836,54 +960,55 @@ export default class ExpandedView extends React.PureComponent<Props, State> {
...
-                            <CallDuration
-                                style={{margin: '4px'}}
-                                startAt={this.props.callStartAt}
+                    <div style={this.style.topContainer}>
+                        {this.renderRecordingBadge()}
+                        <CallDuration
+                            style={{margin: '4px'}}
+                            startAt={this.props.callStartAt}
+                        />
+                        <span style={{margin: '4px'}}>{untranslatable('•')}</span>
```
cpoile, line 986: Just curious, do we need the `untranslatable` here?

cpoile (approved):
Wow, this is a lot of work. 😅 
Code looks great, thanks!
I didn't want to point out all the hard-coded colors, I figure you know what you're doing and those shouldn't be themed.
LGTM before we change our minds.

webapp/src/components/expanded_view/component.tsx
streamer45: How do you translate the dot? :p `untranslatable` is really just a way to add exceptions to the linter rules otherwise it will complain there are untranslated strings (not using formatMessage and the like).

webapp/src/components/expanded_view/component.tsx
streamer45: Agreed. If it wasn't for the little special cases (e.g. `:hover`) I'd be happy to keep doing this but yeah we should review and at least avoid CSS in three different ways. Created https://mattermost.atlassian.net/browse/MM-52387

webapp/src/components/expanded_view/call_participant.tsx
streamer45: I've never doubled checked that and assumed it was some custom yellow color but you are right they are the same. Will update, thanks!

webapp/src/components/call_widget/component.tsx
streamer45: I think that's already themed. The background is just an overlay when the device is selected. Unless you mean something else?

streamer45:
@cpoile Let me know on the above conversations.

webapp/src/components/call_widget/component.tsx
cpoile: Nope, just wanted to make sure that hardcoded value was intentional and not a missed theming opportunity. Thanks!

webapp/src/components/expanded_view/component.tsx
cpoile: It complains? Interesting. I thought we could still use some raw strings, like blank spaces, dots, etc. Anyway, no matter, I was just curious.

streamer45: We can't since I added the formatjs linter rule to avoid missing translations. It will complain about any string without translation unless they are excluded.

PR #361, opened 2023-03-11
Fix flaky test

streamer45:
#### Summary

It seems like the test was consistently flaky due to yet another tutorial step appearing.


streamer45:
Turns out the problem was likely that pressing `Enter` may lead to different results based on where the focus is.

cpoile (approved):

PR #362, opened 2023-03-13
[MM-51296] Prevent calls widget from getting umounted while in a call

streamer45:
#### Summary

Turns out there are a couple of actions in MM that will force the widget component to get unmounted either permanently (e.g. switching to admin console) or very briefly (joining a new team). This has not been a big concern so far as the use case is somewhat limited but still going to admin console while in a call seems like a perfectly acceptable use case and the widget completely disappearing doesn't look good. On top of that the internal component state would be cleared leading to the attached issue (and likely more) and all the calls client listeners would be leaked (registered twice or more).

Dealing with internal component state and reconciliation with the calls client during mounts/unmounts is tricky and can lead to even more problems so in order to simplify I am considering shifting the full lifecycle ownership of the widget component to the plugin side and avoid ever unmounting it while on a call. In practical terms this means we stop relying on `registry.registerGlobalComponent` and instead directly render the react component in a custom DOM element appended at body level.

This approach has some clear advantages:

- Truly global component that doesn't depend on the MM view/route users are in.
- Getting full control of when we mount the component can lead to some simplification and avoid potential race conditions during initialization (e.g. https://mattermost.atlassian.net/browse/MM-50751).
- Mounting the component separately could also unblock work to have the widget survive plugin upgrades (https://mattermost.atlassian.net/browse/MM-46431).

My main concern at this point is about whether touching the DOM from this side could have undesired side effects on the rest of the app. I think it should be fine as long as we keep everything in a dedicated place but I am not an expert. Thoughts are welcome.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-51296


cpoile (approved):
This is great, nice work! 

The widget doesn't survive a click on a channel that you're not a member of, because the whole page reloads. But that's expected. (And that reloading will be fixed as soon as the monorepo opens for PRs).

PR #363, opened 2023-03-14
[MM-51322] Add fallback logic for disconnected presenter

streamer45:
#### Summary

@cpoile I looked at logs and there's nothing suggesting the event wasn't sent so I must assume it either got lost somehow or your client didn't receive it for some unrelated reason. I moved the server side logic to a better place but other than that nothing has changed on that side.

I am opting for adding a fallback on the client since we can safely assume that a disconnection means the screen sharing state should be cleared. That said, if websocket events are failing to get to clients, we clearly have more serious problems to deal with.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-51322


cpoile (approved):
That's a good backup.
But... our reliable websockets might not be?

streamer45:
> That's a good backup. But... our reliable websockets might not be?

Hard to prove at this point. We should be logging received messages in the background in case stuff like this happens.

PR #364, opened 2023-03-14
[MM-51405] Fix bad alignment of file recordings icon

streamer45:
#### Summary

Compass icons striking again...this time I couldn't reproduce this outside of Community, even loading the same webapp version so maybe there's something else loading overriding definitions. Either way the problem is that on Community we have a `.CompassIcon` class changing the display to `flex` which causes the element to incorrectly wrap. 

image

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-51405


cpoile (approved):

streamer45:
CircleCI having a bad day it seems :grimacing:

PR #367, opened 2023-03-16
[MM-51439] Bump calls-recorder

streamer45:
#### Summary

@cpoile I think we should consider cutting a dot release `v0.14.1` once this goes in given the severity of the issue. We may want to do it on a separate branch though, need to think about it since I don't want to ship everything else (e.g. i18n support).

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-51439

### DIFF
--- a/plugin.json
+++ b/plugin.json
@@ -137,6 +137,6 @@
     },
     "props": {
         "min_rtcd_version": "v0.9.0",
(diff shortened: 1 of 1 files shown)


cpoile (approved):
I agree: release branch and cherry pick what's essential.

PR #368, opened 2023-03-16
MM-49663 - Prevent race condition between starting and joining

cpoile:
#### Summary
- Every so often on mobile you can make one of the clients fail to set it's UI correctly by pressing the `Join Call` at the same time on two devices. Turns out, this was finding a race on the server where:
  - User A joins call
  - User A creates a call
  - User B joins call
  - User B sends `user_connected` event
  - User A sends `call_started` event
- Which messes up the logic on the mobile side. I could have fixed it on the client side, but this is much more robust I think.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-49663

### DIFF
--- a/server/channel_state.go
+++ b/server/channel_state.go
@@ -43,6 +43,7 @@ type callState struct {
 	RTCDHost        string                `json:"rtcd_host"`
 	HostID          string                `json:"host_id"`
 	Recording       *recordingState       `json:"recording,omitempty"`
+	StartComplete   bool                  `json:"start_complete"`
 }
 
 type channelState struct {
--- a/server/session.go
+++ b/server/session.go
@@ -14,8 +14,11 @@ import (
 )
 
 const (
-	msgChSize = 20
+	msgChSize            = 20
+	startCheckRetryLimit = 10
 )
+
+var errStartIncomplete = errors.New("tried to join a call that has not sent call_start event")
 
 type session struct {
 	userID         string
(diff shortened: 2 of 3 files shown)


streamer45 (commented):
Repeating some of the thoughts I shared internally for context.

I think the existing behaviour is confusing but overall benign as the server state (our single source of truth) is consistent throughout the whole sequence of events. The out of order of ws messages, even though not ideal, can be easily dealt with by clients, as it's the case of the web implementation.

If the change on mobile side is minimal I'd suggest to address it there to fix the issue at hand. If not, given this is somewhat of an edge case, I think we can wait to fix it properly but in a slightly different way.

We have plans to decouple the two actions (starting and joining a call) such as we can have calls running with no participants in them (e.g. scheduled or automated calls). I think moving in that direction will give us a better chance to review this logic and likely make it more consistent as I suspect starting a call could become a simple synchronous HTTP POST request, preventing this edge case altogether.

What do you think?

cpoile:
I'm open to either way, but I'm starting to think that doing this on the server is more correct... The change on the mobile side is not as minimal as I thought. It was designed assuming you wouldn't get a user_connected event before a call_started event. So the mobile state code has been built assuming there is such a thing as a "call" for a user to join, which I realize is different from how things are split up on the webapp client. I can do this on the mobile side, but it will be hacky: make an empty call with a participant list (which will get filled with users as they connect, even if there's no call), and then fill in the rest of the call info once the `call_start` event arrives. We can do it for sure, but it's janky and there is a weird invalid state where you have participants and a call, but not really a call because all the call's data is empty.

What's the main issue with doing this on the server? Or to put it another way, if we could rewrite this, would we allow it to send out a `user_connected` event before the `call_started` event?

streamer45:
Okay we definitely don't want to add an hacky solution to either side.

> What's the main issue with doing this on the server? Or to put it another way, if we could rewrite this, would we allow it to send out a `user_connected` event before the `call_started` event?

No issue at all with doing it on server but I'd prefer to do it without having to resort to any waiting/retrying logic, especially to cover an edge case such as this one. 

One possible solution, as I mentioned above, would be to fully decouple the actions. That should keep things clean and also unblock new functionality.

Alternatively we could grab a cluster lock and make the `handleJoin` function mostly serial but again I am not sure if it's worth it.

cpoile:
I'm all for doing the right thing on the server, so let's chat about fully decoupling the actions. https://mattermost.atlassian.net/browse/MM-51499
Closing this for now. :)

PR #370, opened 2023-03-17
[MM-51483] Fix potential failure in rendering screen sharing track

streamer45:
#### Summary

There is a chance that the screen sharing track gets received before the video player element gets mounted in the DOM which would result in missing the chance to attach the video source. This was spotted during performance testing I am running as several recordings were completely missing the screen content.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-51483

### DIFF
--- a/standalone/src/recording/components/recording_view/index.tsx
+++ b/standalone/src/recording/components/recording_view/index.tsx
@@ -1,7 +1,7 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-import React, {CSSProperties, useEffect, useRef, useState, useCallback} from 'react';
+import React, {CSSProperties, useEffect, useState, useCallback} from 'react';
 import {useIntl} from 'react-intl';
 import {useSelector} from 'react-redux';
 
--- a/standalone/src/recording/components/recording_view/index.tsx
+++ b/standalone/src/recording/components/recording_view/index.tsx
@@ -1,7 +1,7 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-import React, {CSSProperties, useEffect, useRef, useState} from 'react';
+import React, {CSSProperties, useEffect, useRef, useState, useCallback} from 'react';
 import {useIntl} from 'react-intl';
 import {useSelector} from 'react-redux';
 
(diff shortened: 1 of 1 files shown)


cpoile (commented):
I'm not sure what the difference here is -- useState here is being used in basically the same way as useRef was, so I'm not sure how this fixes the problem?

streamer45:
> I'm not sure what the difference here is -- useState here is being used in basically the same way as useRef was, so I'm not sure how this fixes the problem?

`useEffect` definitely wasn't triggering when the ref got set and it seems like it's not a good idea to use it for refs anyway. Now we are using a callback instead and setting the state to make sure `useEffect` triggers.

Mostly following https://stackoverflow.com/questions/60476155/is-it-safe-to-use-ref-current-as-useeffects-dependency-when-ref-points-to-a-dom

cpoile:
> `useEffect` definitely wasn't triggering when the ref got set and it seems like it's not a good idea to use it for refs anyway. Now we are using a callback instead and setting the state to make sure `useEffect` triggers.

Ah I see!
Just a minor nit: in the original code, the `screenPlayerRef` wasn't a dependency in the `useEffect`. So although the SO answer wasn't exactly the problem, the underlying issue was still the same. So nice find! :)

standalone/src/recording/components/recording_view/index.tsx
```diff
@@ -102,7 +106,7 @@ const RecordingView = () => {
         return (
             <div style={style.screenContainer as CSSProperties}>
                 <video
-                    ref={screenPlayerRef}
+                    ref={screenRefCb}
```
cpoile, line 109: I think you can use the `setScreenPlayerNode` directly, and then there's no need for the `useCallback`: 
```suggestion
                    ref={setScreenPlayerNode}
```

cpoile (approved):

standalone/src/recording/components/recording_view/index.tsx
streamer45: @cpoile Happy to simplify but are we 100% confident? I can run a few more tests but wasn't sure myself so opted for the more elaborate approach.

streamer45: I don't think that's working. What we could do is keep everything as it was and just add the `ref.current` dependency to `useEffect` but I still can't figure out if that's safe to do in our case.

cpoile: > don't think that's working. What we could do is keep everything as it was and just add the `ref.current` dependency to `useEffect` but I still can't figure out if that's safe to do in our case.

That probably won't work, the `ref.current` won't trigger a rerender (that's what the SO question was about, afaict). 
The `ref={setScreenPlayerNode}` should work... But I guess it's not a big deal, just thought it would simplify things. :)

PR #371, opened 2023-03-17
[MM-51443] Add support for configuring recording quality settings

streamer45:
#### Summary

PR adds some basic way to configure the call recordings quality. This is to give system admins some degree of control on the overall performance impact, scalability and recordings file sizes. 

Full raw analysis and reasoning behind the specifics of the proposed settings can be found at https://gist.github.com/streamer45/a57a8cb79cf1ea534611601a6df4ef26 (**warning:** not for the faint-hearted).

I have some next steps to extract the most useful content from that and turn it into documentation (https://mattermost.atlassian.net/browse/MM-51496).

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-51443

#### Related PRs

https://github.com/mattermost/calls-recorder/pull/26
https://github.com/mattermost/calls-offloader/pull/20


cpoile (approved):
I can imagine that admins are going to want to specify custom settings. :)

About this line in the perf tests: `Scaling up is worse than linear as adding more recording jobs will demand exponentially more resources.`
Can you explain that? Why is it exponential?

iirc, h264 is very old now, can't we try h265 or av1?

streamer45:
> I can imagine that admins are going to want to specify custom settings. :)
> 
> About this line in the perf tests: `Scaling up is worse than linear as adding more recording jobs will demand exponentially more resources.` Can you explain that? Why is it exponential?

Not exactly exponential but clearly not linear, meaning that running 8 recording jobs at Low quality takes more than double the CPU consumption of running 4 (33% vs 74%), and the gap gets wider with higher qualities. We should be able to easily map data points to a function if you are really curious. Why it happens, well there are probably many reasons, surely there's some lock contention going on at system level but almost nothing scales linearly to be honest.

> iirc, h264 is very old now, can't we try h265 or av1?

Old and kicking :) Truth is that h264 is the only thoroughly supported format nowadays and our only realistic chance to avoid multiple transcodings to support all the OS/devices combos. 

`av1` is the future but its performance is still abysmal especially for real-time scenarios, such as this one.

streamer45:
> I can imagine that admins are going to want to specify custom settings. :)

At some point we may expose the whole thing through some advanced config for the more nerdy admins but for now I think it makes sense to provide proven and tested profiles.

streamer45:
Added a second iteration of the analysis at https://gist.github.com/streamer45/a57a8cb79cf1ea534611601a6df4ef26?permalink_comment_id=4513906#gistcomment-4513906

PR #373, opened 2023-03-17
[MM-51500] Use correct channelID when setting profile pictures in store

streamer45:
#### Summary

Not the first time this hit me. The events sent to the bot are not broadcasted for obvious reasons so the channelID will be sent as part of the data. Profile pictures were failing to get stored properly and the component wouldn't render, causing the weird behaviour.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-51500


cpoile (approved):
ahhh. Nice. 👍

PR #377, opened 2023-03-21
[MM-51558] Stop recording job on bot disconnect or call end

streamer45:
#### Summary

We were failing to stop a recording on call end as we were incorrectly assuming the bot would be the last one to leave the call. 

PR fixes this by attempting to stop the recording any time the bot leaves a call with one ongoing. This is usually benign, meaning a stop action was triggered or the timeout fired but there could still be instances where the bot disconnects and the recording job continues until timeout. In such cases we want to stop as soon as we detect a disconnection so a new recording can be started if necessary.

I also added a failsafe call in the call end API endpoint in case of a stuck call with an ongoing recording.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-51558

### DIFF
--- a/server/api.go
+++ b/server/api.go
@@ -235,7 +235,15 @@ func (p *Plugin) handleEndCall(w http.ResponseWriter, r *http.Request, channelID
 			return
 		}
 
-		p.LogInfo("call state is still in store, force ending it")
+		p.LogInfo("call state is still in store, force ending it", "channelID", channelID)
+
+		if state.Call.Recording != nil && state.Call.Recording.EndAt == 0 {
+			p.LogInfo("recording is in progress, force ending it", "jobID", state.Call.Recording.JobID)
+
+			if err := p.jobService.StopJob(state.Call.Recording.JobID); err != nil {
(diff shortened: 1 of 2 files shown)


server/api.go
```diff
@@ -235,7 +235,15 @@ func (p *Plugin) handleEndCall(w http.ResponseWriter, r *http.Request, channelID
 			return
 		}
 
-		p.LogInfo("call state is still in store, force ending it")
+		p.LogInfo("call state is still in store, force ending it", "channelID", channelID)
+
+		if state.Call.Recording != nil && state.Call.Recording.EndAt == 0 {
+			p.LogInfo("recording is in progress, force ending it", "jobID", state.Call.Recording.JobID)
```
cpoile, line 241: We may want to log the channelID here (and in the other log statements added below), to make it simpler?

cpoile (approved):
Nice!

server/api.go
streamer45: It certainly won't hurt. Updated.

PR #380, opened 2023-03-22
MM-50533 - Prioritize raised hand when ordering participants

cpoile:
#### Summary
- Participants who raised their hands first will always be ordered first, maintaining the raised hand order regardless of other states (e.g., voice on).
- Mobile: https://github.com/mattermost/mattermost-mobile/pull/7229

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-50533

### DIFF
--- a/webapp/src/utils.ts
+++ b/webapp/src/utils.ts
@@ -1,37 +1,31 @@
-import {parseSemVer} from 'semver-parser';
-
-import {
-    getCurrentRelativeTeamUrl,
-    getCurrentTeamId,
-    getTeam,
-} from 'mattermost-redux/selectors/entities/teams';
-
-import {getCurrentUserId} from 'mattermost-redux/selectors/entities/users';
-import {Client4} from 'mattermost-redux/client';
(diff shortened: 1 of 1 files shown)


webapp/src/utils.ts
```diff
@@ -162,12 +156,6 @@ export function stateSortProfiles(profiles: UserProfile[], statuses: { [key: str
             };
         }
 
-        if (stateA.unmuted && !stateB.unmuted) {
-            return -1;
-        } else if (stateB.unmuted && !stateA.unmuted) {
-            return 1;
-        }
-
         if (stateA.raised_hand && !stateB.raised_hand) {
```
streamer45, line 159: I suppose it's far less likely or useful but should we have this take precedence over the presenter state as well?

streamer45 (approved):
Thanks :+1:

webapp/src/utils.ts
streamer45: I think my reasoning is purely to keep the queue consistency at all costs but usability wise yeah, I don't think it should ever be a problem. Let's merge this as is and we can revisit anytime if needed.

PR #381, opened 2023-03-22
[MM-51592] Add logic to cover reconnect edge case

streamer45:
#### Summary

PR fixes some interesting issues that have been spotted on Community in the last few days. Participants get kicked out of a call almost randomly after successfully reconnecting to the websocket channel. Today our dev meeting recording was cut short due to the calls bot suffering this unfortunate fate.

The problem is that one assumption was made when implementing the reconnection logic. I assumed that a client should always get back the same connection ID if reconnecting to the same instance (or HA node) due to the reliable websocket implementation allowing to re-use the same connection object. For some reasons, yet to be understood, this is not happening consistently.

I still think there's something suspect on the server side which I'll be investigating separately but still it makes sense to cover for such case just to be on the safe side.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-51592

### DIFF
--- a/e2e/tests/start_call.spec.ts
+++ b/e2e/tests/start_call.spec.ts
@@ -135,6 +135,28 @@ test.describe('start new call', () => {
 
         await devPage.leaveCall();
     });
+
+    test('ws reconnect', async ({page}) => {
+        const devPage = new PlaywrightDevPage(page);
+        await devPage.startCall();
+        await devPage.wait(1000);
+
+        const reconnected = await page.evaluate(() => {
+            return new Promise((resolve) => {
+                window.callsClient.ws.on('open', (connID: string, originalConnID: string, isReconnect: boolean) => {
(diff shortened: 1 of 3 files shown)


e2e/tests/start_call.spec.ts
```diff
@@ -135,6 +135,28 @@ test.describe('start new call', () => {
...
+        const reconnected = await page.evaluate(() => {
+            return new Promise((resolve) => {
+                window.callsClient.ws.on('open', (connID: string, originalConnID: string, isReconnect: boolean) => {
+                    resolve(isReconnect);
+                });
+                window.callsClient.ws.ws.close();
+            });
+        });
+
+        expect(reconnected).toBe(true);
```
cpoile, line 153: This is really nice.

cpoile (approved):
Nice!

PR #385, opened 2023-03-23
MM-51590: Remember recording alert dismiss state in ExpandedView

cpoile:
#### Summary
- Simply setting global state, I tried to do it in as typesafe a way as possible. 

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-51590


webapp/src/components/expanded_view/recording_info_prompt.tsx
```diff
@@ -81,7 +88,7 @@ export default function RecordingInfoPrompt(props: Props) {
                     defaultMessage: 'Your recording will end in {count, plural, =1 {# minute} other {# minutes}}.'}
                 , {count: getMinutesLeftBeforeEnd()})}
                 confirmText={formatMessage({defaultMessage: 'Dismiss'})}
-                onClose={() => updateDismissedAt(Date.now())}
+                onClose={updateDismissed}
```
cpoile, line 91: Not strictly necessary, could use the original, but thought it's good to be consistent in case we add something in the future to the `updateDismissed` block that affects here.

webapp/src/reducers.ts
```diff
@@ -565,7 +565,7 @@ const voiceChannelCalls = (state: { [channelID: string]: callState } = {}, actio
             [action.data.channelID]: {
                 ...state[action.data.channelID],
                 hostID: action.data.hostID,
-                hostChangeAt: Date.now(),
+                hostChangeAt: action.data.hostChangeAt || state[action.data.channelID].hostChangeAt,
```
cpoile, line 568: This caused some confusion. If I had a 'shaking fist' emoji. :)

webapp/src/components/call_widget/component.tsx
```diff
@@ -1440,7 +1440,12 @@ export default class CallWidget extends React.PureComponent<Props, State> {
                 body={body}
                 confirmText={confirmText}
                 declineText={isHost ? null : formatMessage({defaultMessage: 'Leave call'})}
-                onClose={() => this.setState({recDisclaimerDismissedAt: Date.now()})}
+                onClose={() => {
+                    this.setState({recDisclaimerDismissedAt: Date.now()});
+                    if (window.currentCallData) {
+                        window.currentCallData.recording.promptDismissedAt = Date.now();
+                    }
+                }}
```
cpoile, line 1448: This is ugly, but since it's just this component there's no point in pulling it out and using a `useCallbasck`.

webapp/src/reducers.ts
streamer45: Indeed. If the timestamp was sent by the server the bug would have been avoided in the first place.

cpoile: Yep, sorry I misunderstood what you meant.  👍

streamer45 (approved):
Nice! I suppose we don't want to go as far as closing on widget -> prompt gets closed on pop out and vice versa.

standalone/src/init.ts
```diff
@@ -1,92 +1,80 @@
-import {Reducer} from 'redux';
+import {
+    CallHostChangedData,
+    CallStartData,
```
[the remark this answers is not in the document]
cpoile: Sorry, I turned on optimize imports, so it's going to sort every file it touches. Maybe I should go through and sort them all at once?

PR #389, opened 2023-04-05
[MM-51218] Fix global widget bounds

streamer45:
#### Summary

The self-join notification was still accounting for the total widget height, causing the wrong calculation. Fixing that by turning it into its own component.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-51218


streamer45:
Also pushed a fix for the flaky e2e snapshots.

@cpoile When you get a chance, please give this a try and see if it fixes the issue on your end as well. Thanks :)

cpoile (approved):
Thanks, lgtm!

PR #390, opened 2023-04-05
Improve joined users widget notification

streamer45:
#### Summary

PR reworks the (now broken) abysmal logic to show the notification pill for recently joined participants. 
Still the CSS animation could use some improvements but that's a bit more tricky and will need some figuring out.

### DIFF
--- a/e2e/tests/join_call.spec.ts
+++ b/e2e/tests/join_call.spec.ts
@@ -36,7 +36,12 @@ test.describe('join call', () => {
         expect(await joinCallToast.screenshot()).toMatchSnapshot('channel-toast.png');
 
         await joinCallToast.click();
+
+        await expect(userPage.page.getByTestId('call-joined-participant-notification')).toBeVisible();
+        await expect(userPage.page.getByTestId('call-joined-participant-notification')).toContainText(userState.users[userIdx].username + ' has joined the call.');
+
         await expect(page.locator('#calls-widget')).toBeVisible();
+
         const devPage = new PlaywrightDevPage(page);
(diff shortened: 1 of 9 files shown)


cpoile (commented):
Awesome, it fixes the joining notification never firing problem.
But there's some issues with desktop -- 
1. on joining you get the previous joined members appearing as if they had just joined, and 
2. the v spacing is a bit off with the muted notification:

Actually, that happens on webapp as well: 
image

streamer45:
Thanks @cpoile, could you give it another go?

cpoile (approved):
Nice!

Two nits (don't need to be fixed if it's tough):
There's still some spacing issues with the muted notification on Chrome (not on Firefox, for a first). 
image

And if you join right after person A (before the Person A's join has been removed), you'll get  person A's notification of course, but I don't think that's a big deal. (You could think of that as almost like a feature, to see who had joined "right around the same time" as you did.)

streamer45:
> There's still some spacing issues with the muted notification on Chrome (not on Firefox, for a first).

@cpoile Did you refresh your cache? I don't see that but I suspect it may be some outdated CSS lingering.

> if you join right after person A (before the Person A's join has been removed), you'll get person A's notification of course

Mmm, that sounds like a potential race but I am not overly worried given the use case.

cpoile:
> @cpoile Did you refresh your cache? I don't see that but I suspect it may be some outdated CSS lingering.

Indeed it was, works great now, thanks!

> Mmm, that sounds like a potential race but I am not overly worried given the use case.

👍

PR #401, opened 2023-04-12
Fix wider widget on desktop

streamer45:
#### Summary

It looks like https://github.com/mattermost/mattermost-plugin-calls/pull/393 wasn't really working on desktop due to a missing redux call preventing the selector from returning the teams.

Unfortunately this means that anyone using Desktop on 0.15.0 is essentially stuck with the shorter format.

@cpoile Could you give this a try on your side? May also be worth considering a dot release.


cpoile (approved):
Yes, this fixes it, thank you!

PR #409, opened 2023-04-26
[MM-52346] Update job service API

streamer45:
#### Summary

Updating the job service API to reflect the new changes.

#### Related PR

https://github.com/mattermost/calls-offloader/pull/26

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-52346


streamer45:
E2E are expected to fail until we update everything on the Gitlab side.

server/job_service.go
```diff
@@ -232,15 +233,20 @@ func (p *Plugin) newJobService(serviceURL string) (*jobService, error) {
...
 }
 
-func (s *jobService) StopJob(jobID string) error {
-	return s.client.StopJob(jobID)
+func (s *jobService) StopJob(channelID string) error {
+	// Since MM-52346, stopping a job really means signaling the bot it's time to leave
+	// the call. We do this implicitly by sending a fake call end event.
+	s.ctx.publishWebSocketEvent(wsEventCallEnd, map[string]interface{}{
+		"channelID": channelID,
+	}, &model.WebsocketBroadcast{UserId: s.ctx.getBotID(), ReliableClusterSend: true})
```
streamer45, line 245: Not super happy with this but it's simple enough and a bit more explicit. One alternative could be to don't send anything at all and check on the client (recorder) side for the recording ended state websocket event and disconnect upon receiving that one. Let me know what you think.

cpoile: Either way, doesn't seem to be that big of an issue to me. This is like telling the bot the call is "over", so that's fine. On the other hand, having the recording act on the recording ended websocket event is a bit more logical or "reactive". But up to you.

cpoile (approved):

PR #413, opened 2023-04-27
MM-48534 - Fix: Call post component honouring user's time format

cpoile:
#### Summary
- Honour the user's 24hr preference
- Convert post component to luxon as a bit of a trial-run for the rest of the code (had to introduce a helper fn because luxon isn't as full featured as moment yet).
- As you can see, the "humanized" version of the text from Luxon is very different from moment. It doesn't do the "a few seconds ago" (for relative) or "a few seconds" (for duration) thing. And we can adjust the specificity/resolution of the `Lasted ...` line, maybe to minutes if you think that's better.

image

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-48534
- https://mattermost.atlassian.net/browse/MM-49769


streamer45 (commented):
A few questions: 

1. If `luxon` is not a real replacement for `moment` why use it?
2. How are we handling this kind of cases on web and mobile?
3. How hard is it going to be to make this fully localized?

cpoile:
@streamer45 
1. As we chatted about (but to put here for the record), `moment` is deprecated and we're supposed to move to `luxon` going forward, it's exported by webapp (no bundling cost). Although webapp still uses `moment` as a fallback (but it's not exported). /shrug
2. The `Timestamp` component does this, it's pretty exhaustive and would do everything we needed. It's not componentized, but with some webpack hackery I was able to import it from webapp. So we can use that if we want, there's not a lot of difference for this simple case, so I'm not sure if the complication is needed.
3. Luxon is localized using `Intl`, including the 24 hour format. For example, with locale set to Japanese, but format set to 12hr, the output is:

image

24hr:
image

streamer45:
@cpoile Thanks for taking the time to investigate this further. My main concern was around localization support which you confirmed to be available so happy to move this forward.

streamer45 (commented):
Maybe we should involve UX but a few comments I have on this:

image

- First case seems odd. I'd expect to see either seconds or "less than a minute" or "a few seconds".
- I think I liked the expanded version better, especially since we have the space. So "3 min" would ideally become "3 minutes".

cpoile:
@streamer45 Thanks, good suggestions.

1. I've followed moment's style and am using 'a few seconds' for duration < 44 seconds, and rounding up anything between 44 sec and a minute to 1 minute (to prevent the '0 seconds' case). We can add the rest of moment's humanize if you'd like, that just wordifies the singular versions of each timestop (1 minute -> a minute, 1 hour -> an hour, etc.). For a list, see: https://momentjs.com/docs/#/durations/humanize/  It's up to you if you want to do that, but I think it just adds more strings we would have to translate.
2. Sounds good, done.

A few learnings:

I spent some time with the `Timestamp` component from the webapp, and it turns out it's exported (no ugly hacks needed). But it doesn't do durations, unfortunately, which is what we mostly need. So it won't be much help.

I found that Luxon is better than moment (except for the humanize feature). For one, it's properly internationalized -- before we weren't translating our timestamps or durations (!), but now we are (except for the `a few seconds` which we have to weblate).  Second, getting rid of moment reduces our `main.js` size from `2926839 B` to `2436035 B` (~17% reduction).

In moving from moment to luxon, we're losing the humanized relative timestamps too (the "a few seconds ago", "a minute ago", when you start a call). But we gain internationalization, and I think a better experience: 1) the moment humanization wasn't capitalized correctly anyway, 2) it rounds to the minute, so after the seconds it stays at the minute resolution (so not too distracting).  If we want to revisit later, we can add some humanization (though I'd have to look into it).

streamer45:
Thanks @cpoile for the excellent analysis and work :raised_hands: 

Somehow this is not working as expected on my end as I am still seeing the following happen (running commit b0e4dbc9e9bf57cc160a4f0fa1ae7614012bc933):

0 seconds shows upon starting a call
Then it moves to 5 seconds and stays there for a while
Then it really depends when the component re-renders which could mean you get it stuck at xx seconds for a long time.

streamer45 (approved):
Brilliant :+1:

PR #418, opened 2023-05-02
[MM-48095] Mitigate potential inconsistencies with local client state

streamer45:
#### Summary

PR introduces a couple of changes to how we store and render local state, specifically muted/unmuted and raised/lowered hand. This is an effort to mitigate potential inconsistencies in case of unreliable network causing websockets events to be delayed, or worse, lost.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-48095


cpoile (approved):
Cool, let's see if it helps!

streamer45:
Forgot to tweak a couple of e2e tests as some were relying on the former client state fields.

PR #419, opened 2023-05-03
[MM-52604] Refactor e2e users state generation

streamer45:
#### Summary

PR refactors a bit how we are handling users state in e2e tests. Originally we'd be using a constant array for simplicity but that wasn't ideal so we are removing that and rely on the state of the server instead by generating usernames on the fly through the expected pattern (`calls-user` prefix + test index).

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-52604

### DIFF
--- a/e2e/tests/recordings.spec.ts
+++ b/e2e/tests/recordings.spec.ts
@@ -1,6 +1,8 @@
 import {test, expect, chromium} from '@playwright/test';
 
-import {userState, baseURL, defaultTeam, pluginID} from '../constants';
+import {baseURL, defaultTeam, pluginID} from '../constants';
+
+import {getUserStoragesForTest} from '../utils';
 
 import PlaywrightDevPage from '../page';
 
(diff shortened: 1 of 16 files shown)


e2e/tests/recordings.spec.ts
```diff
@@ -10,7 +12,7 @@ test.beforeEach(async ({page, context}) => {
 });
 
 test.describe('call recordings', () => {
-    test.use({storageState: userState.users[6].storageStatePath});
```
streamer45, line 13: This was the line causing the flakyness. Forgot to update the index which meant it could occasionally conflict with other tests running in parellel.

streamer45:
> Cool -- for a minute I didn't understand how the indexing worked (using 0 and 1 for every test), but I think I get it now (the array is unique per test runner). +1

Yes, we allocate 2 users and 2 channels for each worker so that they never conflict. This way tests can be run in full parallel mode.

PR #421, opened 2023-05-04
[MM-52648] Fix inconsistent participant name display in channel toast

streamer45:
#### Summary

PR currently fixes the inconsistency but as @matthewbirtch suggested, it would be good to do a revision of all the places where we display participants names and check whether we should apply user display settings:

image

@tanmay-des Could you help with this? Some of the places would be:

- Participants list (both widget and expanded view)
- Participants grid (expanded view)
- Connected profiles in channel toast tooltip
- Connected profiles in chat thread root post tooltip
- Connected profiles in sidebar active call icon tooltip
- Reaction stream

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-52648

### DIFF
--- a/webapp/src/components/connected_profiles.tsx
+++ b/webapp/src/components/connected_profiles.tsx
@@ -3,7 +3,7 @@ import {OverlayTrigger, Tooltip} from 'react-bootstrap';
 
 import {UserProfile} from '@mattermost/types/users';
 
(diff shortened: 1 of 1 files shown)


streamer45:
I suppose the second part is strictly related to https://mattermost.atlassian.net/browse/MM-46310

PR #430, opened 2023-05-11
MM-527187 - Render rtcd-related settings conditionally

cpoile:
#### Summary
- Doing the same thing we did with recording on/off. 
- Included `Server side turn`, based on the plugin code (please let me know if I'm wrong) and changed the wording to `On` and `Off`
- One choice I made: I moved the settings to below the rtcd setting, so that you can see what appears/disappears clearly. I'm screencapping what it looks, and take a look at the situation of no licence -- I fixed an issue there where we were *not* showing the settings, even if the licence got downgraded, so I fixed that. Also, notice that the settings break up the `Try free` blocks -- I kind of like having the `try free` blocks next to each other, but I chose to do that so that the settings pop in and out *below* the setting, same as the `enable recording` setting. Open to putting it above, if you value having the `try free` blocks next to each other.

### No RTCD setting:
image

### RTCD setting filled in:
image

### Non-enterprise:
image

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-52718


cpoile:
@streamer45 Let me know what you think -- I think the extra help text might be a little noisy, but I think it might be okay enough.

image
image

streamer45:
> @streamer45 Let me know what you think -- I think the extra help text might be a little noisy, but I think it might be okay enough.

Looks good! Could we show it only in the disabled case or do you find that confusing?

The UDP port text value doesn't seem disabled in your second screen, not sure if it's intended.

streamer45:
Oh okay, I guess I got confused by the placeholder for the address.

webapp/i18n/en.json
```diff
@@ -105,6 +105,7 @@
   "dCb7CD": "Start call",
   "dmDHwc": "{count, plural, =1 {# recording} other {# recordings}} available",
   "eB0u7Z": "Calls are a quick, audio-first, way to interact with your team. Upgrade to Mattermost Professional to use calls in channels and group messages.",
+  "eP+xGA": "Disabled because the <code>RTCD service URL</code> field is non-empty.",
```
streamer45, line 108: Some non-blocking thoughts:

- The `<code>` formatting seems a bit strong on the eye to me. Could we switch to italic maybe? That would be consistent with how we mention fields in our documentation I believe.
- What do you think about rephrasing to something like `Disabled as not applicable when the RTCD service URL field is non-empty.`?

plugin.json
```diff
@@ -1,171 +1,170 @@
...
+        "hosting": "on-prem"
+      },
+      {
+        "key": "ServerSideTURN",
+        "display_name": "Server Side TURN",
+        "type": "bool",
+        "default": false,
+        "help_text": "(Optional) When set to on it will pass and use configured TURN candidates to server initiated connections.",
+        "hosting": "on-prem"
+      },
```
streamer45, line 65: This should go later, after all the other TURN settings.

webapp/i18n/en.json
cpoile: - Ok, changed to italic, but I personally don't like it as much as the code formatting. But it's not a big deal for me, so whatever you prefer is great.
- rephrased to: `Not applicable when the RTCD service URL field is in use.` which is a bit clearer imo. Let me know?

CleanShot 2023-05-24 at 15 49 03@2x

streamer45: > * Ok, changed to italic, but I personally don't like it as much as the code formatting. But it's not a big deal for me, so whatever you prefer is great.

Fair enough. What about a link to https://docs.mattermost.com/configure/plugins-configuration-settings.html#rtcd-service-url? 

> * rephrased to: `Not applicable when the RTCD service URL field is in use.` which is a bit clearer imo. Let me know?

Yep, looks great.

streamer45 (approved):
Thanks 👍

webapp/i18n/en.json
cpoile: Yep, that looks fine to me: 

CleanShot 2023-05-24 at 15 57 36@2x

streamer45: :100:

streamer45:
@cpoile I think we just need to retrigger e2e on this one and should be good to go.

Issue #309, opened 2023-01-25
Trigger start a call from mattermost API


bayunvnt:
Hello, i tried to post a "/call start" from API "https://mattermost.domain.id/api/v4/commands/execute", but the calls is not started, there is any way to trigger call start from an api? thankyou

streamer45:
@bayunvnt Nope, unfortunately it's not possible as a call only exists as long as at least one user is connected so to start a call you'd have to create a websocket connection and keep it alive. 

That said, in the near future we may be able to make this more automatable by relying on the recently introduced Calls bot. 

By the way, what you are looking to achieve exactly? Getting to understand your use case would be very helpful.

bayunvnt:
thankyou for your response, i recently use mattermost for communicating with teams, everyday in the morning we need to do a meeting about taslisks, jobs and etc, i want to automate start a call in the morning with an energic message for the rest of teams so they can join directly without manually start a calls.

streamer45:
Thanks for the context. I created an internal ticket ([MM-49965](https://mattermost.atlassian.net/browse/MM-49965)) to track this as it sounds like a very useful addition. Will give an update here as soon as get to implement it.

PR #437, opened 2023-05-29
[MM-51499] Implement API endpoint to start calls

streamer45:
#### Summary

PR proposes a new API endpoint to start calls. This action becomes effectively decoupled from joining as a call can be now started prior to any client connecting to it. This opens interesting opportunities for automation and integration (e.g. calendar schedule). On top of the new changes there's some minor refactoring mainly to extract common application logic and keep it away from the API layer.

Fixes https://github.com/mattermost/mattermost-plugin-calls/issues/309

To note, this PR does not alter the way clients join calls. The two actions (start and join) are still coupled when connecting to calls. On this front more care is needed to avoid breaking changes and introducing more racy behaviour. I am still not entirely sure what the best approach should be but will defer to a future PR.

@cpoile At this point I am mostly looking for a quick pass to start a discussion on whether any of this makes sense. As this would mark the first attempt at supporting a public API for Calls, I'd like to take this step with some extra caution. Some still open questions I'd like to go through:

- We should at least have some basic documentation for the endpoints and expected data format. Not sure whether in this repository itself (may be easier to keep up to date) or the in official docs.
- Should we version the API in any way?
- Should we also provide a sample client or extend [the main Golang one](https://github.com/mattermost/mattermost-server/blob/master/server/public/model/client4.go)?

#### Example

*Start a call*

Request body (JSON):

```json
{
  "channel_id": "", // required
  "thread_id": "", // optional
  "title": "" // optional
}
```

```sh
curl -v -X POST -H 'Authorization: Bearer xxx' \
  http://localhost:8065/plugins/com.mattermost.calls/calls/start -d '{"channel_id": "3gpph3ezpp83fqndj456fd31oa"}'
```

*End a call*

Empty request body.

```sh
curl -v -X POST -H 'Authorization: Bearer xxx' \
  http://localhost:8065/plugins/com.mattermost.calls/calls/3gpph3ezpp83fqndj456fd31oa/end
```

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-49965
https://mattermost.atlassian.net/browse/MM-51499

### DIFF
--- a/server/websocket.go
+++ b/server/websocket.go
@@ -465,39 +465,12 @@ func (p *Plugin) handleLeave(us *session, userID, connID, channelID string) erro
 	return nil
 }
 
-func (p *Plugin) handleJoin(userID, connID, channelID, title, threadID string) error {
-	p.LogDebug("handleJoin", "userID", userID, "connID", connID, "channelID", channelID)
+func (p *Plugin) handleJoin(userID, connID string, req CallStartRequest) error {
+	p.LogDebug("handleJoin", "userID", userID, "connID", connID, "channelID", req.ChannelID)
 
-	// We should go through only if the user has permissions to the requested channel
-	// or if the user is the Calls bot.
-	if !(p.isBot(userID) || p.API.HasPermissionToChannel(userID, channelID, model.PermissionCreatePost)) {
-		return fmt.Errorf("forbidden")
-	}
-	channel, appErr := p.API.GetChannel(channelID)
-	if appErr != nil {
-		return appErr
-	}
-	if channel.DeleteAt > 0 {
-		return fmt.Errorf("cannot join call in archived channel")
-	}
-
-	if threadID != "" {
-		post, appErr := p.API.GetPost(threadID)
-		if appErr != nil {
-			return appErr
-		}
-
-		if post.ChannelId != channelID {
-			return fmt.Errorf("forbidden")
-		}
-
-		if post.DeleteAt > 0 {
-			return fmt.Errorf("cannot attach call to deleted thread")
-		}
-
-		if post.RootId != "" {
-			return fmt.Errorf("thread is not a root post")
-		}
+	channel, err := p.getChannelForCall(userID, req)
+	if err != nil {
+		return fmt.Errorf("failed to get channel for call: %w", err)
 	}
 
 	state, prevState, err := p.addUserSession(userID, connID, channel)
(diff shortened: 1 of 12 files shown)


server/websocket.go
```diff
@@ -465,83 +465,23 @@ func (p *Plugin) handleLeave(us *session, userID, connID, channelID string) erro
 	return nil
 }
 
-func (p *Plugin) handleJoin(userID, connID, channelID, title, threadID string) error {
-	p.LogDebug("handleJoin", "userID", userID, "connID", connID, "channelID", channelID)
+func (p *Plugin) handleJoin(userID, connID string, req CallStartRequest) error {
```
cpoile, line 468: It might be a little confusing to have `CallStartRequest` as a param to `handleJoin` -- even if the data is the same. What do you think?

server/websocket.go
```diff
@@ -465,83 +465,23 @@ func (p *Plugin) handleLeave(us *session, userID, connID, channelID string) erro
...
-					ChannelId: channelID,
-					Message:   "Currently calls are not enabled for non-admin users. You can change the setting through the system console",
-				},
-			)
-		}
-
-		postID, threadID, err := p.startNewCallPost(userID, channelID, state.Call.StartAt, title, threadID)
-		if err != nil {
+	} else if prevState.Call == nil {
+		if err := p.handleCallStarted(state.Call, req, channel); err != nil {
```
cpoile, line 482: The `addUserSession` -> `handleCallStarted` flow is a bit confusing to me... If I were doing it from scratch, I would want to check if there's an ongoing call, if not then start one. That seems a bit more clear?

cpoile (commented):
Looks fine to me! I left a couple thoughts for your consideration.

> We should at least have some basic documentation for the endpoints and expected data format. Not sure whether in this repository itself (may be easier to keep up to date) or the in official docs.

If we make the client a part of this repo, then the docs should go here too, imo.

> Should we version the API in any way?

If we publish a client, sure, makes sense.

> Should we also provide a sample client or extend [the main Golang one](https://github.com/mattermost/mattermost-server/blob/master/server/public/model/client4.go)?

If we want to publish a client, I think it's easiest to do it in this repo. We could follow what we did in the old Playbooks repo.

streamer45:
Closing as it will be much easier to create it from scratch after all the changes, especially SQL store.

PR #438, opened 2023-05-31
Drop pluginapi

streamer45:
#### Summary

To unblock https://mattermost.atlassian.net/browse/MM-52924 we have to update the mattermost-server dependency to make use of the new config hook. This is currently not possible as we depend on [`mattermost-plugin-api`](https://github.com/mattermost/mattermost-plugin-api). There is an [ongoing discussion](https://community.mattermost.com/core/pl/8dq8ha9aoidf5rxbrgkenxj7oh) to potentially move the functionality to the monorepo but until that's resolved we are blocked. 

Given we don't make heavy use of the API wrapper I am thinking of dropping the dependency altogether by porting the necessary code, mostly the cluster mutex and some license utils.


cpoile (approved):
Yah, doesn't look like it'll be moved over anytime soon, and if we're blocked let's fix that. 👍

PR #442, opened 2023-06-01
Bump minimum supported server version to current ESR

streamer45:
#### Summary

7.7 support just ended last month. Going forward it would be good to keep the current ESR as our target minimum version unless strictly necessary to make an emergency bump.


cpoile (approved):
Sounds good. 👍

PR #443, opened 2023-06-02
MM-53004: apply intl to username conjunctions

calebroseland:
#### Summary
Replaces `utils.getUsersList` with calls to the internationalized `formatList`. Context: https://community.mattermost.com/core/pl/315jwqduqinkxd1fgpm5ixg8de

##### One caveat 
Instead of always being en, it is always translated—so the list segment may be translated when the surrounding message isn't.

CleanShot 2023-06-02 at 16 36 18@2x

##### The benefit
All variations/lanugages are handled automatically.

CleanShot 2023-06-02 at 16 23 11@2x
CleanShot 2023-06-02 at 16 29 45@2x
CleanShot 2023-06-02 at 16 30 41@2x
CleanShot 2023-06-02 at 16 33 33@2x
CleanShot 2023-06-02 at 16 39 00@2x
CleanShot 2023-06-02 at 16 38 03@2x

CleanShot 2023-06-02 at 16 18 26@2x
CleanShot 2023-06-02 at 16 18 05@2x
CleanShot 2023-06-02 at 16 21 03@2x
CleanShot 2023-06-02 at 16 31 30@2x

#### Ticket Link
- Fixes https://mattermost.atlassian.net/browse/MM-53004

### DIFF
--- a/webapp/src/components/channel_link_label/component.tsx
+++ b/webapp/src/components/channel_link_label/component.tsx
@@ -4,7 +4,7 @@ import {OverlayTrigger, Tooltip} from 'react-bootstrap';
 import {UserProfile} from '@mattermost/types/users';
 import {Theme} from 'mattermost-redux/types/themes';
 
-import {getUsersList} from 'src/utils';
+import {getUserDisplayName} from 'src/utils';
 import ActiveCallIcon from 'src/components/icons/active_call_icon';
 
 interface Props {
@@ -14,7 +14,7 @@ interface Props {
 }
 
 const ChannelLinkLabel = (props: Props) => {
-    const {formatMessage} = useIntl();
+    const {formatMessage, formatList} = useIntl();
 
     if (props.hasCall) {
         return (
@@ -24,7 +24,10 @@ const ChannelLinkLabel = (props: Props) => {
                     <Tooltip
                         id='call-profiles'
                     >
-                        {formatMessage({defaultMessage: '{list} {count, plural, =1 {is} other {are}} on the call'}, {count: props.profiles.length, list: getUsersList(props.profiles)})}
+                        {formatMessage({defaultMessage: '{list} {count, plural, =1 {is} other {are}} on the call'}, {
+                            count: props.profiles.length,
+                            list: formatList(props.profiles.map((user) => getUserDisplayName(user)), {type: 'conjunction'})
+                        })}
                     </Tooltip>
                 }
             >
--- a/webapp/src/utils.ts
+++ b/webapp/src/utils.ts
@@ -287,19 +287,6 @@ export function hasExperimentalFlag() {
     return window.localStorage.getItem('calls_experimental_features') === 'on';
 }
 
-export function getUsersList(profiles: UserProfile[]) {
-    if (profiles.length === 0) {
-        return '';
-    }
-    if (profiles.length === 1) {
-        return getUserDisplayName(profiles[0]);
-    }
-    const list = profiles.slice(0, -1).map((profile) => {
-        return getUserDisplayName(profile);
-    }).join(', ');
-    return list + ' and ' + getUserDisplayName(profiles[profiles.length - 1]);
-}
-
 export function playSound(name: string) {
     let src = '';
     switch (name) {
--- a/webapp/src/components/connected_profiles.tsx
+++ b/webapp/src/components/connected_profiles.tsx
@@ -1,9 +1,10 @@
 import React from 'react';
 import {OverlayTrigger, Tooltip} from 'react-bootstrap';
+import {useIntl} from 'react-intl';
 
 import {UserProfile} from '@mattermost/types/users';
 
-import {getUsersList, getUserDisplayName} from '../utils';
+import {getUserDisplayName} from '../utils';
 
 import Avatar from './avatar/avatar';
 
@@ -21,6 +22,7 @@ const ConnectedProfiles = ({pictures, profiles, maxShowedProfiles, size, fontSiz
     const diff = profiles.length - maxShowedProfiles;
 
     const showedProfiles = diff > 0 ? profiles.slice(0, maxShowedProfiles) : profiles;
+    const {formatList} = useIntl();
 
     const els = showedProfiles.map((profile, idx) => {
         return (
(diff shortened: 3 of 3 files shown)


calebroseland:
> @calebroseland Is this ready for review?

Yes, this is ready for review now 👍

cpoile:
@streamer45 Sure, whatever works best. 👍

streamer45:
@calebroseland Would you be able to look at the logic I mentioned above and see if we can apply similarly formatting?

https://github.com/mattermost/mattermost-plugin-calls/blob/c14a70a46958841a0bb7ab78075458a1dc6fa8a8/webapp/src/components/reaction_stream/reaction_stream.tsx#L61-L75

webapp/src/components/reaction_stream/reaction_stream.tsx
```diff
@@ -87,8 +72,10 @@ export const ReactionStream = () => {
                     }}
                 />
                 <span>
-                    <Bold>{participants}</Bold>
-                    <span>{untranslatable(' ')}{formatMessage({defaultMessage: 'raised a hand'})}</span>
+                    {formatMessage({defaultMessage: '{users} raised a hand'}, {
+                        count: handsup.length,
+                        users: <Bold>{formatList(handsup.map(getName), {type: 'conjunction', style: 'short'})}</Bold>,
```
cpoile, line 77: Is there an easy way to maintain the `x, y, and z` and `x, y, and 3 others` behavior using formatList? From my investigation I couldn't find it (but would be really happy if there was a way). I'm doing this in the DM ringing notification, which is much more complicated:

```
        const additionalNum = otherMembers.length - 2;
        let type: 'conjunction' | 'unit' = 'conjunction';
        if (additionalNum > 0) {
            otherMembers.splice(2);
            type = 'unit';
        }

        message = (
            <FormattedMessage
                defaultMessage={'<b>{hostName}</b> is inviting you to a call with <b>{others}' +
                    '{postscript, select, ' +
                    'true { and {additionalNum, plural, one {1 other} other {{additionalNum} others}}} ' +
                    'other {}' +
                    '}</b>'}
                values={{
                    b: (text: string) => <b>{text}</b>,
                    hostName,
                    others: formatList(otherMembers.map((u) => displayUsername(u, teammateNameDisplay)), {type}),
                    postscript: Boolean(additionalNum > 0),
                    additionalNum,
                }}
            />
        );
```

calebroseland: @cpoile apologies for not pushing it yesterday, but this is what I came up with: https://github.com/mattermost/mattermost-plugin-calls/pull/443/files#diff-2fbfe9bc31acb968fab6f96641a9664c8b5f2f444d0679e7bc558086e04f1486R62-R85

It treats the overflow part as a condensing/collapsing effect as a pre-step, where formatList then formats `N others` as a list item itself. 

I'm not 100% sure of how these semantics play out in all languages, but for 9/10+ I think it should work. I tried going down the same path as you for a time —a single formatMessage that handled the `, & N others` part within the same formatMessage, but it was really complicated for the reasons you probably experienced, too.

calebroseland: > `x, y, and z` +1 -> `x, y, and 2 others`

[`8698031` (#443)](https://github.com/mattermost/mattermost-plugin-calls/pull/443/commits/8698031790f8f1722f3e1f482fc1d9fb3a5ca94d) adds the dynamic max part

calebroseland: I also added a limit to the channel link tooltip (current draft: `10`) as that was previously unbounded

webapp/src/utils.ts
```diff
@@ -287,17 +287,11 @@ export function hasExperimentalFlag() {
...
 
-export function getUsersList(profiles: UserProfile[]) {
-    if (profiles.length === 0) {
-        return '';
-    }
-    if (profiles.length === 1) {
-        return getUserDisplayName(profiles[0]);
+export function split<T>(list: T[], i: number, pad = false): [list: T[], overflowed?: T[]] {
+    if (list.length <= i + (pad ? 1 : 0)) {
+        return [list];
```
cpoile, line 292: Could you explain the purpose of the pad? (I must be slow today, heh.)

cpoile (approved):
Looks great, thanks Caleb! I'm using this now instead of the more complicated bit I posted. It will be easier for translators too, I think.

webapp/src/utils.ts
calebroseland: Not the greatest name for that arg in all honesty, not sure what to name it exactly, but it means "pad the threshold by 1"

It allows for:
`a, b, c` (+1 to) `a, b, & 2 others`
instead of: 
`a, b, c` (+1 to) `a, b, c, & 1 other`

webapp/src/components/channel_link_label/component.tsx
```diff
@@ -13,32 +13,42 @@ interface Props {
     profiles: UserProfile[],
 }
 
+const CHANNEL_TOOLTIP_LIMIT = 10;
```
streamer45, line 16: Nice, I loved seeing this going nuts during load-tests but I suppose this is better :p 

Could we move the definition to the `constants.ts` file? Also I'd suggest to reduce the value to 8 for consistency with a couple of other things including the limit for participants in group channels. We can always sync with UX to figure out a different number if needed I suppose.

webapp/src/utils.ts
streamer45: - Would it be possible to use the literal `and` like in the other lists? Not exactly sure why it was designed that way but it seems inconsistent.
-  Also I'd probably remove the comma before that if possible because it reads a little awkward.

streamer45 (commented):
This is great :100: , left a couple of suggestions/questions.

webapp/src/utils.ts
calebroseland: 1. Done
2. The oxford comma is mostly out of our control and depends on the rules of the locale. `de` doesn't use one in this api, whereas `en-US` does use an oxford comma for this api, and `en-GB` on the other hand doesn't. While it's technically possible to remove it via `formatToParts` and reconstructing it, I would vote to let `formatList` do its thing.

In the future it'd be nice to be able to honor more granular locales better in MM for secondary formatting preferences like this—as well as date/time, units, etc.

streamer45: Thanks! Not a big deal just personally found it odd but it's borderline OCD stuff.

streamer45 (approved):
Great improvements here, thanks @calebroseland!

I'll try to sort out the e2e pipeline in the meantime.

Issue #427, opened 2023-05-10
add continous ringtone for calls


jagiella:
**Mattermost Version**
7.1.8 and 7.10 (self-hosted / docker)

**Plugin Version**
0.10.0 BETA and 0.15.1

**Describe the issue**
When starting calls with the Calls-Plugin (not Jitsi) usually nobody joins as people get a single notification which is easily missed. So that makes the whole calls feature (which is awesome by the way!!) useless as everybody has to be contacted by other means (phone) to make them aware of the meeting.

It would be nice to have a continous ringtone notifying everybody in the same channel of the active call(s) until they join or actively mute the ringtone / refuse the call. https://github.com/mattermost/mattermost-plugin-jitsi/issues/233

streamer45:
Thanks for the feedback @jagiella . Part of this is being actively worked on to address https://github.com/mattermost/mattermost-plugin-calls/issues/125

streamer45:
We are finally releasing this functionality (thanks @cpoile for the incredible contribution!). 

Please keep in mind it's an experimental feature which needs to be explicitly turned on from the plugin settings. 

![image](https://github.com/mattermost/mattermost-plugin-calls/assets/1832946/3ef4e68c-a81f-4015-a508-015c73fd6c55)

Also note that it comes with some limitations which will be addressed in the next version (e.g. mobile support).

Feel free to report any issues or leave feedback. Thank you all for your patience!

Issue #125, opened 2022-06-22
Add a popup notification when someone calls you


CP-Vub:
Currently, in the new calls (beta) in v7.0, when someone calls you, you get the same notification as you would get with a normal chat message. This does not really grab the attention of the receiver, who might think it's just a chat message notification or just miss it entirely due to the brevity of the notification, leaving the caller waiting indefinitely in an empty call.

Could an option be added to enable e.g. a popup window or a continuously ringing sound when someone calls you? This would greatly improve the usability of the call functionality. Currently, I am often still resorting back to MS Teams or skype in order to reach a colleague since they missed the call notification in mattermost.

streamer45:
Thanks @CP-Vub for the great feedback. 

We had this kind of functionality mentioned before internally and have considered adding a "ringing" feature for calls in direct message channels. Wondering if that would cover your needs or would you expect something to work across all channel types instead?

/cc @itao

CP-Vub:
Thanks @streamer45 for the quick and positive reply! 

Adding a ringing feature for calls in direct message channels would definitely already solve like 95% of all my personal use-case scenario's.  For group channels, a call typically needs to be anyway communicated well in advance in order to align the agenda's and availability of the group members so it's way less common that I want to call an entire group channel without notice. I'm not saying it wouldn't be handy to have in group channels as well but priority-wise it's lower on the ladder for me personally.

Madscientiste:
Hello, i agree with @CP-Vub 

Adding a ringing sound when someone calls you would just fix the issue for us, currently the default notification sound for when someone is calling is confusing .. 

i wouldn't mind digging into the plugin and doing a PR.

streamer45:
Thanks @Madscientiste , would definitely love a contribution. Let me just bring this up with the team before we commit to any work. I'll share an update here.

Madscientiste:
@streamer45 thanks for the quick reply ! 

just incase, i might need help in the beginning if thats possible .. i'm not very familiar with the source

streamer45:
Of course. We'd be more than happy to support you through the whole process :)

streamer45:
We discussed this today and agreed it makes sense to have an additional notification mechanism for calls in DM channels. Have created a ticket to keep track of this at https://mattermost.atlassian.net/browse/MM-46823.

As soon as we have a UX proposal we can proceed with an implementation. Will give an update when that happens.

francislavoie:
[...]

PR #445, opened 2023-06-08
MM-52857 - DM/GM ringing

cpoile:
#### Summary
- This is the first step in an epic of DM/GM ringing. Apologies for the single commit, it's all kind of related. For now, the following has been completed (also useful for a testing plan):

  - [X] user in DM receiving a call will get a call alert in the LHS.
  - [X] Multiple alerts will stack.
  - [X] if user is in a call, they will receive the call alert in a condensed form above the call widget (all cases).
  - [X] Multiple alerts will stack.
  - [X] if the user is on desktop with the global widget, the LHS call alert will also appear. Otherwise, only the condensed form about the vall widget will appear.
  - [X] dismiss should record the dismissal
    - [X] this way you don't get another ring if you restart the webapp (or log in using another client)
  - [X] dismisal array is sent with call data
    - [X] this means you won't be given a notice if you dismiss and then restart the webapp (or re-log in another client)
  - [X] notifications seen when logging in and calls are waiting
  - [X] do not ring for second call
    - [X] LHS has expanded notification open, second call arrives (condensed) -- webapp without widget, and desktop with and without global widget
    - [X] webapp widget has one condensed notification, second call arrives
  - [X] users with the global widget should have only one sound playing at once (this is testing if the global widget is causing sound, when only the LHS notification on the channels side of the desktop app should be)
  - [X] do not ring twice for same call:
    - [X] call appears in condensed above widget, then widget's call is exited (making the condensed turn into full notification) -- webapp
    - [X] call appears in lhs, then user starts unrelated call (widget appears, lhs call notification disappears, condensed call appears above widget, no call sound)
    - [X] call appears in lhs, then another call notification arrives (no sound)
    - [X] first call appears above widget (webapp), rings. Second call appears, does not ring
  - [X] should ring once for a call, only if it is the "first call" in the list
    - [X] when second call arrives, no sound. When first call is dimissed, the second call takes the first call's place and does not ring. Try this in webapp (no widget), webapp (in a call, widget), desktop (no widget), and desktop (in a call, widget)

TODO: (I will make tickets for these)
  - [ ] when incoming call is hung up, it shouldn't disappear -- it should turn into cancelled state. Needs to be manually dismissed.
  - [ ] ringing in a different server
    - [ ] should cause desktop notification, no ringing sound (?)
  - [ ] clicking on the popover should redirect the user to the dm/gm
  - [ ] badge notification when in a call (expanded view)
  - [ ] badge icon on the LHS channel name
    - [ ] badge shows even if call notification is dismissed
  - [ ] delay in notice (3 seconds)
  - [ ] if user is already on the dm/gm channel, ringing should only last 5 seconds
  - [x] dismiss should work across clients
  - [ ] animations (slide in, dismiss, halo around avatar)
  - [ ] tests

To try this out, you'll need: https://github.com/mattermost/mattermost/pull/23444
And you'll need to tag this commit with v0.17.0 before deploying.

#### Some screenshots
image
image
image
image

I think I need to get some UX input on hover states, I'll leave that for a polish stage -- this is mostly for functionality.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-52857
- Designs: https://www.figma.com/file/GjmY3V0G6K88FruOcuMP7G/MM-51513-Ringing-in-DMs%2C-GMs?type=design&node-id=918-212836&t=lFfBj1Bje7OHRohL-0

Fixes https://github.com/mattermost/mattermost-plugin-calls/issues/125
Fixes https://github.com/mattermost/mattermost-plugin-calls/issues/427


streamer45:
Unrelated but let's fix it in here, the switch call modal for DMs has a little bug due to invalid property name used:

image

webapp/src/components/incoming_calls/hooks.ts
```diff
@@ -0,0 +1,83 @@
...
+                type: HAVE_RANG_FOR_CALL,
+                data: {
+                    callUniqueID,
+                },
+            });
+        };
+
+        // If we're on the widget, that means we're on a call. If we're also on desktopWithGlobalWidget then
+        // don't ring because the ringing will be handled by the main webapp.
+        const isDesktopWithGlobalWidget = window.desktop && compareSemVer(window.desktop.version, '5.1.0') >= 0;
```
streamer45, line 65: Shouldn't this be `5.3.0`? Also I am thinking we could use the `shouldRenderDesktopWidget` util.

webapp/src/components/incoming_calls/hooks.ts
```diff
@@ -0,0 +1,83 @@
...
+        const ringHandledByWebapp = onWidget && isDesktopWithGlobalWidget;
+
+        // @ts-ignore Our mattermost import is old and at the moment un-updatable.
+        if (!call.ring || ringHandledByWebapp || haveRung || currentUser.notify_props.desktop === NotificationLevel.NONE || currentUser.notify_props.calls_desktop_sound === 'false') {
+            return;
+        }
+
+        // @ts-ignore
+        notificationSounds?.ring(currentUser.notify_props.calls_notification_sound || 'Dynamic');
+        const timer = setTimeout(() => stopRinging(), 10000);
```
streamer45, line 75: Let's make 10s a named constant.

webapp/src/index.tsx
```diff
@@ -581,8 +592,17 @@ export default class Plugin {
...
                             },
                         });
+
+                        const callExists = incomingCalls(store.getState()).findIndex((ic) => ic.callID === data[i].channel_id) >= 0;
+                        if (data[i].call && !callExists) {
+                            // This allows us time to load channels and call states
+                            setTimeout(() => {
+                                store.dispatch(incomingCallOnChannel(data[i].channel_id, data[i].call.host_id, data[i].call.start_at));
+                            }, 2000);
+                        }
```
streamer45, line 605: Not in love with this :p Could you help me understand the problem/requirements?

webapp/src/selectors.ts
```diff
@@ -125,6 +125,15 @@ export const voiceChannelCallHostChangeAt = (state: GlobalState, channelID: stri
     return pluginState(state).voiceChannelCalls[channelID]?.hostChangeAt;
 };
 
+export const voiceChannelCallDismissedNotification = (state: GlobalState, channelID: string) => {
+    const dismissed = pluginState(state).voiceChannelCalls[channelID]?.dismissedNotification;
+    if (!dismissed) {
+        return false;
+    }
+    const currentUserID = getCurrentUserId(state);
+    return (dismissed.indexOf(currentUserID) !== -1);
```
streamer45, line 134: It's 2023, could we use `includes()` :p ?

webapp/src/utils.ts
```diff
@@ -77,6 +77,15 @@ export function getChannelURL(state: GlobalState, channel: Channel, teamId: stri
     return channelURL;
 }
 
+export function shouldRenderCallsIncoming() {
+    const win = window.opener ? window.opener : window;
+    if (win.desktop && window.location.pathname.indexOf('/messages/') === -1) {
```
streamer45, line 82: Same, `includes()` would be a little easier on the brain.

webapp/src/reducers.ts
```diff
@@ -778,6 +791,49 @@ const recentlyJoinedUsers = (state: recentlyJoinedUsersState = {}, action: conne
...
+    case REMOVE_INCOMING_CALL:
+        return state.filter((ic) => ic.callID !== action.data.callID);
+    case CALL_HAS_ENDED:
+        return state.filter((ic) => ic.callID !== action.data.callID);
+    default:
+        return state;
+    }
+};
+
+type HaveRangForCallsAction = {
```
streamer45, line 818: Shouldn't this be "HaveRung" grammatically speaking?
Honestly I'd be fine with a simpler "DidRing" but clearly not a big deal.

streamer45 (commented):
This is great! Did a first pass and left mostly minor comments.

webapp/src/utils.ts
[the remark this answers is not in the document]
cpoile: Good question.
I was kind of assuming `/messages` wouldn't change... But it might... And anyway, I discovered this was wrong (you could break it under some combinations of reloading and where you were in the app).  So I've changed the `index.tsx` and the `call_container` to fix that, and I changed this util fn to do the check a bit more safely. Hopefully this will be a bit more robust for when we remove the tabs: there's a good chance that when the tabs are removed we will need to display the LHS incoming widget no matter what you're looking at, and this should work then. 
(Remember, the reason we're doing this check is that we don't want multiple LHS incoming containers to ring at the same time, which is what was happening because the desktop loaded separate webapps per tab.)

webapp/src/components/incoming_calls/hooks.ts
cpoile: Yep, my mistake, thanks.

webapp/src/components/incoming_calls/hooks.ts
cpoile: I wish there was a more organized place for us to put constants. Maybe someday (if it matters)?

webapp/src/index.tsx
cpoile: Thanks for pointing this out -- I had fixed the underlying problem and forgot to remove it. (I turned the incomingCallChannel into a thunk that would load the channel and the host if they weren't in the store, and that's what the delay was originally trying to fix.)

webapp/src/selectors.ts
cpoile: You're no fun. 😝

webapp/src/utils.ts
cpoile: At least I'm consistent...

webapp/src/components/incoming_calls/hooks.ts
streamer45: Well, I thought `constants.ts` was that place, did you forget it existed perhaps? :p

cpoile: hah... must have forgotten that :)

server/api.go
```diff
@@ -260,6 +261,49 @@ func (p *Plugin) handleEndCall(w http.ResponseWriter, r *http.Request, channelID
...
+	if err := p.kvSetAtomicChannelState(channelID, func(state *channelState) (*channelState, error) {
+		if state == nil || state.Call == nil {
+			return nil, nil
+		}
+
+		if state.Call.ID != callID {
+			return nil, fmt.Errorf("previous call has ended and new one has started")
+		}
+
+		state.Call.DismissedNotification = append(state.Call.DismissedNotification, userID)
```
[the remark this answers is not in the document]
streamer45: I was referring to that line specifically, using an array we could be appending infinite number of duplicates if the endpoint gets called again without any deduplication check. Not a problem any longer since we are using a map now :+1:

cpoile:
@streamer45 Fixed the remaining issues, I think we're good for another pass. 
There's one more remaining issue on my list: `starting a call (desktop, global widget), existing notifications don't transfer over (like in webapp).` I'm not sure if we would even need that, now that I think about it. If you have notifications, and you start a different call with the global widget, they don't disappear (like they do on webapp). So transferring them over to the global widget isn't as important. On the other hand, if you're in a different server, having the notifications pop up on the widget might still be useful. What do you think?

matthewbirtch (changes requested):
A quick first pass tonight and here are a few things I'm noticing:

1. When the team sidebar is hidden, the width of the banners is more narriw and it looks like the join/ignore buttons are not respecting the padding and the join button is even extending beyond the container somehow. One way we could combat this is to reduce the horizontal padding on these buttons to 16px instead of 20px. We could also reduce the space between buttons to 8px instead of 12px.
    image
2. The onboarding checklist floating button overlaps the new DM ringing banner. I think we can place the ringing banner at a higher z-index
    Screenshot 2023-06-14 at 9 14 17 PM

3. In the new ringing banner, can we reduce the `margin-top` on the buttons row from 16px to 12px?
    Screenshot 2023-06-14 at 8 54 07 PM
4. For the hover state on the `Ignore` button, let's go with `background: rgba(var(--button-color-rgb),0.16);`
5. If I join the DM call by clicking the top canner in the DM channel or the call card in the message area, the new banner shows as if another call is still ringing, but I'm in the call already
    image

    https://github.com/mattermost/mattermost-plugin-calls/assets/2040554/5f9fac8c-4955-43a3-a971-55276e30daf6

I still need to test the following:
- scenarios of other multiple DM calls at once
- GMs
- popup window
- desktop app

cpoile:
Thanks @matthewbirtch for the great comments -- I've made the adjustments, much much better now. 

Thinner team bar:
image

Wider team bar:
image

The z-index works (those are both above the onboarding tour button).

@streamer45 In the latest PR I migrated the ChannelCallToast component so that I could reuse the `onJoin` hook. I tested and it works, but because of the changes would you mind doing a test on it too? Thanks.

cpoile:
I'm not sure what's happening with the channel toast e2e diff -- it passes locally, and I can't see why it would be narrower in the trace.

streamer45:
> I'm not sure what's happening with the channel toast e2e diff -- it passes locally, and I can't see why it would be narrower in the trace.

The e2e pipeline is currently half broken. It doesn't even start the tests. See https://community.mattermost.com/private-core/pl/7qm1ndx8hjy9fqm7mtxbp5smkc

streamer45:
@cpoile They should be genuine failures now.

matthewbirtch:
> Thanks @matthewbirtch for the great comments -- I've made the adjustments, much much better now.

Looks great @cpoile, I was able to verify the changes are good to go. Thanks for the adjustments.

I seem to be having trouble connecting to calls now on [my test server](https://calls-matt.test.mattermost.cloud/) and keep getting this error message after it seems to time out trying to connect.

image

I also tried to use the [test server you created](https://mattermost-pr-23758.test.mattermost.cloud/), but was unable to upload the calls plugin dist through the system console due to the config it seems.

image

Sorry, this has been such a hassle for me to just test things.

cpoile:
@matthewbirtch Cloud servers are difficult to set up correctly because of networking. I think our spinwicks work because we have env var overrides that insert the correct settings (mainly the rtcd server's address), but I'm not sure how mattermost.cloud servers are set up. As for uploading a plugin to a spinwick, we'll need to either contact an SRE to enable plugin uploads (they'll need the installation address), or you'll (unfortunately) need to set up a dev environment locally. If you want to do it locally I can help, DM me. So there's a couple options, neither is quick, sorry.

matthewbirtch (changes requested):
@cpoile a few more items I noticed from my review today:

1. Do we have volume control of the ringtone? I'm wondering if we reduce the volume by 15-20% as it feels a bit too loud to me.
2. While in an active call in a channel with the popout view open, I received a DM call and the ringing sound was doubled up and really loud. If I close the popout window it is no longer doubled up, so it must be playing the sound from both windows. I noticed this in the desktop app, not sure if it also happens in browser.
3. I also found the ringing sound very distracting while in another active call. I'm now wondering if we should just use the visual banner while in an active call rather than the ringing sound. It could affect user's ability to hear the call they're actively in
4. I didn't see the banner in the popout window when a DM call comes in. The [designs had a banner that shows in the popout window](https://www.figma.com/file/GjmY3V0G6K88FruOcuMP7G/MM-51513-Ringing-in-DMs%2C-GMs?type=design&node-id=1401-279366&t=XAuOXCc7L5iwajIK-4) as well
5. When I click 'Join' the ringing doesn't immediately stop. It seems to do another half-ring. Is it possible to kill it right away when clicking the button?

cpoile:
Hey @matthewbirtch thanks for the detailed pass!
 
> 1. Do we have volume control of the ringtone? I'm wondering if we reduce the volume by 15-20% as it feels a bit too loud to me.

Yes we do, but I wonder if we need it -- the reason I ask is that the ringtone for me is actually a little soft (it's softer than the volume of my music player, for example), and I wonder if it's going to vary machine to machine. Maybe we should have a volume slidebar eventually? (That's will be another change to the webapp, unfortunately... So would have to go in post-8.0). Maybe we should get feedback from others, and if everyone sees the same volume we can set it to 0.8 or something?

> 2. While in an active call in a channel with the popout view open, I received a DM call and the ringing sound was doubled up and really loud. If I close the popout window it is no longer doubled up, so it must be playing the sound from both windows. I noticed this in the desktop app, not sure if it also happens in browser.

Nice catch, thank you. Fixed.

> 3. I also found the ringing sound very distracting while in another active call. I'm now wondering if we should just use the visual banner while in an active call rather than the ringing sound. It could affect user's ability to hear the call they're actively in

We could make that change, no problem. I'm just wondering, if the purpose of the feature is to bring the call's attention to the user (make it unmissable), maybe this is working as designed? After all, if it's annoying, they can click the dismiss button to silence the ringing?

> 4. I didn't see the banner in the popout window when a DM call comes in. The [designs had a banner that shows in the popout window](https://www.figma.com/file/GjmY3V0G6K88FruOcuMP7G/MM-51513-Ringing-in-DMs%2C-GMs?type=design&node-id=1401-279366&t=XAuOXCc7L5iwajIK-4) as well

Yep, that's on the Todo list.

> 5. When I click 'Join' the ringing doesn't immediately stop. It seems to do another half-ring. Is it possible to kill it right away when clicking the button?

Another nice catch, thank you. Fixed.

cpoile:
> Ya, I don't think we need to go as far as incorporating a volume slider. I just wanted to make sure it's not too jarring and wondered if we could reduce the volume slightly.

Maybe we should get feedback from everyone? Because for me the notification volume feels lower than my system's music volume, so I wouldn't want to make it even lower. 

> Related to this, is there a way to test the different sound options? I have a feeling the volumes will need to be normalized so that some aren't louder than others.

Yes, when you tag the PR's branch with `git tag v0.17.0` and then deploy again, you will be able to see the Calls notification settings in the settings panel:
image

> I agree the purpose is to get user's attention, but if you're already in an active call, it's a bit different IMHO. I suppose we can see what feedback we get with this as-is, but I'd like to prevent distractions that we anticipate if possible.

I'm fine either way, so I'll do what you think is best. Let me know.

> > Yep, that's on the Todo list.
> 
> Do you mean that will be in a separate PR or will come as part of this PR? Just wondering when it comes to approving the PR.

I meant the todo list in the summary -- I'm making tickets for those, for latter PRs.

> One other thing I noticed: If I have the popout window open in an active call and I have the window sized to fill the screen, and I get another DM call, the OS notification shows on top. However, clicking the OS notification doesn't seem to do anything because I am focused on the popup window. It likely just opens the DM in the background window.

I did a lot of experimenting with this just now, and I'm having trouble reproducing it consistently. It only happens with webapp (not desktop), and mostly with Chrome (but maybe once or twice with Firefox). I'm just using the DM notification for testing, since I'm assuming that's what you're clicking on. Sometimes Chrome focuses the expanded view instead of the webapp view after clicking the notification, but only if the expanded view completely covers the webapp (but sometimes it focuses the webapp even if it's completely covered by the expanded view). But sometimes that doesn't hold if you immediately opened the expanded view.
So there doesn't seem to be any consistent reproducible steps. The issue might be the `window.focus()` code in the webapp, which doesn't seem to work sometimes when the expanded view is covering... I'm not sure. I think we'll need to look into this to see if it's because the `window` of the `window.focus()` is sometimes the expanded view instead of the webapp? If that's the case, it's our fault and we can fix it. If that never happens, then it's probably a browser issue and it's out of our hands. For followup: https://mattermost.atlassian.net/browse/MM-53258

cpoile:
Here's the consensus, which I will deal with in https://mattermost.atlassian.net/browse/MM-53226:

- No ringing if you are already on a call
- No ringing if you are already getting a ring
- Ringing if you receive a second notification and the first notification has finished ringing (from figma comments)
- Ringing if the server is in the foreground or background (webapp) or if you are on that server or on a different server (desktop) (except for the above)

cpoile:
Thanks @matthewbirtch, I made a ticket for the tooltip: https://mattermost.atlassian.net/browse/MM-53320

webapp/src/actions.ts
```diff
@@ -295,3 +297,83 @@ export const displayGenericErrorModal = (title: MessageDescriptor, message: Mess
...
+
+export function incomingCallOnChannel(channelID: string, hostID: string, startAt: number) {
+    return async (dispatch: DispatchFunc, getState: GetStateFunc) => {
+        let channel = getChannel(getState(), channelID);
+        if (!channel) {
+            const res = await dispatch(loadChannel(channelID));
+            channel = res.data;
+        }
+
+        if (channel && (isDMChannel(channel) || isGMChannel(channel))) {
```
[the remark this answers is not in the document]
cpoile: Sure. It's a little less clear (compound negative tests), but ok.

webapp/src/actions.ts
```diff
@@ -295,3 +297,83 @@ export const displayGenericErrorModal = (title: MessageDescriptor, message: Mess
...
+            if (voiceChannelCallDismissedNotification(getState(), channelID)) {
+                return;
+            }
+
+            const otherUser = getUser(getState(), hostID);
+            if (!otherUser) {
+                await dispatch(getProfilesByIdsAction([hostID]));
+            }
```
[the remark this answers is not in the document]
cpoile: Maybe I made a wrong assumption? I was assuming the hostID was the person who started the call (i.e., the caller).

webapp/src/actions.ts
```diff
@@ -295,3 +297,83 @@ export const displayGenericErrorModal = (title: MessageDescriptor, message: Mess
...
+
+            // if this is the first call in the list, it should ring.
+            const calls = incomingCalls(getState());
+            const ring = calls.length === 0;
```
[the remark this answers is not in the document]
cpoile: With the next PR (the one that handles second call ringing), the rule was designed: no overlapping ringing, so if a call came in while a previous one was ringing, the new call will not ring. I asked Abhijit to clarify if this means the second will _never_ ring.

webapp/src/actions.ts
```diff
@@ -295,3 +297,83 @@ export const displayGenericErrorModal = (title: MessageDescriptor, message: Mess
...
+    };
+};
+
+export const dismissIncomingCallNotification = (callID: string) => {
+    return async (dispatch: DispatchFunc) => {
+        Client4.doFetch(
+            `${getPluginPath()}/calls/${callID}/dismiss-notification`,
+            {method: 'post'},
+        );
+        await dispatch(removeIncomingCallNotification(callID));
```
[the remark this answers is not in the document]
cpoile: Good question. Nope, I specifically didn't wait on it so that making the notification and ringing go away wasn't delayed. I added a catch just to clean up a bit better.

webapp/src/components/channel_call_toast.tsx
```diff
@@ -0,0 +1,130 @@
...
+    const [pictures, setPictures] = useState<string[]>([]);
+
+    const [, onJoin] = useDismissJoin(currChannelID);
+
+    useMemo(() => {
+        const thePictures = [];
+        if (currChannelID !== connectedID && connectedUsers) {
+            if (connectedUsers.length > 0 && profiles.length === connectedUsers.length) {
+                for (let i = 0; i < connectedUsers.length; i++) {
+                    thePictures.push(Client4.getProfilePictureUrl(profiles[i].id, profiles[i].last_picture_update));
```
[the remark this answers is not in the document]
cpoile: Just the URL; looks like it's just a utility function.

webapp/src/components/incoming_calls/call_container.tsx
```diff
@@ -0,0 +1,51 @@
...
+    const onACallWithoutGlobalWidget = useOnACallWithoutGlobalWidget();
+
+    if (!shouldRenderCallsIncoming() || callsIncoming.length === 0 || onACallWithoutGlobalWidget) {
+        // don't show incoming calls if we're on a call without the global widget because
+        // we'll see the notification above the widget
+        return null;
+    }
+
+    const wider = myTeams?.length > 1;
+    const calls = [...callsIncoming].sort((a, b) => b.startAt - a.startAt);
```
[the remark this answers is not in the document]
cpoile: Sure, good idea. I made a separate memoized selector for that.

webapp/src/components/incoming_calls/hooks.ts
```diff
@@ -0,0 +1,99 @@
...
+
+export const useOnACallWithoutGlobalWidget = () => {
+    const connectedChannel = useSelector(connectedChannelID);
+    return Boolean(connectedChannel && !shouldRenderDesktopWidget());
+};
+
+export const useRinging = (call: IncomingCallNotification, onWidget: boolean) => {
+    const dispatch = useDispatch();
+    const currentUser = useSelector(getCurrentUser);
+    const callUniqueID = `${call.callID}${call.startAt}`;
```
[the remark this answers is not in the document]
cpoile: Ok, made the change. This was bigger than I thought it would be. And there are still some confusing usages of channelID <-> callID (recording, sending messages to the desktop).

webapp/src/components/incoming_calls/hooks.ts
```diff
@@ -0,0 +1,99 @@
...
+        // don't ring because the ringing will be handled by the main webapp.
+        const ringHandledByWebapp = onWidget && shouldRenderDesktopWidget();
+
+        // @ts-ignore Our mattermost import is old and at the moment un-updatable.
+        if (!call.ring || ringHandledByWebapp || didRing || currentUser.notify_props.desktop === NotificationLevel.NONE || currentUser.notify_props.calls_desktop_sound === 'false') {
+            return;
+        }
+
+        // @ts-ignore
+        notificationSounds?.ring(currentUser.notify_props.calls_notification_sound || 'Dynamic');
```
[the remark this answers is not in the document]
cpoile: That was an assumption, based on the ordering in the designs. Matt clarified that we should use the Calm sound, so made that change (and turned it into a global).

webapp/src/actions.ts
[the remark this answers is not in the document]
streamer45: I think it's just the name that's confusing. Calling it `otherUser` when we mean the caller. In the context of a DM it's fine I suppose but in a GM it feels weird. No problem with the logic, although technically the host could change so it's not always an exact mapping. What you really want would be what we set as `OwnerID`, not sure if we expose it though but not a big deal.

webapp/src/components/incoming_calls/hooks.ts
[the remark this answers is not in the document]
streamer45: That's okay, as long as they are used in a way that doesn't cause conflicts we should be good.

webapp/src/components/channel_call_toast.tsx
```diff
@@ -0,0 +1,134 @@
...
+import ActiveCallIcon from 'src/components/icons/active_call_icon';
+import {useDismissJoin} from 'src/components/incoming_calls/hooks';
+import Timestamp from 'src/components/timestamp';
+import {
+    connectedChannelID, isLimitRestricted,
+    voiceChannelCallInCurrentChannel, voiceConnectedCurrentChannel, voiceProfilesInCurrentChannel,
+} from 'src/selectors';
+import {callStartedTimestampFn} from 'src/utils';
+
+const ChannelCallToast = () => {
```
streamer45, line 18: Not sure if it's a regression but I noticed that if you click to close the toast and a new call starts on the same channel you don't get it back, although my expectation is that you should.

streamer45 (approved):
Thanks @cpoile , amazing effort :bow:

matthewbirtch (approved):
Thanks Chris! I think this looks good to me!

webapp/src/components/channel_call_toast.tsx
cpoile: Looks like it's that way on main too -- you have to change channels for it to come back. I made a ticket for it, and I'll take care of it: https://mattermost.atlassian.net/browse/MM-53322

webapp/src/actions.ts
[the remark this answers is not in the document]
cpoile: I agree -- I fixed it by moving over to ownerID and then referring to them as the Caller from then on.

PR #447, opened 2023-06-12
Expose new settings for TCP and IPv6

streamer45:
#### Summary

Adding config settings to cover the new functionality, to configure the TCP listener(s) and to enable experimental IPv6 support. 

I've created a follow up ticket to see if we can perhaps simplify a bit the and hide more advanced settings ([MM-53139](https://mattermost.atlassian.net/browse/MM-53139)).

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-50294
https://mattermost.atlassian.net/browse/MM-52268


cpoile (approved):
Looks good, thanks!

cpoile:
@streamer45 I forgot to ask -- is there an easy way to turn off TCP support?

streamer45:
> @streamer45 I forgot to ask -- is there an easy way to turn off TCP support?

@cpoile Nope. The reasoning to make it a default was:

- Candidates are generated in additions to the others but UDP ones are always preferred if connectivity is possible so it won't add any extra cost in most cases.
- In practice it's easy enough to prevent TCP from working by simply configuring the networking. Usually you'd have to explicitly turn it on anyway through some firewall rules.

PR #448, opened 2023-06-12
Replace Boards e2e test with Playbooks

streamer45:
#### Summary

Boards product has been officially disabled so e2e tests are failing now. I am replacing the same test case with Playbooks so we still test switching products.


streamer45:
Github jobs getting randomly cancelled today :suspect:

cpoile (approved):

PR #452, opened 2023-06-23
MM-53226 - Desktop notifications, visibility rules, ringing rules

cpoile:
#### Summary
  - [X] MM-53226 Notification and visibility
    - [X] when webapp is not updated for notifyMe, it will not crash or error, we just won't get a notification
    - [X] ringing in a different server (in desktop) should cause desktop notification
    - [X] ringing in webapp when the window is "hidden" (on a different workspace, fully behind another window, etc.) will cause a notification -- but only when the user has enabled desktop notification in their browser
    - [X] No ringing if you are already on a call
    - [X] No ringing if you are already getting a ring
    - [X] Ringing if you receive a second notification and the first notification has finished ringing
    - [X] Clarification of the ringing rules: Ringing if the server is in the foreground or background (webapp) or if you are on that server or on a different server (desktop) (except for the above)
    - [X] Stop ringing immediately once user joins any call (even if not the one that was ringing)
    - [X] enable ringing flag

- ~~NOTE: the notification is a little odd, in that if the user has notifications for a DM channel turned on (default), they will get two -- one for the DM message marking the start of the call, and second our custom desktop alert for the call. It's better for GMs, since those messages do not default to giving alerts (to my constant annoyance)~~
- Notification is now turned off in DM channels for <8.1. It is turned on for GM channels.
- Notifications and ringing respect the user's channel and status settings, and will not notify or ring when:
  - the user has muted the channel
  - the user is DND
  - the user is OOO

More test cases:

  - [X] for the below, the visual notification in the product should always show.
    - [X] user's system notification preferences, everything default, `Notification sound for incoming calls` off: no ringing, (when server is hidden) yes desktop notification and yes desktop notification sound
    - [X] user's system notification preferences, `Notification sound` off: ringing yes (if `Notification sound for incoming calls` is on, else no), no desktop notification sound (but should receive desktop notification)
    - [X] user's system notification preferences, `Send desktop notification` never: no ringing, no desktop notification sound, no desktop notification
    - [X] user's channel notification preferences, system notification preferences on default, channel preference `Notification sound` off: ringing, no desktop notification sound (but should receive desktop notification)
    - [X] user's channel notification preferences, system notification preferences on default, channel preference `Send desktop notification` never: ringing (if user's system preference `Notification sound for incoming calls` is on, else no), no desktop notification sound, no desktop notfication
    - [X] mute channel should mute all sound and notifications from channel

Notifications look like:
For DMs:
image

For GMs:
image

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-53226


webapp/src/index.tsx
```diff
@@ -615,17 +616,19 @@ export default class Plugin {
...
 
-                        const callExists = incomingCalls(store.getState()).findIndex((ic) => ic.callID === data[i].channel_id) >= 0;
-                        if (data[i].call && !callExists) {
-                            // dismissedNotification is populated after the actions array has been batched, so manually check:
-                            const dismissed = data[i].call?.dismissed_notification;
-                            if (dismissed) {
-                                const currentUserID = getCurrentUserId(store.getState());
-                                if (Object.hasOwn(dismissed, currentUserID) && dismissed[currentUserID]) {
-                                    continue;
+                        if (ringingEnabled(store.getState())) {
```
cpoile, line 619: I realize we can continue early here, but I think that would be more confusing, and prone to mistakes later. Here (and elsewhere) we're just saying: if ringing is enabled, do the following).

webapp/src/websocket_handlers.ts
```diff
@@ -110,7 +114,7 @@ export function handleCallStart(store: Store, ev: WebSocketMessage<CallStartData
         if (channel) {
             followThread(store, channel.id, channel.team_id);
         }
-    } else {
+    } else if (ringingEnabled(store.getState())) {
```
cpoile, line 117: This is ugly, but the linting rules won't let me do a lone if inside the else.

streamer45 (approved):
Looks good!

For my own sanity could you help me understand if this more complex flow is expected:

Given three users, A, B, C:

- A calls B
- Ringing on B's side starts
- C calls B
- B dismisses A's call notification
- Ringing on B's side stops
- A ends the call to B
- A calls B once more (starts a new call)
- Ringing on B's side starts again, this time from the stacked call since the "oldest" one (from C) is still on-going and "promoted" to the expanded form.

cpoile:
@streamer45 Thanks! Yep, that's how it's supposed to work -- the rule is simply that a call that arrives while another call is ringing will never ring. But a call that arrives while nothing is ringing will ring.

matthewbirtch:
> NOTE: the notification is a little odd, in that if the user has notifications for a DM channel turned on (default), they will get two -- one for the DM message marking the start of the call, and second our custom desktop alert for the call. It's better for GMs, since those messages do not default to giving alerts (to my constant annoyance)

What can we do about this? It's definitely weird and undesirable. And as an FYI, there is an initiative to change the GM notification behavior to align with DMs, so that may become an issue down the road for GMs too.

I don't seem to get the 'is inviting you' version of the notification in the desktop app. this is what I see:
image


> * ringing in a different server (in desktop) should cause desktop notification
> * [x]  ringing in webapp when the window is "hidden" (on a different workspace, fully behind another window, etc.) will cause a notification -- but only when the user has enabled desktop notification in their browser

I couldn't test out the different server ringing since I can only get my local server running, but I was able to test the other items in the list and they seem to work as outlined.

cpoile:
> What can we do about this? It's definitely weird and undesirable. And as an FYI, there is an initiative to change the GM notification behavior to align with DMs, so that may become an issue down the road for GMs too.

We would have to modify the webapp code and add a rule to not show the desktop notification in this case. So, the rule would be: do not show the default notification if it is a call started post, the user is running v0.17.0 of calls, ringing is enabled globally, and ringing is enabled in their notification settings. I think that should cover it. This would go into MM v8.0, so we might want to turn off desktop notifications until MM v8.1, or else live with the double notification for DMs for one version.

> I don't seem to get the 'is inviting you' version of the notification in the desktop app. this is what I see: image

You need to be running the most recent version of server master, which exposes the notification function.

> > * [x]   ringing in webapp when the window is "hidden" (on a different workspace, fully behind another window, etc.) will cause a notification -- but only when the user has enabled desktop notification in their browser
> 
> I couldn't test out the different server ringing since I can only get my local server running, but I was able to test the other items in the list and they seem to work as outlined.

Thanks!

matthewbirtch:
> We would have to modify the server webapp code and add a rule to not show the desktop notification in this case. So, the rule would be: do not show the default notification if it is a call started post, in a DM or GM channel, the user is running v0.17.0 of calls, ringing is enabled globally, and ringing is enabled in their notification settings. I think that should cover it. This would go into MM v8.1, so we might want to turn off desktop notifications until MM v8.1, or else live with the double notification for DMs for one version (MM v8.0).

As discussed in our DM on Mattermost, Let's avoid the doubled-up notification for DMs and disable the new desktop notification for DMs until this can be resolved in 8.1

matthewbirtch (approved):
Approving assuming we can resolve the double notification for DMs as discussed

cpoile:
@matthewbirtch @streamer45 Updated the summary with the notification rule changes (for later testing), latest commit implements them.

cpoile:
@matthewbirtch Updated the summary with more test cases, implemented by the latest commit. 
@streamer45 Rebased https://github.com/mattermost/mattermost-plugin-calls/pull/455 so it's up to date if you want to make another cut (but you don't have to, these are edge cases for combinations of notifications settings, it't not crucial for the basic functionality).

matthewbirtch:
> @matthewbirtch Updated the summary with more test cases, implemented by the latest commit. @streamer45 Rebased #455 so it's up to date if you want to make another cut (but you don't have to, these are edge cases for combinations of notifications settings, it't not crucial for the basic functionality).

Verified that the muted/DND/OOO settings are having the desired effect and working as outlined and I also verified the preferences settings cases outlines

cpoile:
@matthewbirtch Thanks, also fixed the default ringing issue. Now it will default to on if the user hasn't set it (but it will still respect the other muted/OOO/etc. settings.

PR #455, opened 2023-06-26
MM-53322 - Channel toast fixes

cpoile:
#### Summary
- This fixes some channel toast issues: 
1. when you dismiss the toast and (without changing channel) the call ends and a new call begins, the toast would not reappear.
2. when you dismiss the toast and move to another channel then back, the toast would reappear.

I'm introducing a dimissedCalls reducer to record the dismiss, using the callID. We might need this later if we start having more complicated logic for regular channel notifications.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-53322


streamer45 (approved):
Nice, thanks! Something to consider a couple of e2e tests for, should be simple enough.

PR #461, opened 2023-06-28
MM-53261 - Expanded view notifications

cpoile:
#### Summary
- Expanded view notifications

- I tweaked the css on the condensed notification (the one that appears above the global widget and above the first notification on the LHS of the webapp and desktop) to be closer to the designs (slightly larger icons, better spacing -- but the spacing on the lhs of the notification, around the avatar and lhs of the message, is a little different than the designs -- those I think are the tweaks Matt made in an earlier PR so I didn't adjust those):
image

- And here is the expanded view notification:
image

A couple notes:
- Clicking join on the webapp brings up the switch call modal within the expanded view, and if you press leave&join then the expanded view is closed and you join the new call. Clicking join on the desktop focuses the switch call modal within the desktop's channels view, and if you press leave&join then the expanded view is closed and you join the new call. Little bit of a difference, due to the way communication works between the expanded view and the webapp vs. desktop.
- The expanded notifications will force the width of the expanded view to expand, which will cause the participant RHS list to be hidden if there isn't enough space, and will force the hangup icon off the right of the view. This will happen on very small screens. I'm not sure of what we want to do in this situation because it's kind of an edge case: you would have to be on a small screen (so small that the rest of the rhs panels would be kind of useless and wonky in the first place) and you would need to receive 2 or more incoming notifications at the same time. If it does happen, you can always close the notifications to bring the hang up icon back on screen. I'll leave it up to UX to decide if it's something we want to spend time fixing it, but the solution is not straightforward (I've tried expanding past the left and hiding the overflow, but it's a bit tricky and doesn't look great at all). Maybe there's another solution other than overflow to the left?

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-53261

### DIFF
--- a/webapp/src/components/call_widget/component.tsx
+++ b/webapp/src/components/call_widget/component.tsx
@@ -1602,7 +1602,6 @@ export default class CallWidget extends React.PureComponent<Props, State> {
                         key={c.callID}
                         call={c}
                         onWidget={true}
-                        global={Boolean(this.props.global)}
                     />
                 ))}
             </div>
--- a/webapp/src/components/expanded_view/component.tsx
+++ b/webapp/src/components/expanded_view/component.tsx
@@ -7,6 +7,7 @@ import React from 'react';
 import {IntlShape} from 'react-intl';
 import {RouteComponentProps} from 'react-router-dom';
 import {compareSemVer} from 'semver-parser';
+
 import styled, {createGlobalStyle, css} from 'styled-components';
 import {MediaControlBar, MediaController, MediaFullscreenButton} from 'media-chrome/dist/react';
 
@@ -20,6 +21,8 @@ import {
     UserState,
 } from '@calls/common/lib/types';
 import {mosThreshold} from '@calls/common';
+
+import {ExpandedCallContainer} from 'src/components/incoming_calls/expanded_call_container';
 
 import {Emoji} from 'src/components/emoji/emoji';
 
@@ -180,6 +183,9 @@ export default class ExpandedView extends React.PureComponent<Props, State> {
                 flexDirection: 'column',
                 flex: '1',
             },
+            headerSpreader: {
+                marginRight: 'auto',
+            },
             closeViewButton: {
                 display: 'flex',
                 alignItems: 'center',
@@ -187,7 +193,6 @@ export default class ExpandedView extends React.PureComponent<Props, State> {
                 width: '40px',
                 height: '40px',
                 borderRadius: '4px',
-                marginLeft: 'auto',
             },
             participants: {
                 display: 'grid',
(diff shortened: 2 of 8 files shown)


streamer45 (approved):
Nice work!

Slightly unrelated to the specifics of this PR but I was wondering whether we should retain the notification in case the user cancels on the switch modal.

image

cpoile:
@streamer45  (I wrote a lengthy explanation here, but I think it would be better to explain it offline next meeting.) In short, it's doable, but would add complication. We would have to decide if it's worth it.

matthewbirtch:
@cpoile I'm wondering if we should consider stacking vertically for the case where multiple notifications come in at once. Like this:
image

cpoile:
@matthewbirtch Yep, that's doable, thanks!

What do you think of Claudio's suggestion in https://github.com/mattermost/mattermost-plugin-calls/pull/461#pullrequestreview-1505782711 to not dismiss the notification banner if you press `Cancel` in the switch call modal? (in other words: Keep the notification banner (not the sound) around until the user presses the `Leave and join new call` button)

cpoile:
@matthewbirtch I've assumed that we want the width of all the notifications to be the same (even if the name of an individual notification doesn't take the full width):

image

matthewbirtch:
> @matthewbirtch I've assumed that we want the width of all the notifications to be the same (even if the name of an individual notification doesn't take the full width):
> 
> image

@cpoile yep, keep them a fixed width I think

matthewbirtch:
> What do you think of Claudio's suggestion in [#461 (review)](https://github.com/mattermost/mattermost-plugin-calls/pull/461#pullrequestreview-1505782711) to not dismiss the notification banner if you press `Cancel` in the switch call modal? (in other words: Keep the notification banner (not the sound) around until the user presses the `Leave and join new call` button)

I think this makes sense, with one addition. 

Keep the notification banner (not the sound) around until the user presses the `Leave and join new call` button **OR if they have pressed the `x` button on the banner to dismiss it.**

DHaussermann:
Tested and passed
- I have run through the tests outlined by @cpoile in https://github.com/mattermost/mattermost-plugin-calls/pull/452 and https://github.com/mattermost/mattermost-plugin-calls/pull/445 
- Reviewed the implementation of notifications in expand view for this PR 
- Re-tested configuration setting and ensured they are respected as well as made sure there is no UI impact when the ringing Beta feature is disabled 
- Completed some exploratory testing as well as ensured the feature is working as expected in desktop and multiple browsers.
- There was an issue found with the state of the data in the LHS after the login process but it is not specific to the rigging feature this was captured here https://mattermost.atlassian.net/browse/MM-53512

matthewbirtch (approved):
LGTM!

PR #462, opened 2023-06-28
[MM-53433] Use JobID to differentiate bot initiated sessions

streamer45:
#### Summary

With the introduction of new job types (e.g. transcriptions) we need a way to properly identify bot initiated sessions. So far we assumed all bot connections were recording related but that won't always be the case going forward. The idea is to pass a ~`ContextID`~ `JobID` parameter that univocally identifies a resource/job type.

#### Related PR

https://github.com/mattermost/calls-recorder/pull/38

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-53433


webapp/src/types/types.ts
```diff
@@ -22,6 +22,17 @@ export type ChannelState = {
...
+export type CallsClientJoinData = {
+    channelID: string;
+    title?: string;
+    threadID?: string;
+
+    // Calls bot only
+    // contextID is the id used to track the context of the bot connection to
+    // a call (e.g. recording, transcription).
+    contextID?: string;
+}
```
streamer45, line 34: I think this could fit in https://github.com/mattermost/calls-common since it's part of the public API. For this PR though it's practically unnecessary.

cpoile (approved):

streamer45:
@cpoile Heads up, I applied the changes as suggested in https://github.com/mattermost/rtcd/pull/115#discussion_r1298720356

PR #465, opened 2023-07-06
[MM-53508] Fix potentially empty profiles in expanded view

streamer45:
#### Summary

We had this strange issue where profiles would get emptied out almost randomly when loading the Calls expanded view:

image

image

Today I was finally able to hunt this down to the `prefetchThread` action that internally calls `getThread`:

```js
getThread(userId: string, teamId: string, threadId: string, extended = true)
```

We were setting the last parameter (`extended`) based on whether CRT was enabled or not. This had the unfortunate effect of causing an almost empty profile to get loaded into the state. It's still unclear why we'd ever want to do that but hopefully someone from webapp can answer that.

@calebroseland Adding you as a sanity check since you originally added this code and if you have any additional thoughts on the above.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-53508

### DIFF
--- a/webapp/src/actions.ts
+++ b/webapp/src/actions.ts
@@ -9,7 +9,6 @@ import {getCurrentUserId, isCurrentUserSystemAdmin} from 'mattermost-redux/selec
 import {getConfig} from 'mattermost-redux/selectors/entities/general';
 import {getThread as fetchThread} from 'mattermost-redux/actions/threads';
 import {getThread} from 'mattermost-redux/selectors/entities/threads';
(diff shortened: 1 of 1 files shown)


cpoile (approved):
Awesome bug hunting, thanks Claudio!

calebroseland (approved):
@streamer45 there was historical context to calling `extended` with whether CRT was enabled or not, but all the other usages of `fetchThread` are also hardcoded to true now, so this should be the appropriate fix. Thanks for finding it!

PR #468, opened 2023-07-12
MM-53340 - Suppress desktop notifications for `Call Started` posts, MM >= v8.1

cpoile:
#### Summary
- This PR uses the desktopNotificationsHook introduced in https://github.com/mattermost/mattermost/pull/23884 to supress desktop notifications for DM and GM channels for `call started` posts only.

@DHaussermann Test plan:
- The built in desktop notification for a post in a DM/GM channel should appear if:
  - Calls ringing is set to false in the system console settings
  - Calls ringing is set to true but MM is version < v8.1
  - The post is any other type other than the `call started` post (so, just like normal). In other words: this feature ONLY affects the `calls started` post types.
- The new calls desktop notification will appear if:
  - Mattermost server is >= v8.1 (you will have to manually set the version tag on the server to at least 8.1), a call is started in a DM/GM channel, calls ringing is enabled
- If the new calls desktop notification appears, the old regular message notification will NOT appear.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-53340


webapp/src/desktop_notifications.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export function desktopNotificationHandler(store: Store, post: Post, channel: Channel, args: DesktopNotificationArgs): {error?: string, args?: DesktopNotificationArgs} {
+    if (args.notify) {
+        // Calls will notify if:
+        //  1. it's a custom_calls post (call has started)
+        //  2. in a DM or GM channel
+        //  3. calls is enabled and is v0.17.0+ (it is if this is running)
+        //  4. calls ringing is enabled on the server
+
+        // @ts-ignore our imported webapp types are old
```
streamer45, line 17: What would it take to update them?

webapp/src/desktop_notifications.ts
```diff
@@ -0,0 +1,26 @@
...
+
+import {ringingEnabled} from 'src/selectors';
+import {isDmGmChannel} from 'src/utils';
+
+export function desktopNotificationHandler(store: Store, post: Post, channel: Channel, args: DesktopNotificationArgs): {error?: string, args?: DesktopNotificationArgs} {
+    if (args.notify) {
+        // Calls will notify if:
+        //  1. it's a custom_calls post (call has started)
+        //  2. in a DM or GM channel
+        //  3. calls is enabled and is v0.17.0+ (it is if this is running)
```
streamer45, line 14: Would the version be `>= v0.18.0`? At least for the new notification. Also I suppose there's the extra requirement of `MM >= v8.1`

webapp/src/index.tsx
```diff
@@ -281,6 +282,10 @@ export default class Plugin {
             return slashCommandsHandler(store, joinCall, message, args);
         });
 
+        registry.registerDesktopNotificationHook?.(async (post, msgProps, channel, teamId, args) => {
+            return desktopNotificationHandler(store, post, channel, args);
+        });
```
streamer45, line 287: I assume it's not a problem but just confirming that this won't have any effect on unsupported MM versions.

streamer45 (approved):
Thanks :+1:

webapp/src/desktop_notifications.ts
cpoile: We'd have to rearrange webapp and export types correctly in an npm package. I've talked to Harrison about it, probably 6 months ago, and he said it was something he definitely wants to do. But given recent org changes, I'm not sure if it's on the roadmap. I could take that on, if we can fit it in between a feature?

streamer45: Thanks. Not urgent, I was mostly looking to understand what kind of effort was needed and potentially get a ticket open.

webapp/src/desktop_notifications.ts
cpoile: Ah, right, thanks, fixed!

webapp/src/index.tsx
cpoile: 👍 The `?.` prevents a crash, and nothing will happen if this is run on an unsupported version.

webapp/src/desktop_notifications.ts
cpoile: A ticket to estimate the effort: https://mattermost.atlassian.net/browse/MM-53667

DHaussermann:
@cpoile I have done a round of testing on this and the business logic works as expected. 

There is one thing that may not be working as expected. There is a section about the "new calls desktop notification" I have looked for this with the ringing feature on and off and I'm not sure what you mean.
Setup:
- Created a spin-up of  v`8.1.0-rc1`
- Deployed the plugin build from this branch

When ringing feature is off I get a desktop notification as expected but it looks no different to me.
image
When ringing feature is on I don't get a desktop notification as expected and the LHS notification looks normal.

Can you clarify this for me?

cpoile:
@DHaussermann That notification you're seeing is the default DM notification -- with ringing off, we aren't intercepting any notifications, so we should expect to see the default one (as you see). Hope that helps?

cpoile:
@DHaussermann Not sure if I answered your question above or not, let me know, thanks!

cpoile:
I double checked the case Dylan mentioned and it's working as designed, so I'm going to merge for the release cut today. Thanks Dylan for the thorough work!

PR #469, opened 2023-07-12
MM-53320 - Tooltip for the incoming condensed notification message

cpoile:
#### Summary
- Tooltip shows when the message text is ellipsized (ellipsised? ellipsis...ized.)
- @matthewbirtch : I tweaked the message to make this a bit more consistent. Before: `Call from Tasha`, now: `Call from Tasha with Andre...` with the full message on hover. I did this so that we get the full list of names on hover, since we have that option now. Let me know if you'd rather we go back to the original.

#### Screenshots
image
image

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-53320


matthewbirtch:
Looks good to me @cpoile. One quick question about this screenshot:

image

Is there a reason we would have `and 1 other` in the tooltip instead of just showing the full list of names?

webapp/src/components/incoming_calls/call_incoming_condensed.tsx
```diff
@@ -27,23 +27,36 @@ type Props = {
...
+    const messageRef = useRef<HTMLDivElement>(null);
     const caller = useSelector((state: GlobalState) => getUser(state, call.callerID));
+    const [callerName, others] = useGetCallerNameAndOthers(call, 2);
+    const [showTooltip, setShowTooltip] = useState(false);
 
     useRingingAndNotification(call, onWidget);
     const [onDismiss, onJoin] = useDismissJoin(call.channelID, call.callID);
 
-    const callerName = displayUsername(caller, teammateNameDisplay, false);
+    useLayoutEffect(() => {
```
streamer45, line 38: What's the reasoning behind using `useLayoutEffect` vs `useEffect` in this case?

cpoile: I defaulted to that because it's a layout calculation, but you're right -- there's no chance that we'll see a rerender, so it's fine to use useEffect.

cpoile:
> Is there a reason we would have `and 1 other` in the tooltip instead of just showing the full list of names?

@matthewbirtch Nope, no reason -- I changed it to list up to 10 names, then elide the rest. Does that seem good?

matthewbirtch:
> @matthewbirtch Nope, no reason -- I changed it to list up to 10 names, then elide the rest. Does that seem good?

I guess considering that GMs are limited to 8 (including yourself) we wouldn't really run in to that limit

matthewbirtch (approved):
LGTM!

streamer45 (approved):
Looks good :+1:

cpoile:
@DHaussermann Ready for you, thanks!

DHaussermann (approved):
Tested and passed 
- Tooltip is functional
- Tooltip can accommodate all 6 additional members of a GM
- Name display in tooltip respects user display settings (not new to this PR - just checking) 
- No change to existing theming for tooltip

LGTM!

PR #473, opened 2023-07-13
MM-53260 - Clicking notification redirects to channel

cpoile:
#### Summary
- Clicking the incoming call notification and condensed notification will redirect the user to the channel and stop any ringing (whether from that call or another -- it assumes that interacting with the notification indicates that the user sees the notification, so we can stop the ringing).

~~/cc @matthewbirtch for visibility, you can try it out if you want and optionally review, but it's pretty straightforward~~ You might want to review, as I touched the UI of the condensed incoming notification (should still look the same):
image
image

@DHaussermann Testing involves a bunch of combinations:
- The regular incoming call notification in webapp (the bigger one on LHS)
- The condensed incoming call notification in webapp on LHS, as well as above the global widget, and in the expandedview on desktop. (on webapp, clicking the condensed notification in the expanded view will change the channel in the webapp, but as far as I know we cannot force a browser to focus, for security reasons -- @streamer45 maybe you know of a way?).
- The desktop global widget: when the desktop's call is started in regular channels, DM channels, and GM channels, and those three in combination with currently viewing those three channel types. I encountered a weird bug where the widget did not have a currentTeam when it was viewing a regular channel, so clicking the condensed incoming call notification above the global widget caused a crash. To test for this, try all the combinations to make sure clicking the condensed incoming notification sends the user to the right channel (and focuses, when using desktop). Thanks.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-53260


cpoile:
> But I'd argue that if this doesn't work reliably, we shouldn't be allowing the element to be clicked. It causes a side effect on some other window while nothing happens on the currently focused one which seems odd and likely against any web ux/accessibility common sense.

That does indeed work for Chrome, Opera, Firefox -- amazing (I was missing the `open` call). But unfortunately it doesn't work on Safari. So I'm going to disable clicking on the condensed notification in the expanded view for webapp.

One other thing: the way the global widget handles the cursor is a little buggy with the condensed notifications -- sometimes it changes the cursor to a pointer, but sometimes it doesn't. But otherwise everything works correctly. If we could fix it, that would be great, but I've tried setting the cursor explicitly, and setting the inner button's cursor, but there seems to be something different about that area of the global widget. Even the `join` and `x` buttons sometimes don't change their cursors, but at least they're a bit more consistent than the outer container button.

And the reason the condensed notification uses a button but the normal LHS notification doesn't, is that the global widget doesn't seem to run the onClick handler for regular divs. But the function is identical, so that's just an FYI.

streamer45:
> That does indeed work for Chrome, Opera, Firefox -- amazing (I was missing the open call). But unfortunately it doesn't work on Safari. So I'm going to disable clicking on the condensed notification in the expanded view for webapp.

I didn't think you felt that strong about Safari support :p

> One other thing: the way the global widget handles the cursor is a little buggy with the condensed notifications -- sometimes it changes the cursor to a pointer, but sometimes it doesn't.

Yeah, widget buttons suffer from the same sort of inconsistent behaviour. I suspect it's an Electron on Mac thing because on Linux it behaves as expected. May even be a MAS issue, haven't tested.

webapp/src/components/incoming_calls/call_incoming_condensed.tsx
```diff
@@ -30,8 +29,12 @@ export const CallIncomingCondensed = ({call, onWidget = false, joinButtonBorder
     const teammateNameDisplay = useSelector(getTeammateNameDisplaySetting);
     const caller = useSelector((state: GlobalState) => getUser(state, call.callerID));
 
-    useRingingAndNotification(call, onWidget);
     const [onDismiss, onJoin] = useDismissJoin(call.channelID, call.callID);
+    const onContainerClick = useOnChannelLinkClick(call);
+    useRingingAndNotification(call, onWidget);
+
+    // Do not allow clicks in the webapp expanded view because Safari does not let us switch and focus parent.
+    const onWebappExpanded = !isDesktopApp() && window.location.pathname.includes(`${pluginId}/expanded/`);
```
streamer45, line 37: To note, I am not opposed to enable the functionality as designed except on Safari. But fine either way.

streamer45 (approved):

webapp/src/components/incoming_calls/call_incoming_condensed.tsx
cpoile: @matthewbirtch Let me know if you're okay shipping the clicking feature on everything except Safari? (I will turn off clicking if the user is using Safari)

DHaussermann (approved):
Tested and passed
- Moved to the DM/GM where the call is happening from various other channels including Threads view 
- No issues from various team or when RHS is open 
- No issues with trying to select the notification when modal are open that are not full screen as modals are masked and must be dismissed first
- Confirmed the collapsed version of the notification (shown when you're already on a call) also has the redirect function
- No applicable cases when on sys admin side
- No applicable cases from notification that show inside the expanded view of another is ongoing
- Tested with multiple browsers 
- No change to clicking the dismiss button
- No change to joining the call with the button which does not cause redirection

LGTM!

webapp/src/components/incoming_calls/call_incoming_condensed.tsx
cpoile: @matthewbirtch re-ping (might have gotten lost over the weekend)

cpoile: Let's merge this now as-is since it's been tested. If we feel it's useful to do what we've talked about above please make a ticket @matthewbirtch , thanks.

matthewbirtch: Sorry for the delay folks. Been playing catch-up after being off last week. I guess we'll monitor it and see how it feels. If it's needed we can create a follow-up ticket.

PR #474, opened 2023-07-14
MM-53292 - Telemetry for DM/GM ringing notifications

cpoile:
#### Summary
- Adds the following three events to our telemetry:
  - `notification_join`
  - `notification_dismiss`
  - `notification_click_goto_channel`
- And the `channels` source, which along with the `widget` and `expanded_view` show where the event came from.
- A couple notes for feedback:
  - I assumed it doesn't matter if the event from the condensed notification comes from the global widget vs the webapp widget (events for those two sources will have the source set to `widget`).
  - Similarly, I assumed it doesn't matter if the LHS notification is a regular vs condensed notification (events for those two components will have the source set to `channels`).
- The expanded view on desktop was mistakenly reporting `clientType` of `web` instead of `desktop`, so telemetry prior to v8.1 & Calls v0.18.0 will be incorrect on that variable.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-53292

### DIFF
--- a/webapp/src/components/incoming_calls/hooks.ts
+++ b/webapp/src/components/incoming_calls/hooks.ts
@@ -276,7 +276,7 @@ export const useOnChannelLinkClick = (call: IncomingCallNotification, onWidget =
 export const telemetrySource = (onWidget: boolean) => {
     if (onWidget) {
         return Telemetry.Source.Widget;
-    } else if (window.location.pathname.includes(`${pluginId}/expanded/`)) {
+    } else if (window.opener) {
         return Telemetry.Source.ExpandedView;
     }
 
--- a/webapp/src/components/incoming_calls/hooks.ts
+++ b/webapp/src/components/incoming_calls/hooks.ts
@@ -14,9 +14,10 @@ import {useIntl} from 'react-intl';
 import {useDispatch, useSelector, useStore} from 'react-redux';
 
 import {DID_NOTIFY_FOR_CALL, DID_RING_FOR_CALL} from 'src/action_types';
-import {dismissIncomingCallNotification, ringForCall, showSwitchCallModal} from 'src/actions';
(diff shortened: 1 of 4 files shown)


webapp/src/components/incoming_calls/hooks.ts
```diff
@@ -253,13 +260,25 @@ export const useOnChannelLinkClick = (call: IncomingCallNotification) => {
...
+        dispatch(trackEvent(Telemetry.Event.NotificationClickGotoChannel, source));
         const win = window.opener ? window.opener : window;
         win.postMessage({type: 'browser-history-push-return', message: {pathName: channelURL}}, window.origin);
     };
 };
+
+export const telemetrySource = (onWidget: boolean) => {
+    if (onWidget) {
+        return Telemetry.Source.Widget;
+    } else if (window.location.pathname.includes(`${pluginId}/expanded/`)) {
```
streamer45, line 279: Is the usual `window.opener` check not valid here?

streamer45 (approved):
> I assumed it doesn't matter if the event from the condensed notification comes from the global widget vs the webapp widget (events for those two sources will have the source set to widget).

UX wise it's still the widget so it's correct.

> Similarly, I assumed it doesn't matter if the LHS notification is a regular vs condensed notification (events for those two components will have the source set to channels).

I think it's fine. I suppose the only interesting insight from differentiating would be figuring out how often users receive multiple call invitations or when already in a call. But there are probably more direct ways to find that out.

> The expanded view on desktop was mistakenly reporting clientType of web instead of desktop, so telemetry prior to v8.1 & Calls v0.18.0 will be incorrect on that variable.

Nice catch!

webapp/src/components/incoming_calls/hooks.ts
cpoile: I don't see why it wouldn't be... Changed.

PR #479, opened 2023-07-25
[MM-53804] Add simple re-try logic to job creation

streamer45:
#### Summary

Now that we made the service operate almost statelessly to work well with Kubernetes, we need to account for the fact that subsequent HTTPs requests may hit different instances and fail due to missing authentication. The simplest way to deal with this problem is to implement a basic re-try mechanism.

I am not in love with this approach but the cleaner alternative is to either rethink the whole authentication logic (e.g. using a pre-shared secret) or implement some sort of shared auth service (e.g. using an external database). Neither of these are of simple/quick execution so opting for the dirtier fix.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-53804

### DIFF
--- a/server/job_service.go
+++ b/server/job_service.go
@@ -8,6 +8,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"math/rand"
 	"os"
 	"strings"
 	"time"
@@ -22,8 +23,12 @@ import (
 	recorder "github.com/mattermost/calls-recorder/cmd/recorder/config"
 )
 
-const jobServiceConfigKey = "jobservice_config"
-const runnerUpdateLockTimeout = 2 * time.Minute
+const (
+	jobServiceConfigKey             = "jobservice_config"
+	runnerUpdateLockTimeout         = 2 * time.Minute
+	maxReinitializationAttempts     = 10
+	reinitializationAttemptInterval = time.Second
+)
 
 var (
 	recordingJobRunner  = ""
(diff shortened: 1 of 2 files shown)


server/job_service.go
```diff
@@ -254,36 +259,11 @@ func (s *jobService) StopJob(channelID string) error {
...
-	// a common case when the offloader is running in kubernetes deployment. The
-	// solution is to re-initialize the service which will cause a new registration
-	// attempt.
-	if errors.Is(err, offloader.ErrUnauthorized) {
-		if err := s.reinitializeOnAuthError(); err != nil {
-			return job.Job{}, err
-		}
-		return s.client.GetJob(jobID)
-	}
-	return jb, err
```
streamer45, line 269: Technically we could use `retryJobService` here but we are likely going to deprecate this API given it won't work in K8S or any scenario involving load balancing of multiple offloader instances.

server/job_service.go
```diff
@@ -254,36 +259,11 @@ func (s *jobService) StopJob(channelID string) error {
...
-	// solution is to re-initialize the service which will cause a new registration
-	// attempt.
-	if errors.Is(err, offloader.ErrUnauthorized) {
-		if err := s.reinitializeOnAuthError(); err != nil {
-			return nil, err
-		}
-		return s.client.GetJobLogs(jobID)
-	}
-
-	return data, err
```
streamer45, line 286: Same reasoning as above.

cpoile (approved):
Looks good! True that it's a bit of brute force, but good enough for now. Thanks!

PR #483, opened 2023-07-31
[MM-53205] Remove reaction(s) from store when sending user leaves call

streamer45:
#### Summary

It doesn't look like we were updating the reducer when the user left the call. This should fix the issue. 
There could still be a race between disconnect and reaction events but that's more of an edge case. To cover that we'd have to check whether the user is connected to the call before queuing the reaction but at this time it seems a bit overkill.
 
#### Ticket Link

https://mattermost.atlassian.net/browse/MM-53205


cpoile (approved):
Nice, thank you for fixing it!

PR #484, opened 2023-07-31
[MM-53757] Don't show recording dialog on host change if previously dismissed

streamer45:
#### Summary

Adding an additional check in case the host changes after the banner is dismissed. In such case we should only show it again if the current user is the new host.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-53757


cpoile (approved):
Thank you!

PR #485, opened 2023-08-04
MM-53809 - Prevent desktop notification when the calls thread is open in the expanded view

cpoile:
#### Summary
- Simplest solution: don't notify the user if the user is in the call in that thread.
- Does work when call is started within the thread, and/or multiple times.
- ~~I'm not totally happy with this though, since it even blocks direct mentions... Maybe make an exception for those? But then we would have to do our own parsing.~~ fixed.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-53809

### DIFF
--- a/webapp/src/desktop_notifications.ts
+++ b/webapp/src/desktop_notifications.ts
@@ -3,10 +3,14 @@ import {Post} from '@mattermost/types/posts';
 import {DesktopNotificationArgs, Store} from 'src/types/mattermost-webapp';
 
 import {CALL_START_POST_TYPE} from 'src/constants';
-import {ringingEnabled} from 'src/selectors';
+import {
+    callThreadIDForCurrentCall,
+    channelIDForCurrentCall,
+    ringingEnabled,
+} from 'src/selectors';
 import {isDmGmChannel} from 'src/utils';
 
-export function desktopNotificationHandler(store: Store, post: Post, channel: Channel, args: DesktopNotificationArgs): {error?: string, args?: DesktopNotificationArgs} {
+export function desktopNotificationHandler(store: Store, post: Post, channel: Channel, args: DesktopNotificationArgs): { error?: string, args?: DesktopNotificationArgs } {
     if (args.notify) {
         // Calls will notify if:
         //  1. it's a custom_calls post (call has started)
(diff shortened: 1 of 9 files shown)


webapp/src/selectors.ts
```diff
@@ -240,6 +240,9 @@ export const callThreadIDForCallInChannel = (state: GlobalState, channelID: stri
     return pluginState(state).calls[channelID]?.threadID || '';
 };
 
+export const callThreadIDForCurrentCall = (state: GlobalState) =>
```
streamer45, line 243: If we follow the rules of consistency I totally made up (:p) this would probably become `threadIDForCurrentCall`.
But I also realize I forgot to update `callThreadIDForCallInChannel` which probably caused the confusion.

webapp/src/desktop_notifications.ts
```diff
@@ -21,6 +33,20 @@ export function desktopNotificationHandler(store: Store, post: Post, channel: Ch
...
+
+        // Do not notify for a call's thread if the user is currently in that call...
+        if (channelIDForCurrentCall(store.getState()) === post.channel_id &&
+            callThreadIDForCurrentCall(store.getState()) === post.root_id) {
+            let mentions = [];
+            if (msgProps.mentions) {
+                mentions = JSON.parse(msgProps.mentions);
+            }
+
+            // ...and wasn't directly mentioned.
```
streamer45, line 45: Nice, I wasn't sure myself if we wanted this exception but it probably makes sense.

webapp/src/desktop_notifications.ts
```diff
@@ -21,6 +33,20 @@ export function desktopNotificationHandler(store: Store, post: Post, channel: Ch
             ringingEnabled(store.getState())) {
             return {args: {...args, notify: false}};
         }
+
+        // Do not notify for a call's thread if the user is currently in that call...
+        if (channelIDForCurrentCall(store.getState()) === post.channel_id &&
+            callThreadIDForCurrentCall(store.getState()) === post.root_id) {
```
streamer45, line 39: Would this work as expected on Desktop? On the main app we wouldn't be on a call redux state wise I think.

cpoile: Yep, looks like we get call state from `loadCallState` (from ws event `call_state`) and from the `handleCallStart` (ws event `call_start`).

streamer45: Okay, I was mostly worried about the channel ID for current call being set but `DESKTOP_WIDGET_CONNECTED` should take care of it already. We are good :+1:

streamer45 (approved):

webapp/src/desktop_notifications.ts
cpoile: Ah yah, that too. Hah, funny enough, this took a bit longer than expected because I originally added the rootID to that desktop connected message (all through the stack), was about to push it, and then saw the redux state and thought, "there's an easier way to do this". 🤕

PR #486, opened 2023-08-04
[MM-53167] Persist participants list

streamer45:
#### Summary

We were already caching some stats for telemetry purposes so extending on those to keep track of all participants which we persist the usual way through post props. 

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-53167

### DIFF
--- a/server/channel_state.go
+++ b/server/channel_state.go
@@ -23,8 +23,8 @@ type userState struct {
 }
 
 type callStats struct {
-	Participants   int   `json:"participants"`
-	ScreenDuration int64 `json:"screen_duration"`
+	Participants   map[string]struct{} `json:"participants"`
+	ScreenDuration int64               `json:"screen_duration"`
 }
 
 type callState struct {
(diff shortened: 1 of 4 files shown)


crspeller:
@streamer45 Just to make sure, this would be a list of user ids of anyone who ever joined the call right?

streamer45:
> @streamer45 Just to make sure, this would be a list of user ids of anyone who ever joined the call right?

Yes the list is built incrementally so that it tracks all users who ever joined the call (except the bot), regardless of whether they have left before the end of the call.

cpoile (approved):
If we kept the timestamp of when they joined (instead of the empty struct), would that be useful?

streamer45:
> If we kept the timestamp of when they joined (instead of the empty struct), would that be useful?

We could, however it would be either the first or last join time unless we make it an array which becomes even more complex. I don't think we need that information for anything just yet. We can always augment later if necessary, props are pretty flexible.

PR #490, opened 2023-08-09
[MM-53986] Fix logic to toggle recordings

streamer45:
#### Summary

PR fixes an issue with toggling the recording state (start/stop). The logic was a bit overly complex which resulted in a bug preventing a host from stopping a recording if this was initializing but not started yet. Slash commands worked better although they were flawed as well (host would be allowed to start a recording while one was initializing, resulting in a server side error).

I am adding a dedicated selector so that we can share the same, simplified logic.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-53986

### DIFF
--- a/webapp/src/components/expanded_view/component.tsx
+++ b/webapp/src/components/expanded_view/component.tsx
@@ -99,6 +99,7 @@ interface Props extends RouteComponentProps {
     callHostID: string,
     callHostChangeAt: number,
     callRecording?: CallRecordingReduxState,
+    isRecording: boolean,
     hideExpandedView: () => void,
     showScreenSourceModal: () => void,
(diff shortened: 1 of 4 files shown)


cpoile (approved):
I haven't looked deeply, but is there a chance you can multi-click? 
Also wondering if the isRecording can be in a wrong state.

streamer45:
> I haven't looked deeply, but is there a chance you can multi-click? Also wondering if the isRecording can be in a wrong state.

Well, most of the recording state handling code suffers from potential races since we update the state from both a websocket handler and the response to the HTTP request so it's technically possible for the host client to get in the wrong state. At the end of the day we need to decide how much to invest in this sort of edge cases.

PR #491, opened 2023-08-09
[MM-53988] Improve logic for showing recording stopped banner

streamer45:
#### Summary

PR fixes an issues that would cause the *Recording has stopped. Processing...* message to appear to a newly changed host regardless of when the recording ended.

~~I am adding a threshold (currently one minute) after which we wouldn't show this. Technically we could just show it to the host at the time of stopping the recording but felt it may be useful to keep it around in case of a sudden host change. Overall though I don't have a strong opinion myself and feel it's more of a UX decision.~~

Regardless of what we do here, at some point (likely as part of host controls) we may want to design a message to inform a user that they became the host.

/cc @matthewbirtch in case you'd like to chime in on any of the above.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-53988

### DIFF
--- a/webapp/src/components/expanded_view/recording_info_prompt.tsx
+++ b/webapp/src/components/expanded_view/recording_info_prompt.tsx
@@ -5,7 +5,6 @@ import CompassIcon from 'src/components/icons/compassIcon';
 import RecordCircleIcon from 'src/components/icons/record_circle';
 import {
     CallRecordingDisclaimerStrings,
-    CALL_HOST_CHANGE_THRESHOLD,
 } from 'src/constants';
 import {CallRecordingReduxState} from 'src/types/types';
 import {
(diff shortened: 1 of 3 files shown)


matthewbirtch:
> I am adding a threshold (currently one minute) after which we wouldn't show this. Technically we could just show it to the host at the time of stopping the recording but felt it may be useful to keep it around in case of a sudden host change. Overall though I don't have a strong opinion myself and feel it's more of a UX decision.

To be honest, I'm not sure if this is something we need to do. It doesn't seem all that important to me. I'd say leave it out.

> Regardless of what we do here, at some point (likely as part of host controls) we may want to design a message to inform a user that they became the host.

I agree that we will want to do this as part of host controls effort (although I dont' know if/when that will get roadmapped).

webapp/src/components/expanded_view/recording_info_prompt.tsx
```diff
@@ -102,24 +102,39 @@ export default function RecordingInfoPrompt(props: Props) {
...
     if (disclaimerDismissedAt > props.recording?.start_at && props.hostChangeAt > disclaimerDismissedAt && !props.isHost) {
-        return null;
+        if (!shouldShowError) {
+            return null;
+        }
+    }
+
+    // If the user became host after the recording has ended we only show
+    // the "Recording has stopped" message if the change happened very
+    // recently (i.e. in the last minute).
```
cpoile, line 133: Needs to be updated?

cpoile (approved):
Thanks!

PR #493, opened 2023-08-14
[MM-54000] Fix alignment and stretching of call icon in post component

streamer45:
#### Summary

PR disables shrinking of the call icon to prevent the stretch effect when the content wraps (e.g. frequent in RHS). Also setting the vertical alignment to be centered.

#### Before

before_ended

#### After

after_ended

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-54000


matthewbirtch (approved):
Looks good based on the screenshot @streamer45. Thanks for fixing!

I couldn't review on my local instance as I'm having trouble getting it running so I'm trusting the screenshot :)

cpoile (approved):
Thanks!

PR #494, opened 2023-08-15
[MM-54144] Fix screen thumbnail content overflow

streamer45:
#### Summary

It's possibly the third time I try to fix this. It's easily reproducible on Community but not on local setup so maybe there's some conflicting definition taking place. Anyhow, forcing a max height on the video player element seems to be enough.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-54144


cpoile (approved):
Third time's the charm.

PR #495, opened 2023-08-16
[MM-54157] Fix potential nil access upon starting a recording

streamer45:
#### Summary

PR fixes a potential panic due to accessing a `nil` `recState`. 

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-54157

### DIFF
--- a/server/recording_api.go
+++ b/server/recording_api.go
@@ -36,6 +36,11 @@ func (p *Plugin) recJobTimeoutChecker(callID, jobID string) {
 	// If the recording hasn't started (bot hasn't joined yet) we notify the
 	// client.
 	if recState.StartAt == 0 {
+		if recState.JobID != jobID {
+			p.LogInfo("a new job has started in between, exiting", "callID", callID, "jobID", jobID)
+			return
+		}
+
 		p.LogError("timed out waiting for recorder bot to join", "callID", callID, "jobID", jobID)
 
 		state.Call.Recording = nil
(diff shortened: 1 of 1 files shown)


server/recording_api.go
```diff
@@ -185,29 +191,23 @@ func (p *Plugin) handleRecordingStopAction(state *channelState, callID string) (
...
-	// since it could take a few seconds to complete but we want clients
-	// to get their local state updated as soon as it changes on the server.
-	p.publishWebSocketEvent(wsEventCallRecordingState, map[string]interface{}{
-		"callID":   callID,
-		"recState": recState.getClientState().toMap(),
-	}, &model.WebsocketBroadcast{ChannelId: callID, ReliableClusterSend: true})
-
-	// We don't want to keep the lock while making the API call to the service since it
-	// could take a while to return.
-	p.unlockCall(callID)
```
streamer45, line 198: Unrelated to the panic but since we moved to the new job API we don't need to do this, the stop action is completely asynchronous and safe to be called under lock.

server/recording_api.go
```diff
@@ -36,6 +36,11 @@ func (p *Plugin) recJobTimeoutChecker(callID, jobID string) {
 	// If the recording hasn't started (bot hasn't joined yet) we notify the
 	// client.
 	if recState.StartAt == 0 {
+		if recState.JobID != jobID {
+			p.LogInfo("a new job has started in between, exiting", "callID", callID, "jobID", jobID)
+			return
+		}
```
streamer45, line 42: Adding an extra check to avoid making potentially inconsistent state updates due to previously failed jobs. Ideally we'd cancel this timeout check whenever we stop a recording but the synchronization needed would make the code more complex so it seems like a few extra goroutines that return early are not too bad.

cpoile (approved):

PR #496, opened 2023-08-17
Implement `setup` flag for load-test client

streamer45:
#### Summary

I used to constantly patch the local load-test client every time I wanted to test on Community since we cannot use a system admin account there. So I am adding a proper `setup` flag that when set to `false` avoids any sort of setup (e.g. registering users, joining channels) which works well for cases where the needed entities are already in place.

### DIFF
--- a/lt/client.go
+++ b/lt/client.go
@@ -81,7 +81,7 @@ type config struct {
 	screenSharing bool
 	recording     bool
 	simulcast     bool
-	noSetup       bool
+	setup         bool
 }
 
 type user struct {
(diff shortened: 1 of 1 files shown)


lt/client.go
```diff
@@ -793,23 +798,25 @@ func (u *user) Connect(stopCh chan struct{}, channelType model.ChannelType) erro
...
-	defer cancel()
-	_, _, err = client.AddTeamMember(ctx, u.cfg.teamID, user.Id)
-	if err != nil {
-		return err
-	}
-	cancel()
-
-	if channelType == "O" || channelType == "P" {
-		// join channel
+	if !u.cfg.noSetup {
```
cpoile, line 801: I find negative tests a little ugly, but I guess it makes sense here.

cpoile (approved):
Nice QOL improvement

lt/client.go
streamer45: We could in theory have a `-setup` flag default to `true` but not sure if it becomes any clearer.

cpoile: It's a double negative, too. Even worse...

cpoile: Sorry, I meant the current state is a double negative. `if u.cfg.setup {` is great.

streamer45: Cool, should be clearer now.

cpoile (approved):
Thanks!

PR #489, opened 2023-08-08
[MM-53951] Catch some potentially unhandled errors

streamer45:
#### Summary

We had a recording job failing during initialization (see ticket). Adding a couple of extra catch blocks to try and get better output. 

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-53951

PR #497, opened 2023-08-21
Fix initialization for standalone client

streamer45:
#### Summary

A change in https://github.com/mattermost/mattermost-plugin-calls/pull/489 caused the standalone client to fail to connect when starting a call. We move the check into the logic dedicated to the recording client which can make the assumption of always joining an existing/ongoing call.

This problem was caught by e2e tests but unfortunately they are failing to update their status to Github due to an internal security issue which is being worked on (see https://community.mattermost.com/private-core/pl/yxta5iem53y1br93ksjdj7tkyh).

### DIFF
--- a/standalone/src/init.ts
+++ b/standalone/src/init.ts
@@ -141,7 +141,8 @@ async function fetchChannelData(store: Store, channelID: string) {
     );
 
     if (!resp.call) {
-        throw new Error('call missing from response');
+        // This is expected if the client is starting the call.
+        return;
     }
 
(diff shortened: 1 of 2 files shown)


cpoile (approved):
Nice that it was caught at least. :)

streamer45:
Tests passed: https://mattermost-cypress-report.s3.us-east-1.amazonaws.com/calls-pw/2023_08_21__22_40_51/index.html , merging

PR #498, opened 2023-08-22
[MM-54216] Bump ws rate limiting burst

streamer45:
#### Summary

It looks like Firefox can be spamming a very long list of candidates if running on an instance with many network interfaces and hit the rate limiter which will drop messages, potentially preventing proper connectivity. What's worse, upon first unmute or screen sharing attempt it will do more of the same.

Note to self: this should be the last time we bump this. If the problem reappears we need to investigate a different solution.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-54216

### DIFF
--- a/server/session.go
+++ b/server/session.go
@@ -47,7 +47,8 @@ type session struct {
 	leaveCh chan struct{}
 	left    int32
 
-	limiter *rate.Limiter
+	// rate limiter for incoming WebSocket messages.
+	wsMsgLimiter *rate.Limiter
 }
 
 func newUserSession(userID, channelID, connID string, rtc bool) *session {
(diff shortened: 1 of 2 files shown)


server/session.go
```diff
@@ -47,7 +47,8 @@ type session struct {
 	leaveCh chan struct{}
 	left    int32
 
-	limiter *rate.Limiter
+	// rate limiter for incoming WebSocket messages.
+	wsMsgLimiter *rate.Limiter
```
cpoile, line 51: Thank you for the nice var name!

cpoile (approved):

streamer45:
E2E passed: https://mattermost-cypress-report.s3.us-east-1.amazonaws.com/calls-pw/2023_08_22__23_27_10/index.html

PR #500, opened 2023-08-23
Fix recording error logic

streamer45:
#### Summary

The condition to show the error was actually inverted, causing the banner to get stuck indefinitely in case of error.

Note: we should likely review this logic and at least have it in a single place, although I don't know how much can be simplified.


cpoile (approved):
Yah, it's just a bunch of ifs basically (isn't everything?), not sure how much value over effort we'd get by building a system for it.

PR #501, opened 2023-08-23
[MM-54232] Simplify job service logic

streamer45:
#### Summary

PR greatly simplifies the overall logic needed to initialize and use the job service which in turn will make the service properly load balanceable. See related PR for the full context.

#### Related PR

https://github.com/mattermost/calls-offloader/pull/38

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-54232


cpoile (approved):
This is definitely simpler, nice!

streamer45:
Tested this a bit in a HA setup and it's working fine. The only edge case I could find is if the credentials are somehow deleted from the `KVStore` after the plugin initialized but I don't think we should invest much on that since it's not something that should ever happen.

streamer45:
This will need a new `calls-offloader` version to be released in order to get E2E tests to pass.

PR #508, opened 2023-08-29
MM-54243 - Slash command for host change

cpoile:
#### Summary
- We might want to cherry pick the first commit to main?
  - re: the fix -- I wanted to continue in that loop so that we still get the other channels and return whatever valid data we have.
- Please let me know if I overlooked any edge cases. Decided to not send an empty host string when the host disconnects -- this way that user can reconnect and they will automatically be the host again (no need to rerun the slash command)

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-54243
- Fixes https://mattermost.atlassian.net/browse/MM-54313


server/store.go
```diff
@@ -53,7 +53,7 @@ func (p *Plugin) KVGet(key string, fromWriter bool) ([]byte, error) {
 	if !fromWriter {
 		data, appErr := p.API.KVGet(key)
 		if appErr != nil {
-			return nil, appErr
+			return nil, fmt.Errorf("failed to kvget for key: %s error: %s", key, appErr.Error())
```
streamer45, line 56: Could we wrap the error using the dedicated `%w`?

server/host_controls.go
```diff
@@ -0,0 +1,51 @@
...
+	if state == nil || state.Call == nil {
+		return errors.New("no call ongoing")
+	}
+
+	if state.Call.HostID == newHostID {
+		return errors.New("user is already the host")
+	}
+
+	inCall := false
+	for id := range state.Call.Users {
```
streamer45, line 29: `Users` should be a map, why are we looping?

server/host_controls.go
```diff
@@ -0,0 +1,51 @@
...
+	}
+	defer p.unlockCall(channelID)
+
+	if state == nil || state.Call == nil {
+		return errors.New("no call ongoing")
+	}
+
+	if state.Call.HostID == newHostID {
+		return errors.New("user is already the host")
+	}
```
streamer45, line 26: Wondering if this should be a no-op vs error to make it idempotent. Not a big deal.

server/session.go
```diff
@@ -130,7 +130,7 @@ func (p *Plugin) addUserSession(state *channelState, userID, connID, channelID s
 		}
 	}
 
-	if state.Call.HostID == "" && userID != p.getBotID() {
+	if state.Call.HostID == "" && userID != p.getBotID() && !state.Call.HostLocked {
 		state.Call.HostID = userID
 	}
```
streamer45, line 135: Slightly unrelated but maybe it would be cleaner if we always call `state.Call.getHostID()` here and keep any specific logic inside the method. Same would apply for when removing a session.

streamer45 (commented):
Thanks, left some feedback.

> We might want to cherry pick the first commit to main?

We should.

> re: the fix -- I wanted to continue in that loop so that we still get the other channels and return whatever valid data we have.

Sounds reasonable given the error is likely to be due to a stuck call.

server/host_controls.go
cpoile: Oh god, embarrassing. Thanks.

server/host_controls.go
cpoile: Yep, makes sense, thanks.

server/session.go
cpoile: That's nicer, keeps the logic all in one place, thanks.

server/host_controls.go
```diff
@@ -0,0 +1,44 @@
...
+	"fmt"
+	"github.com/mattermost/mattermost/server/public/model"
+	"github.com/pkg/errors"
+)
+
+func (p *Plugin) changeHost(requesterID, channelID, newHostID string) error {
+	if isAdmin := p.API.HasPermissionTo(requesterID, model.PermissionManageSystem); !isAdmin {
+		return errors.New("no permissions to change host")
+	}
+
```
streamer45, line 13: Can we add a preemptive error to make sure we cannot assign the bot as host?

server/host_controls.go
```diff
@@ -0,0 +1,44 @@
...
+	if state.Call.HostID == newHostID {
+		return nil
+	}
+
+	if _, ok := state.Call.Users[newHostID]; !ok {
+		return errors.New("user is not in the call")
+	}
+
+	state.Call.HostID = newHostID
+	state.Call.HostLocked = true
```
streamer45, line 33: I suppose that semantically if we lock the host it doesn't really matter whether the user is on the call but we can review in the future, no action to take here.

server/channel_state.go
```diff
@@ -165,8 +166,22 @@ func (us *userState) getClientState() UserStateClient {
...
+
+	// if current host isn't in the call anymore, remove
+	if _, ok := cs.Users[cs.HostID]; !ok {
+		cs.HostID = ""
+	}
 
+	// don't change hosts if we have one already
+	if cs.HostID != "" {
+		return cs.HostID
+	}
```
streamer45, line 181: I admit I find this check a little counter intuitive. I wonder if it would be clearer to return early if the host is already present and avoid the side effect altogether.

cpoile: I agree it's a little counter-intuitive, but I'm not sure how to return earlier -- we have to do the two actions above before checking if we have a host...

streamer45: I was thinking of simply doing this:

```go
	if _, ok := cs.Users[cs.HostID]; ok {
                return cs.HostID
	}
```

And otherwise proceed with the usual "longest standing participant" check. We likely just need to move the `getHostID` call after we set the user in `addUserSession` but it should work.

cpoile: Oh, I see what you mean now -- combine the two checks.

streamer45: @cpoile Not sure it will work as expected unless you also do the latter change in `addUserSession`.

cpoile: Sorry, not sure what you mean by "latter change"

streamer45: @cpoile We need to change the order or the state we act upon won't be the expected one since we add the user to the map after we call `getHostID()`.

https://github.com/mattermost/mattermost-plugin-calls/blob/c3513b7a7458832d443fd991a121cbeedb294265/server/session.go#L133C1-L138

cpoile: And when I was testing, I wasn't looking at the host until I connected two people... which worked. Hah. 😮‍💨

streamer45 (approved):
Thanks :raised_hands:

streamer45:
@cpoile Tests are failing, not sure if related to the latest change but likely I guess.

cpoile:
@streamer45 This is probably related to calls recordings in cloud, right?

streamer45:
@cpoile Did you merge with `main`? Looks like it's running an out of date recorder.

streamer45:
@cpoile This would have saved @lieut-data some headaches earlier today. I'd be happy to get it in as a first improvement while we wait for the proper feature. The only thing I'd ask you would to hide this behind an enterprise license check for the time being.

cpoile:
Without enterprise, the host command acts like it's not there.

streamer45:
@cpoile This will need some love now that multi sessions are merged. Feel free to rebase/cherry-pick and force push.

cpoile:
Replaced by #668

PR #510, opened 2023-08-30
[MM-53081] Redux store refactor

streamer45:
#### Summary

PR reviews most of our reducers and selectors to try and improve a few things:

- Renaming a bunch of very badly thought and/or confusing objects.
  - Preferring to be on the verbose side (e.g. `profilesInCallInCurrentChannel`), clarity over efficiency. 
- Make selectors more performant through proper use of `createSelector`.
  - Mostly tried to optimize all those instances in which we wanted something for the current call or channel. There may be a couple of additional improvements but nothing major I think.
- Trying to find some consistency in how we design reducers.
  - On this I am trying to stick to the idea that for anything immutable related to a call (e.g. startAt, threadID) we can use a common call state but if something can change throughout the call then we should have a dedicated reducer (e.g. hostID, screenSharingID).

**Notes**

- E2E passing is the first step to validate this but there's always a chance we will break something untested so we should plan a long enough soaking time on Community.
- It would be preferable if any future non urgent PRs touching these files could be based on top of this one to avoid endless conflicts.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-53081


webapp/src/reducers.ts
```diff
@@ -26,30 +26,29 @@ import {
...
-    VOICE_CHANNEL_USER_SCREEN_ON,
-    VOICE_CHANNEL_USER_UNMUTED,
-    VOICE_CHANNEL_USER_UNRAISE_HAND,
-    VOICE_CHANNEL_USER_VOICE_OFF,
-    VOICE_CHANNEL_USER_VOICE_ON,
-    VOICE_CHANNEL_USERS_CONNECTED,
-    VOICE_CHANNEL_USERS_CONNECTED_STATES,
-    VOICE_CHANNEL_USER_JOINED_TIMEOUT,
+    CALL_END,
+    CALL_REC_PROMPT_DISMISSED,
```
streamer45, line 30: I stripped most of the unnecessary prefixes in this file with the assumption that it's all Calls related. Please let me know if you'd prefer a more verbose take on this side as well.

cpoile: Definitely this is better, thanks!

webapp/src/reducers.ts
```diff
@@ -593,79 +613,105 @@ const callsRecordings = (state: { [callID: string]: CallRecordingState } = {}, a
...
-    dismissedNotification: { [userID: string]: boolean },
+// callState should only hold immutable data, meaning those
+// fields that don't change for the whole duration of a call.
+export type callState = {
+    ID: string;
+    startAt: number;
+    channelID: string;
+    threadID: string;
+    ownerID: string;
+}
```
cpoile, line 624: Maybe we should put this in the types file?

webapp/src/reducers.ts
```diff
@@ -593,79 +613,105 @@ const callsRecordings = (state: { [callID: string]: CallRecordingState } = {}, a
...
 };
 
-const voiceChannelRootPost = (state: { [channelID: string]: string } = {}, action: { type: string, data: { channelID: string, rootPost: string } }) => {
+export type hostsState = {
+    [channelID: string]: {
+        hostID: string;
+        hostChangeAt?: number;
+    };
+}
+
```
cpoile, line 657: Maybe this too?

webapp/src/reducers.ts
```diff
@@ -760,15 +806,15 @@ const callsUserPreferences = (state = CallsUserPreferencesDefault, action: { typ
     }
 };
 
-interface recentlyJoinedUsersState {
-    [channelID: string]: string[],
+export type recentlyJoinedUsersState = {
+    [channelID: string]: string[];
```
cpoile, line 810: I guess this too, if it's being used outside the current file?

webapp/src/selectors.ts
```diff
@@ -35,199 +44,243 @@ import {pluginId} from './manifest';
...
 
-export const voiceChannelCallOwnerID = (state: GlobalState, channelID: string): string | undefined => {
-    return pluginState(state).voiceChannelCalls[channelID]?.ownerID;
+export const callOwnerIDForCallInChannel = (state: GlobalState, channelID: string): string | undefined => {
+    return pluginState(state).calls[channelID]?.ownerID;
 };
 
-export const voiceChannelCallHostID = (state: GlobalState, channelID: string) => {
-    return pluginState(state).voiceChannelCalls[channelID]?.hostID;
+const hostsInCalls = (state: GlobalState): hostsState => {
```
cpoile, line 203: I'm not crazy about the name here... But I can't think of anything better. Something like `hostByCall`?

cpoile (approved):
Sorry I don't have any more substantive comments, it looks good. Should bring more clarity when you're deep into the code and searching for the right selector.
Thanks!

webapp/src/selectors.ts
streamer45: It's really just internal given it's the base selector so we may as well just call it `hosts` or `callsHosts`. Ultimately it doesn't matter that much I guess.

webapp/src/reducers.ts
streamer45: Yeah, I am trying to find a compromise between client side state, server side state and redux state. Not super easy at times. Will look into this.

webapp/src/selectors.ts
```diff
@@ -35,199 +44,243 @@ import {pluginId} from './manifest';
...
     return users && users.length > 0;
 };
 
-export const connectedChannelID = (state: GlobalState): string | null =>
-    pluginState(state).connectedChannelID;
+export const channelIDForCurrentCall = (state: GlobalState): string =>
+    pluginState(state).channelID || '';
 
-export const connectedChannel: (state: GlobalState) => Channel | null =
+export const channelForCurrentCall: (state: GlobalState) => Channel =
```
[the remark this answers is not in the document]
streamer45: Yeah, would be safer. I'll update.

webapp/src/reducers.ts
streamer45: @cpoile Let me track a proper review of types as a separate effort especially since I am planning some changes on that front.

streamer45: Created https://mattermost.atlassian.net/browse/MM-54429

PR #512, opened 2023-08-31
[MM-54328] Fix call state race condition

streamer45:
#### Summary

PR fixes a race condition that could cause the call state to be incorrectly updated with stale data. Contrary to what I originally thought, this didn't happen on join. In fact, we don't fetch the call state when joining but when switching channel. So if switching to the channel the call was ongoing while some other users were joining, it could happen (actually quite easy to reproduce) that the final state missed some profiles.

The solution here is to send the whole call state through the websocket channel upon joining the call. This ensures the call is up to date as other events are guaranteed to be received in order.

To be noted, this PR only fixes the above case of joining a call. There are still a few possible races when loading Mattermost the first time. Unfortunately this includes opening the popout window although I couldn't easily reproduce this case so far. To fix that we likely need something a bit more involved because it's not straightforward to guarantee sequencing of events there. Need to think it a bit more.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-54328


webapp/src/index.tsx
```diff
@@ -833,11 +781,15 @@ export default class Plugin {
         this.unsubscribers.push(store.subscribe(() => {
             const currentChannelId = getCurrentChannelId(store.getState());
             if (currChannelId !== currentChannelId) {
+                const firstLoad = !currChannelId;
                 currChannelId = currentChannelId;
 
-                fetchChannelData(currChannelId).then((actions) =>
-                    store.dispatch(batchActions(actions)),
-                );
```
streamer45, line 840: I am removing this as I don't think it's needed anymore. We fetch all channels anyway on first load and then rely on websocket to keep things up to date. I couldn't see any side effect so far but will need to monitor.

server/websocket.go
```diff
@@ -650,6 +651,16 @@ func (p *Plugin) handleJoin(userID, connID, channelID, title, threadID string) (
 		}, &model.WebsocketBroadcast{ChannelId: channelID, ReliableClusterSend: true})
 	}
 
+	clientStateData, err := json.Marshal(state.Call.getClientState(p.getBotID(), userID))
+	if err != nil {
+		p.LogError("failed to marshal client state", "err", err.Error())
+	} else {
+		p.publishWebSocketEvent(wsEventCallState, map[string]interface{}{
```
streamer45, line 658: I realize I could have re-used `wsCallStart` by extending it but it seemed cleaner to have a new event, especially as we'll likely need to break/deprecate some before 1.0.

server/websocket.go
```diff
@@ -650,6 +651,16 @@ func (p *Plugin) handleJoin(userID, connID, channelID, title, threadID string) (
...
+			"channel_id": channelID,
+			"call":       string(clientStateData),
```
streamer45, line 660: Unfortunately we need to do this horrible JSON in JSON trick since the generic map would otherwise fail to marshal out structure.

cpoile (commented):
This looks like it will simplify things, thank you!

server/websocket.go
streamer45: It will arrive before because it's queued before. But it wouldn't really matter since the state would be the same and loaded on top. If anything we could skip it but it would be a breaking change so not going to do it here.

server/websocket.go
streamer45: Yes, but it would need to be MM server side change (gob encoding) which should then be aware of call specific types. Not sure I want to go there just yet.

standalone/src/recording/index.tsx
```diff
@@ -142,6 +126,23 @@ async function wsHandlerRecording(store: Store, ev: WebSocketMessage<WebsocketEv
 
         break;
     }
+    case `custom_${pluginId}_call_state`: {
+        const data = ev.data as CallStateData;
+
+        const profiles = profilesInCallInChannel(store.getState(), data.channel_id);
```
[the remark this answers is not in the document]
streamer45: Yeah, I had to tweak a couple of things to make sure that was the case, the key point would be this:

https://github.com/mattermost/mattermost-plugin-calls/blob/dee47eeaff59554d21d0c41f7edf5b1d037cb213/standalone/src/init.ts#L267

webapp/src/actions.ts
```diff
@@ -425,3 +433,90 @@ export const stopRingingForCall = (callID: string): ActionFunc => {
         return {};
     };
 };
+
+export const loadCallState = (channelID: string, call: any) => async (dispatch: DispatchFunc, getState: GetStateFunc) => {
```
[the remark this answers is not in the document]
streamer45: We have the type already, not sure why it was left as `any`, maybe I was debugging something. I'll update, thanks.

server/websocket.go
cpoile: Ah -- I mixed the two, I meant the other way around. If call state arrives before join -- would that be possible? If it is, would that break anything on the redux side?

server/websocket.go
cpoile: Ah yah, that's too much. Forget I asked. Maybe... Could we add `channel_id` to the call state, and then send the whole thing (no embedding)?

standalone/src/recording/index.tsx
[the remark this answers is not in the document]
cpoile: Ah ok, that's good. But just to double check -- if the call state includes the user list, is it guaranteed to be up to date at this point? If so, could we use that instead of relying on the store's state?

server/websocket.go
streamer45: I don't think so, joining "twice" wouldn't do anything. But it also shouldn't be possible, events are queued and received in the order they are sent. We are also making sure to use the TCP channel intra-cluster so there's no way they'd get out of order.

server/websocket.go
streamer45: We can only send a map, not custom data so the problem would be the same. It was a simple object we could probably create a custom marshaler to covert from struct to `map[string]any` but even then it's ugly.

standalone/src/recording/index.tsx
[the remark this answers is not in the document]
streamer45: The call state only includes the user ids of the participants. We still need to potentially fetch the profiles. But yes, it's guarantee to be up to date once sent because we are sending the event with the call mutex locked on the server. Any subsequent change (e.g. join/leave) will new events which are guaranteed to come after the initial one.

server/websocket.go
cpoile: But if it's a map[string]interface{}, that can hold a single level struct that's been converted to json, and then you can use that directly as the object in the front end, right?  (I don't mean to delay the PR over it, I'm just trying to think how we can make this better, because sending over a string just feels wrong)

server/websocket.go
cpoile: Cool, thanks!

cpoile (approved):

server/websocket.go
streamer45: @cpoile If you save it as a JSON string like we are doing yes. But otherwise we'll eventually try to marshal that in GOB and fail to do so because the plugin RPC layer doesn't have a marshaler for our type.

cpoile: I guess I was thinking of something like marshalling into a json blob, then unmarshalling into a map[string]interface{}, then sending that? Something along those lines. (Not sending a go struct.)

streamer45: Right, that would work for a simple struct but will break in our case because we'd have to do that "deeply", and check any substructs like userState, recordingState and such. It's doable but you need to have this monster marshaler, maybe even use reflect to make it actually helpful, not sure it's worth it.

cpoile: Yah, anyway, just thinking out loud. :)

PR #507, opened 2023-08-28
MM-54272 - Upgrade npm/node to match Mattermost webapp

cpoile:
#### Summary
- Needed to upgrade a few deps.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-54272

PR #513, opened 2023-08-31
Automated TS import/exports order

streamer45:
#### Summary

The [`eslint-plugin-simple-import-sort`](https://github.com/lydell/eslint-plugin-simple-import-sort) plugin seems to do a better job at automatically fixing issues. That said, I don't know if we should add some special rules. What I'd like to avoid is IDEs interfering if possible. @cpoile Could you check whether we need some special extra rules to get some consistency with your environment as well?


streamer45:
Closing in favor of https://github.com/mattermost/mattermost-plugin-calls/pull/507

PR #532, opened 2023-09-25
[MM-54551] Fix potential nil dereference panic

streamer45:
#### Summary

We had a report of failures to join calls starting in v0.19. At first it looked like a networking issue but upon further inspection of the logs we were able to find multiple plugin crashes due to SIGSEGV. 

Even if properly set to `false` in `config.json` the returned object was a nil pointer. Adding a couple of extra checks to avoid future regressions.

Given the severity of this I'd like to cut a new dot release once this is merged.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-54551

### DIFF
--- a/server/utils.go
+++ b/server/utils.go
@@ -43,7 +43,12 @@ func (p *Plugin) setHandlerID(nodeID string) error {
 
 func (p *Plugin) getNotificationNameFormat(userID string) string {
 	config := p.API.GetConfig()
-	if !*config.PrivacySettings.ShowFullName {
+	if config == nil {
+		p.LogError("failed to get config")
+		return model.ShowUsername
+	}
+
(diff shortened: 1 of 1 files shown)


cpoile (approved):
Huh. Guess we do need to always check that. Thanks!

cpoile:
Any idea why it's returning a nil config?

streamer45:
Also to note, what made debugging this worse was that until we enabled DEBUG level logs we couldn't get the plugin panics to show. Something to review on the server most likely.

cpoile:
If we're looking at that, then we should note that Plugin panics should be errors in the first place.

streamer45:
> If we're looking at that, then we should note that Plugin panics should be errors in the first place.

Yeah, currently it's something handled by the framework (https://mattermost.atlassian.net/browse/MM-54562), unless we start to implement recovers, something I tried to avoid so far.

PR #540, opened 2023-10-02
Add go version to cd

cpoile:
#### Summary
- We'll probably need this when we cut a release

#### Ticket Link
- none


streamer45 (approved):
Sounds good. We should probably bump straight to 1.21 as soon as monorepo does.

PR #542, opened 2023-10-02
[MM-54589] Implement job status API

streamer45:
#### Summary

PR adds a new bot API endpoint to explicitly report job statuses (e.g. failure, started).

Of **note**, we are now publishing a `github.com/mattermost/mattermost-plugin-calls/server/public` Go module with shared types. I'll slowly start to port over some common structures as needed (e.g. some of those defined in https://github.com/mattermost/rtcd/pull/115). This is still subject to breaking changes and should only be used internally for the time being.

#### Related PR

https://github.com/mattermost/calls-recorder/pull/46

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-54589

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -16,8 +16,9 @@ require (
 	github.com/mattermost/calls-offloader v0.3.2
 	github.com/mattermost/calls-recorder v0.4.2
 	github.com/mattermost/logr/v2 v2.0.16
+	github.com/mattermost/mattermost-plugin-calls/server/public v0.0.1
 	github.com/mattermost/mattermost/server/public v0.0.9-0.20230824163353-69c11cfe1403
-	github.com/mattermost/rtcd v0.11.1
+	github.com/mattermost/rtcd v0.11.3-0.20230913232654-bdaaa43e3e1c
 	github.com/mattermost/squirrel v0.2.0
 	github.com/pion/interceptor v0.1.12
 	github.com/pion/rtp v1.7.13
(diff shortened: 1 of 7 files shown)


go.mod
```diff
@@ -29,6 +30,8 @@ require (
 
 replace github.com/pion/interceptor v0.1.12 => github.com/streamer45/interceptor v0.0.0-20230202152215-57f3ac9e7696
 
+replace github.com/mattermost/mattermost-plugin-calls/server/public => ./server/public
```
streamer45, line 33: I actually tried to use Go workspaces to avoid this but ended up in a bit of rabbit hole with some unexpected failures on dependencies. Will look into this at another time as it's working fine for now.

server/bot_api.go
```diff
@@ -245,6 +248,73 @@ func (p *Plugin) handleBotPostRecordings(w http.ResponseWriter, r *http.Request,
...
+		}
+
+		if state.Call.Recording.EndAt > 0 {
+			res.Err = "recording has ended"
+			res.Code = http.StatusBadRequest
+			return
+		}
+
+		state.Call.Recording.EndAt = time.Now().UnixMilli()
+		state.Call.Recording.Err = status.Error
```
cpoile, line 296: Maybe I'm misreading, but every api hit will assume you're stopping the recording? We're not using the `Status` field?

cpoile (commented):
Just one question:

server/bot_api.go
streamer45: You are correct. We are only handling one status at the moment so it doesn't matter but yes I definitely need to add a check as we'll be soon using it for the started event as well. I'll send an update :+1:

cpoile (approved):
Awesome, makes more sense now (even if only logically). Thanks!

cpoile (approved):

PR #543, opened 2023-10-03
MM-54446 - E2E tests for notifications features

cpoile:
#### Summary
- This basically turns all the test cases from https://github.com/mattermost/mattermost-plugin-calls/pull/445 and https://github.com/mattermost/mattermost-plugin-calls/pull/452 into e2e test cases.
- I'm doing some window trickery to track notifications in and out, which was a heck of a lot simpler than trying to mock the browser's notifications and sound api.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-54446


e2e/page.ts
```diff
@@ -22,6 +24,12 @@ export default class PlaywrightDevPage {
         await this.page.goto(`${baseURL}/${defaultTeam}/messages/@${username}`);
     }
 
+    async createAndGoToGM(userIDs: string[]) {
+        const channel = await apiCreateGroupChannel(this.page.request, userIDs);
+        await this.page.goto(`${baseURL}/${defaultTeam}/channels/${channel.name}`);
+        return channel;
+    }
```
streamer45, line 31: Is there a reason we are not creating GMs in advance like the rest of resources?

e2e/playwright.config.ts
```diff
@@ -7,7 +7,7 @@ const config: PlaywrightTestConfig = {
     retries: process.env.CI ? 2 : 1,
     workers: 4,
     fullyParallel: true,
-    timeout: 90 * 1000,
+    timeout: 150 * 1000,
```
streamer45, line 10: I'd prefer if we used a per-test timeout override (`test.setTimeout(150000)`) rather than doing this globally as most tests shouldn't require all that time.

e2e/utils.ts
```diff
@@ -1,31 +1,44 @@
...
+    return [`${names[0]}StorageState.json`, `${names[1]}StorageState.json`, `${names[2]}StorageState.json`];
+}
+
+export function getUserIDsForTest() {
+    const storageStates = getUserStoragesForTest();
+    return storageStates.map((ss) => {
+        const data = JSON.parse(fs.readFileSync(ss, 'utf-8'));
+        const idx = data.cookies.findIndex((c: { name: string; }) => c.name === 'MMUSERID');
+        return data.cookies[idx].value;
+    });
```
streamer45, line 41: That's quite elaborate. We may reach a point where we need to either store these information in advance or use an API client to fetch them but this works too for now :)

e2e/utils.ts
```diff
@@ -41,3 +54,17 @@ export async function startCall(userState: string) {
...
+export async function startDMWith(userState: string, targetUserName: string) {
+    const userPage = await newUserPage(userState);
+    await userPage.gotoDM(targetUserName);
+    return userPage;
+}
+
+export async function startGMWith(userState: string, targetUserIDs: string[]) {
+    const browser = await chromium.launch();
+    const context = await browser.newContext({storageState: userState});
+    const userPage = new PlaywrightDevPage(await context.newPage());
```
streamer45, line 66: Any reason for not using `newUserPage()` here?

webapp/src/actions.ts
```diff
@@ -409,20 +416,30 @@ export const removeIncomingCallNotification = (callID: string): ActionFunc => {
 export const ringForCall = (callID: string, sound: string) => {
     return async (dispatch: DispatchFunc) => {
         notificationSounds?.ring(sound);
+
+        // window.e2eNotificationsSoundedAt is added when running the e2e tests
+        if (window.e2eNotificationsSoundedAt) {
+            window.e2eNotificationsSoundedAt.push(Date.now());
+        }
```
streamer45, line 423: I am sure you put some serious thought into this but can't help but feeling a bit unsatisfied by the need to tinker with application code purely to support tests.

e2e/global-setup.ts
```diff
@@ -123,6 +124,17 @@ async function globalSetup(config: FullConfig) {
         headers,
     });
 
+    // set automatic replies setting (ooo) to true for notifications tests
+    resp = await adminContext.put('api/v4/config/patch', {
```
streamer45, line 128: What is needed for exactly?

e2e/page.ts
```diff
@@ -66,4 +74,40 @@ export default class PlaywrightDevPage {
...
+
+    async hideDocument(hide = true) {
+        await this.page.evaluate((hidden) => {
+            Object.defineProperty(document, 'visibilityState', {value: hidden ? 'hidden' : 'visible', writable: true});
+            Object.defineProperty(document, 'hidden', {value: hidden, writable: true});
+            document.dispatchEvent(new Event('visibilitychange'));
+        }, hide);
+    }
+
+    async expectNotifications(numDesktopNotificationsRejected: number, numDesktopNotificationsSent: number, numNotificationsSounded: number, numNotificationsStoppedAt: number) {
```
streamer45, line 95: I understand it gets verbose but looking at `expectNotifications(0,0,0,0`) getting called from someone that hasn't designed all of this feels very confusing :p

streamer45 (commented):
This is an incredible effort :tophat: 

Left a few comments and questions.

e2e/page.ts
cpoile: Well, it was originally because this was easier. Then I converted it (see the commit), but realized we still need the channel name and id in various places, so we still need an api call for each test to get that. Hopefully this `/group/search` solution won't break....

e2e/playwright.config.ts
cpoile: Right, good point.

e2e/utils.ts
cpoile: This one of the instances where I tried a few ways to store them ahead of time and I couldn't make it work so I gave up and did this. I figured it's basically the same as starting a new page or making a request with the userStorage context, you know?

e2e/utils.ts
cpoile: I guess I didn't remember? :)

webapp/src/actions.ts
cpoile: That's another case of me spending too long trying to spy the browser and giving up (notifications, sound start, sound stop), plus we'd have to add this sort of thing to the notifications rejected code anyway (this is the test that would have caught the upstream regression we dealt with a few weeks ago). 
It was unsatisfying at first for me too... But then I realized: it's simple, it always works, and it's less prone to breakage.

e2e/global-setup.ts
cpoile: For the `user is OOO: no ringing, no desktop notification` test, automatic replies need to be enabled.

e2e/page.ts
cpoile: And here I thought I was doing well by reducing a ton of lines in the test. 😝 
Open to alternatives (no args object...).

streamer45:
@cpoile I don't think anything was failing other than the recording ones. You need to merge with main.

e2e/utils.ts
streamer45: I suppose that's the convoluted way Playwright stores state anyway so it's fine :)

e2e/global-setup.ts
streamer45: Okay, should it perhaps be a per test setting to enable rather than global?

e2e/page.ts
streamer45: I won't push on this since it's a test but in app we'd go with the object just so you know :)

cpoile:
@streamer45 Thanks -- but also there are some flaky tests that are timing out. Not sure if it's a real timeout, or if the test failed (and timed out looking for an element that wasn't there).

webapp/src/actions.ts
streamer45: This is a bit like mocking I guess. We are essentially putting a limit to the reach of the test. I suppose this would make them pass even if the webapp never actually sent the browser notification.

cpoile: If by `even if the webapp never actually sent the browser notification` you mean the physical desktop notification, yep -- and that's what we would have mocked on the browser anyway (the fact that the `new Notification` was called, not that the test "desktop" showed a notification...). This is basically the same, but better. :)

e2e/global-setup.ts
cpoile: Yep, makes sense. Done.

e2e/page.ts
cpoile: ehhh... Then I probably would remove it. :P

webapp/src/actions.ts
streamer45: While I definitely agree it's simpler from an implementer's point of view, I am struggling to see how this is better from a pure e2e perspective because we don't have any guarantees that the webapp will do its job to the fullest. I think this sort of more advanced cases would be nice to cover with something like [Proxy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) which would essentially let us intercept calls to the native APIs. But I also agree on being pragmatic, maybe something to consider for the future.

e2e/global-setup.ts
streamer45: To note this would likely cause all sorts of racy failures in the future if there's another unrelated test relying on the setting being on/off. That's on me for even suggesting it but I suppose we'll cross that bridge when needed.

cpoile: Right, it's not a very isolated test anymore.

streamer45 (approved):
Great effort :1st_place_medal:

PR #545, opened 2023-10-05
Fix e2e lint issue

cpoile:
#### Summary
- tsconfig/eslint doesn't play nice if you try to include parent or sibling directories, so let's just include the little bit that I was referencing (like we do for other types in the e2e test project).
- also added the e2e linting to the overall checks... Not sure if you want that, but would have prevented this, so I think it's probably a good thing.

#### Ticket Link
- none


streamer45 (approved):
Thanks! Indeed I'd expect these to run on CI.

PR #516, opened 2023-09-08
MM-54272 - Upgrade npm, use modern webapp

cpoile:
#### Summary
- Needed to upgrade a few deps.
- Opening a new PR off main was easiest. Answered questions in the [previous PR](https://github.com/mattermost/mattermost-plugin-calls/pull/507)

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-54272

PR #555, opened 2023-11-01
MM-54272 - Upgrade npm and mattermost-webapp dependency

cpoile:
#### Summary
- Opening a new PR off main was easiest. Answered questions in the [previous PR](https://github.com/mattermost/mattermost-plugin-calls/pull/507)
- And again, replacing: https://github.com/mattermost/mattermost-plugin-calls/pull/516

Note: doesn't include the newer simple imports plugin; all the import reordering here was due to a few of the imports changing and that causing the existing import-order plugin to ask for changes. I tried to confine most of those ordering changes to the second commit. The first commit are the important changes.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-54272


webapp/package.json
```diff
@@ -36,17 +43,16 @@
...
     "babel-loader": "8.2.3",
     "babel-plugin-formatjs": "10.3.7",
     "babel-plugin-styled-components": "2.0.6",
     "babel-plugin-typescript-to-proptypes": "2.0.0",
     "css-loader": "5.2.6",
-    "eslint": "8.33.0",
+    "eslint": "8.49.0",
     "eslint-import-resolver-webpack": "0.13.1",
     "eslint-plugin-formatjs": "4.6.0",
-    "eslint-plugin-import": "2.27.5",
```
cpoile, line 49: I shouldn't have removed this just yet, and yet it's still being pulled into the package-lock, so just leaving as is. Will fix up in the next import order PR (after transcriptions has been rebased and merged into master).

webapp/install_mattermost_webapp.sh
```diff
@@ -0,0 +1,26 @@
...
+
+if [ ! -d .git ]; then
+  git init
+  git remote add -f origin https://github.com/mattermost/mattermost.git
+  git config core.sparseCheckout true
+  echo "webapp/channels\nwebapp/platform/types\nwebapp/platform/client" >> .git/info/sparse-checkout
+fi
+
+git fetch --depth=1 origin master
+git checkout -f $COMMITHASH
```
streamer45, line 19: I am having some troubles with this because adding the remote with the `-f` flag is cloning the whole repository (not ideal) and if simply add remote and shallow clone (depth=1) then fetching the commit doesn't work (obviously).

I am wondering if we could do something like https://stackoverflow.com/a/43136160 instead.

cpoile: @streamer Thanks, can you tell me if this is better?

streamer45: Thanks @cpoile , much better. We could actually retain the sparse checkout you had before to optimize further if we want, I am not opposed to that at all.

webapp/install_mattermost_webapp.sh
```diff
@@ -10,13 +10,12 @@ cd mattermost-webapp
...
-  git config core.sparseCheckout true
-  echo "webapp/channels\nwebapp/platform/types\nwebapp/platform/client" >> .git/info/sparse-checkout
+  git config --local uploadpack.allowReachableSHA1InWant true
+  git remote add origin https://github.com/mattermost/mattermost.git
+  git fetch --depth=1 origin $COMMITHASH
+  git reset --hard FETCH_HEAD
 fi
 
-git fetch --depth=1 origin master
-git checkout -f $COMMITHASH
```
streamer45: I think we still need to cover this case. Keeping fetch + reset out of the if block should hopefully do it.

cpoile: Yah, you're right, missed that. Thanks.

streamer45:
Not sure how worried we should be about this, just a bit annoying at this point:

```
npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE   package: 'eslint-import-resolver-webpack@0.13.1',
npm WARN EBADENGINE   required: {
npm WARN EBADENGINE     node: '^16 || ^15 || ^14 || ^13 || ^12 || ^11 || ^10 || ^9 || ^8 || ^7 || ^6'
npm WARN EBADENGINE   },
npm WARN EBADENGINE   current: { node: 'v18.17.1', npm: '9.6.7' }
npm WARN EBADENGINE }

cpoile:
@streamer45 yep I saw that, I'll try to fix that in the pr for the simpler import ordering.

streamer45 (approved):
Great, it's working very well. Just started an [E2E pipeline](https://git.internal.mattermost.com/qa/calls-e2e-testing/-/pipelines/384331) running the new builder to make sure it works as expected, after which we can merge. Thanks!

streamer45:
@cpoile Looks like there's a stubborn e2e test failing. Not sure if flaky or timing out?

cpoile:
@streamer45 it might be timing out because of load? It looked like the element (join call header button) was visible but it just didn't find it. Trying again.

PR #556, opened 2023-11-03
[MM-55268] Expose metrics through new `ServeMetrics` hook

streamer45:
#### Summary

We implement the `ServeMetrics` hook introduced in https://github.com/mattermost/mattermost/pull/24249. 

@DHaussermann Adding you since we need some basic QA on the related server side PR. So the QA work is the same for both PRs as we'll be using Calls as a test plugin. We just want to verify that metrics will be exposed under http://localhost:8067/plugins/com.mattermost.calls/metrics when running a build from this PR against a MM server with the changes above.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-55268


cpoile (approved):
🎉

DHaussermann (approved):
Tested and passed
I can now see metrics exposed at http://localhost:8067/plugins/com.mattermost.calls/metrics 

Seems to be functional. I can see metrics when the plugin is in use.

LGTM!

PR #561, opened 2023-11-10
MM-55060 - Calls: Add PushTypeCalls for calls-specific push notifications

cpoile:
#### Summary
- This is to support calls-specific push notifications, see also:
  - https://github.com/mattermost/mattermost/pull/25405
  - https://github.com/mattermost/mattermost-push-proxy/pull/110
  - https://github.com/mattermost/mattermost-mobile/pull/7623
- This makes a single check on activation for the push proxy version, and uses that version when checking what type of push notification message to send. This lets us be backward compatible with older push proxies and use correct typing in the future for calls push notification messages, which allows us to internationalize the call started push messages in calls v1.0.
- Also added a `canSendPushNotifications` method, which we probably should have had...

- NOTE: will need to update the MM server dependency once that PR is merged.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-55060


server/simplehttp/client.go
```diff
@@ -0,0 +1,34 @@
...
+	"net/http"
+	"time"
+)
+
+type SimpleClient interface {
+	Do(req *http.Request) (*http.Response, error)
+}
+
+// NewClient creates a SimpleClient intended for one-off requests, like getPushProxyVersion.
+// If we end up needing something more long term, we should consider storing it.
```
streamer45, line 14: Can't we simply use `http.DefaultClient`?

server/utils.go
```diff
@@ -63,6 +66,80 @@ func (p *Plugin) getNotificationNameFormat(userID string) string {
 	return *config.TeamSettings.TeammateNameDisplay
 }
 
+func (p *Plugin) setPushProxyVersion() {
```
streamer45, line 69: Can we re-name this `loadPushProxyVersion`? Since it's more of a get + set.

server/simplehttp/client.go
cpoile: Unless things have changed, it used to be that `http.DefaultClient` shouldn't be used in production (it's missing request timeouts, at least)

streamer45: We could set `http.DefaultClient.Timeout` in a`init()` function. It's just that adding a whole new package and instantiate a new client for a single request seems a bit overkill.

cpoile: K, done, but I'm not totally sure about having it in main (because we're creating mocks right inj). It is simpler though, so good enough for me.

streamer45: I guess the question is, do we really need mocks? It's just an HTTP client, I am thinking using [`httptest`](https://pkg.go.dev/net/http/httptest) would have been sufficient?

server/utils.go
```diff
@@ -176,3 +248,15 @@ func mapKeys[K comparable, V any](m map[K]V) []K {
...
 	return keys
 }
+
+type SimpleClient interface {
+	Do(req *http.Request) (*http.Response, error)
+}
+
+// NewClient creates a SimpleClient intended for one-off requests, like getPushProxyVersion.
+// If we end up needing something more long term, we should consider storing it.
+func NewClient() *http.Client {
```
streamer45, line 258: The reason I mentioned the `init()` function is because this is technically racy as we are modifying a global object. It would probably work fine for our current use case but it may cause problems in the future if anything else relies on `DefaultClient`.

server/simplehttp/client.go
cpoile: Yah, you're right -- fixed, thanks.

server/utils.go
cpoile: When returning a client struct with the timeout set like we are now, I don't think there's still an issue? But let me know and I'll fix it (and I'll learn something).

streamer45: Yep, now it should work fine, just sharing the default transport this way.

streamer45 (approved):
Thanks for bearing with me @cpoile, looks good!

PR #565, opened 2023-11-14
[MM-53432] Improvements

streamer45:
#### Summary

Some minor improvements as requested in the parent PR.

### DIFF
--- a/server/job_service.go
+++ b/server/job_service.go
@@ -8,6 +8,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"strconv"
 	"strings"
 	"time"
 
@@ -302,6 +303,9 @@ func (s *jobService) RunJob(jobType job.Type, callID, postID, jobID, authToken s
 		transcriberConfig.PostID = postID
 		transcriberConfig.TranscriptionID = jobID
 		transcriberConfig.AuthToken = authToken
+		if val := os.Getenv("MM_CALLS_TRANSCRIBER_NUM_THREADS"); val != "" {
+			transcriberConfig.NumThreads, _ = strconv.Atoi(val)
+		}
 
 		jobCfg.Runner = transcriberJobRunner
 		// TODO: we should probably have a dedicated timeout here given
--- a/server/utils.go
+++ b/server/utils.go
@@ -17,6 +17,7 @@ import (
 	"strconv"
 	"strings"
 	"time"
+	"unicode/utf8"
 
 	"github.com/Masterminds/semver"
 )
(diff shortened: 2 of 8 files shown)


server/utils.go
```diff
@@ -176,3 +182,57 @@ func mapKeys[K comparable, V any](m map[K]V) []K {
...
+		for i, u := range users {
+			name += u.GetDisplayName(nameFormat)
+			if i != len(users)-1 {
+				name += "-"
+			}
+		}
+	}
+
+	// Truncating if too long (e.g. group channels)
+	if len(name) > channelNameMaxLength {
```
jupenur, line 231: In the case where `len(name) == channelNameMaxLength + 1` you end up truncating `really long name` to `really long nam…`, and for the truncated string `len(name) == channelNameMaxLength + 3`, which is counter-intuitive, confusing, and potentially harmful.

You could also end up truncating in the middle of a rune, resulting in an invalid UTF-8 encoding in the output, which is definitely incorrect and harmful.

cpoile: Maybe something like this: https://github.com/mattermost/mattermost/blob/bb02b350481a0f111b6f0ad807de674806118a96/server/public/model/link_metadata.go#L47

server/job_service.go
```diff
@@ -302,6 +303,9 @@ func (s *jobService) RunJob(jobType job.Type, callID, postID, jobID, authToken s
 		transcriberConfig.PostID = postID
 		transcriberConfig.TranscriptionID = jobID
 		transcriberConfig.AuthToken = authToken
+		if val := os.Getenv("MM_CALLS_TRANSCRIBER_NUM_THREADS"); val != "" {
+			transcriberConfig.NumThreads, _ = strconv.Atoi(val)
```
cpoile, line 307: Should we be handling the error with a default, or assume it's correct (because its an admin-controlled env variable)? (Just curious about that kind of thing system-wide, @jupenur )

server/utils_test.go
```diff
@@ -145,3 +146,31 @@ func TestCheckMinVersion(t *testing.T) {
...
+		{
+			name:     "spaces",
+			input:    "some file name with spaces.mp4",
+			expected: "some_file_name_with_spaces.mp4",
+		},
+		{
+			name:     "special chars",
+			input:    "somefile*with??special/\\chars.mp4",
+			expected: "somefile_with__special__chars.mp4",
+		},
```
cpoile, line 168: Should we test the too-long case also?

cpoile (approved):

server/job_service.go
jupenur: `calls-transcriber` implements validation and errors on a zero value. In general we should strive to validate all inputs regardless of the level of trust assumed just to avoid mistakes resulting from bad assumptions, and to make a habit of it. That said, there'd be no direct security impact here even if there was no validation.

0/5 on whether we'd want to add some extra error handling here, too, just to make the admin UX better.

streamer45: Yes, this is really there to help me test and/or for very advanced use cases at the moment. Until we surface this through proper config I am okay with keeping it this way as the ultimate validation is done on the other side and when zero (error case) a default is used.

server/utils_test.go
streamer45: Yep, added on the specific function, thanks :+1:

jupenur (approved):

PR #566, opened 2023-11-16
[MM-53432] Model size config setting

streamer45:
#### Summary

This is an important setting as it directly affects quality and performance of the whole process. It will also be useful to run the upcoming performance tests.

#### Screenshot 

image


cpoile (approved):
Nice, thanks!

PR #567, opened 2023-11-16
Add create summary button.

crspeller:
#### Summary
Adds a create summary button when the AI plugin is available. 

The button offloads the logic of what to do afterwards to the AI plugin.

Note that the button positioning varies from the designs because we don't support buttons below the attachments.

Designs: https://mattermost.atlassian.net/wiki/spaces/WD/pages/2510323713/UX+Documentation+for+v1.0#Summarize-Calls-plugin-recordings

image

### DIFF
--- a/webapp/src/components/custom_post_types/post_type_recording.tsx
+++ b/webapp/src/components/custom_post_types/post_type_recording.tsx
@@ -33,15 +33,20 @@ const CreateMeetingSummaryButton = styled.button`
 	justify-content: center;
 	gap: 6px;
 	border-radius: 4px;
-	background: rgba(var(--center-channel-color-rgb), 0.04);
+	background: rgba(var(--center-channel-color-rgb), 0.08);
+    color: rgba(var(--center-channel-color-rgb), 0.64);
+	font-size: 12px;
+	font-weight: 600;
+	line-height: 16px;
 
 	:hover {
-		background: rgba(var(--center-channel-color-rgb), 0.08);
+		background: rgba(var(--center-channel-color-rgb), 0.12);
         color: rgba(var(--center-channel-color-rgb), 0.72);
 	}
 
 	:active {
 		background: rgba(var(--button-bg-rgb), 0.08);
+		color: var(--button-bg);
 	}
 `;
 
--- a/webapp/src/components/custom_post_types/post_type_recording.tsx
+++ b/webapp/src/components/custom_post_types/post_type_recording.tsx
@@ -1,8 +1,73 @@
 import React from 'react';
 import {FormattedMessage} from 'react-intl';
+import {useSelector} from 'react-redux';
+import styled from 'styled-components';
 
-export const PostTypeRecording = () => {
+import {GlobalState} from '@mattermost/types/store';
+import IconAI from 'src/components/icons/ai';
+
+const aiPluginID = 'mattermost-ai';
+
+const useAIAvailable = () => {
+    //@ts-ignore plugins state is a thing
+    return useSelector<GlobalState, boolean>((state) => Boolean(state.plugins?.plugins?.[aiPluginID]));
+};
+
+type CallsPostButtonClickedFunc = (post: any) => void;
+
+const useCallsPostButtonClicked = () => {
+    return useSelector<GlobalState, CallsPostButtonClickedFunc>((state) => {
+        //@ts-ignore plugins state is a thing
+        return state['plugins-' + aiPluginID]?.callsPostButtonClicked;
+    });
+};
+
+const CreateMeetingSummaryButton = styled.button`
+	display: flex;
+	border: none;
+	height: 24px;
+	padding: 4px 10px;
+	margin-top: 8px;
+	margin-bottom: 8px;
+	align-items: center;
+	justify-content: center;
+	gap: 6px;
+	border-radius: 4px;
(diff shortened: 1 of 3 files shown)


webapp/src/components/custom_post_types/post_type_recording.tsx
```diff
@@ -1,8 +1,73 @@
...
+    //@ts-ignore plugins state is a thing
+    return useSelector<GlobalState, boolean>((state) => Boolean(state.plugins?.plugins?.[aiPluginID]));
+};
+
+type CallsPostButtonClickedFunc = (post: any) => void;
+
+const useCallsPostButtonClicked = () => {
+    return useSelector<GlobalState, CallsPostButtonClickedFunc>((state) => {
+        //@ts-ignore plugins state is a thing
+        return state['plugins-' + aiPluginID]?.callsPostButtonClicked;
```
crspeller, line 21: This is supplied by the AI plugin. That way all the logic for what to do when the button is clicked is handled on the AI plugin side.

streamer45: I like it, I think this is the right way of integrating between plugins.

streamer45 (approved):
Looks good :+1:

webapp/src/components/custom_post_types/post_type_recording.tsx
```diff
@@ -1,8 +1,73 @@
...
+
+import {GlobalState} from '@mattermost/types/store';
+import IconAI from 'src/components/icons/ai';
+
+const aiPluginID = 'mattermost-ai';
+
+const useAIAvailable = () => {
+    //@ts-ignore plugins state is a thing
+    return useSelector<GlobalState, boolean>((state) => Boolean(state.plugins?.plugins?.[aiPluginID]));
+};
```
cpoile, line 14: Considering this is a straight lookup, we could add it to the selectors and cut out two functions (which would have to be rerendered on every global state change since react can't tell from the `useSelector` what part of global state we're interested in).

webapp/src/components/custom_post_types/post_type_recording.tsx
cpoile: Same thing here as well. 
But otherwise, totally agree with Claudio -- this is the way to do it!

cpoile (commented):
Awesome! Just commenting about a better way to put the selectors to improve performance.

webapp/src/components/custom_post_types/post_type_recording.tsx
```diff
@@ -1,8 +1,73 @@
...
+	border: none;
+	height: 24px;
+	padding: 4px 10px;
+	margin-top: 8px;
+	margin-bottom: 8px;
+	align-items: center;
+	justify-content: center;
+	gap: 6px;
+	border-radius: 4px;
+	background: rgba(var(--center-channel-color-rgb), 0.04);
```
matthewbirtch, line 36: can we make these adjustments to the default state of the button
```
background: rgba(var(--center-channel-color-rgb), 0.08);
color: rgba(var(--center-channel-color-rgb), 0.64);
font-size: 12px;
font-weight: 600;
line-height: 16px
```

webapp/src/components/custom_post_types/post_type_recording.tsx
```diff
@@ -1,8 +1,73 @@
...
+
+	:hover {
+		background: rgba(var(--center-channel-color-rgb), 0.08);
```
matthewbirtch, line 39: And this adjustment to the hover state:
```
background: rgba(var(--center-channel-color-rgb), 0.08);
```

webapp/src/components/custom_post_types/post_type_recording.tsx
```diff
@@ -1,8 +1,73 @@
...
+        color: rgba(var(--center-channel-color-rgb), 0.72);
+	}
+
+	:active {
+		background: rgba(var(--button-bg-rgb), 0.08);
```
matthewbirtch, line 44: And this adjustment to the active state:
```
color: var(--button-bg);
```

matthewbirtch (changes requested):
Commented inline on a a few minor adjustments to CSS .

For reference, I also updated the [figma file with the states for this button](https://www.figma.com/file/4Dq6Kl0kq8Zf186Ygf4NWT/AI-Plugin?type=design&node-id=869-59543&mode=design&t=D8JNIZsEhMTBbwav-4). Sorry for not including that previously.

crspeller:
@matthewbirtch 

Updated with your changes. 
Normal:
image

Hover:
image

Active:
image

webapp/src/components/custom_post_types/post_type_recording.tsx
crspeller: Not sure what you are suggesting. I think that useSelector will take care of not triggering a re-render if the Boolean result or the function have not changed. Am I reading this wrong? https://react-redux.js.org/api/hooks#useselector

matthewbirtch (approved):
Looks good, thanks @crspeller

webapp/src/components/custom_post_types/post_type_recording.tsx
cpoile: You're right!

streamer45:
@crspeller Feel free to merge as needed.

PR #571, opened 2023-11-22
MM-54743 - Fix: In compact mode, the call card displays inline

cpoile:
#### Summary
- Had to do some trickery to force the display to wrap (because we don't control our custom post's container).
- Doesn't affect non-compact mode, or when a call title is present. If you can think of any other situations I didn't think, please let me know.

| compact, full width | compact, reduced width |
|--|--|
| image | image |

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-54743

### DIFF
--- a/webapp/src/components/custom_post_types/post_type/component.tsx
+++ b/webapp/src/components/custom_post_types/post_type/component.tsx
@@ -33,6 +33,7 @@ interface Props {
     isCloudPaid: boolean,
     maxParticipants: number,
     militaryTime: boolean,
+    compactDisplay: boolean,
 }
 
 const PostType = ({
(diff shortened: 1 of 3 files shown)


webapp/src/components/custom_post_types/post_type/component.tsx
```diff
@@ -136,6 +138,8 @@ const PostType = ({
         );
     }
 
+    const compactTitle = compactDisplay ? '\u200b' : '';
```
streamer45, line 141: Any reason for not using something more semantically defined like the `<br>` tag?

streamer45 (approved):

matthewbirtch (changes requested):
@cpoile this looks good, but I did notice an issue in the RHS in compact mode. There seems to be extra space added that is not there in standard display mode:

**Compact display mode**
Notice the extra space between the username and the call card
image

**Standard display mode**
image

Other than that, this should be good.

cpoile:
Fixed now, thanks for pointing that out Matt!

matthewbirtch (approved):
Looks good. Thanks Chris!

PR #574, opened 2023-11-23
MM-54883 - Surface who is Host more clearly (e.g., when screensharing)

cpoile:
#### Summary
- Add the Host badge to the participants list on the widget and the expanded view
- Add ellipsis to long names (this was a bit tricky, happy we could find a way)

| pre widget | post widget | pre widget with emoji | post widget with emoji |
|--|--|--|--|
| image | image | image | image |

| pre widget screensharing + emoji | post widget screensharing + emoji | pre expanded view ss + emoji | post expanded view ss + emoji | 
|--|--|--|--|
| image | image | image | image | 

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-54883
- Figma: https://www.figma.com/file/cSzDHTojGE4vqGxajyWZd4/MM-54883-Host-Badge-in-Participants-List?type=design&node-id=1-7&mode=design


cpoile:
note: I'll fix the snapshots once we have the design approved.

matthewbirtch (changes requested):
Thanks for this @cpoile. It's looking great, but I do have a few questions/requests:

1. I'm not seeing the `HOST` badge displaying in the desktop app widget, but I'm seeing it in the browser. Will that be a separate PR? I assume that needs to be done separately.

2. Are we able to reduce the space between the Username, (you) and `HOST` badge (in both the widget view and expanded view)? The gap looks big to me and it probably only needs to be 4 px or so. To accomplish this you might need to group together the username, (you) label and `HOST` badge in a div tag so that this spacing adjustment doesn't affect the other spacing in the row. Also it looks like the (you) element has some extra space added on the left side of it. I think it needs to have its own `<span>` tag.
    Screenshot 2023-11-24 at 9 21 15 AM

3. In the expanded view participant list, I'm seeing a big gap between the name and the other you/host elements if the name is short. Is the anything we can do about that so the (you) and `HOST` labels display more inline with the username and close that big gap?
    Screenshot 2023-11-24 at 9 21 40 AM

    Ideally, that would look more like this:
    image

4. In the expanded participants list, can we adjust the color of the (you) label so that it matches the color that is used in the widget for this label?

I know some of this might be tricky, especially with the truncation, so let me know what you think.

matthewbirtch:
> I also realized I had the `Host` and `You` backwards compared to the designs, so fixed that

Sorry, @cpoile I think you had it right in the first place. This was a mistake in the original design which I didn't pick up on. Sorry for the back-and-forth on that. The `(you)` label should be first, then the `HOST` tag. When you make these switch back you may need to adjust spacing between the `(you)` and `HOST` badge again when they both appear.

> The too-wide right margin on the ellipsis is a known problem with browsers

Ok, we can live with that as-is then. Not much we can do if there is a known issue with browsers.

> Re: the desktop widget, that does have the Host badge, maybe refresh?

I'm still not seeing these updates on the desktop app widget. This is what I see (even after a clear cache and reload.
image

cpoile:
@matthewbirtch Ok, reverted back to `(you) host`. As for desktop, I don't know what's wrong, it's working the way it should for me (the component for the widget and participant list is the same, whether you're on webapp or desktop). So if you're not seeing it, it means the desktop is not using the updated plugin code. Are you sure you're on a server that has this PR's plugin code?

Here is a screenshot of what it looks like now, from my desktop app:
image

matthewbirtch:
> Are you sure you're on a server that has this PR's plugin code?

Hmmm, I'm running everything locally. Server is on the `master` branch, using this PR branch for the calls plugin. Do I need to use a different branch for the server?

cpoile:
@matthewbirtch Nope, I'm on server's master branch as well. I don't know why you're not seeing it; like I said, both webapp and desktop use the same component to show the new `(you) host`, so if you see it on one, you should see it on the other. You've done a hard refresh on desktop?

matthewbirtch:
> @matthewbirtch Nope, I'm on server's master branch as well. I don't know why you're not seeing it; like I said, both webapp and desktop use the same component to show the new `(you) host`, so if you see it on one, you should see it on the other. You've done a hard refresh on desktop?

Ya, I've cleared my cache and reloaded the desktop app a few times and still don't see these changes in the widget. Maybe this is a non-issue and something weird with my local setup. I don't want to block if that's the case

matthewbirtch (approved):
As discussed in our DM, I tested this out on the calls test server in the desktop app and it's working perfectly there. I won't block this as it seems to be just a local caching thing.

All looks good to me

streamer45:
@cpoile Does this need mobile equivalent changes or is it already up to spec on that side?

streamer45 (commented):
A bit OCD I admit but somehow the label is not looking equally spaced when the current user is the host.

image

image

cpoile:
@streamer45 I'll make a new ticket for that, as it needs UX review and decisions: https://mattermost.atlassian.net/browse/MM-56368

cpoile:
@streamer45 Not sure what you mean by looking at the screencaps (my OCD is pretty decent though ;) ). Could you point it out to me?

streamer45:
> @streamer45 Not sure what you mean by looking at the screencaps (my OCD is pretty decent though ;) ). Could you point it out to me?

Gap between the username and `(you)` is double of that between `(you)` and the host label.

streamer45 (approved):
Looks great now, thanks!

cpoile:
@streamer45 Ah, I see. Good eye. Fixed:
image
image

The spacing between the ellipsis and the `(you)` will still be a bit bigger bc of the ellipsis spacing issue I mentioned above. We could fix that by detecting when the ellipsis will probably show by doing a render/measure/rerender trick, if you want to go that far.

streamer45:
All good from my side :+1:

PR #576, opened 2023-11-28
Don't display AI button when plugin doesn't support it

crspeller:
#### Summary
The summarization button is showing up on community, but it does not work since the AI plugin version uploaded there does not support it yet. 

This fixes the issue by checking for the presence of the button clicked function before displaying the button.


streamer45 (approved):

cpoile (approved):
👍

streamer45:
> @streamer45 The e2e failures look unrelated. Maybe flaky tests? Not sure how to re-run those tests.

I don't think they are flaky. We probably need to merge https://github.com/mattermost/mattermost-plugin-calls/pull/549 first.

PR #577, opened 2023-11-29
[MM-55975] Fix regression when presenting session leaves call without previously…

streamer45:
#### Summary

Multi session support caused a regression in the case of a presenter leaving the call without previously unsharing. This would cause the state to be in somewhat of a stuck state, preventing anyone else from sharing. The UI wasn't affected by this bug because the [check](https://github.com/mattermost/mattermost-plugin-calls/blob/53ab625c48e1c0a01290dd75523af6845c18ee2e/webapp/src/reducers.ts#L668-L674) on the redux side is correctly against the session id of user leaving.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-55975

### DIFF
--- a/e2e/tests/media.spec.ts
+++ b/e2e/tests/media.spec.ts
@@ -80,6 +80,54 @@ test.describe('screen sharing', () => {
         await devPage.leaveCall();
         await userPage.leaveCall();
     });
+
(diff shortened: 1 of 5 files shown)


cpoile (approved):
Nice, and good test!

PR #581, opened 2023-12-04
Set a captions_file_id post prop to avoid indirection when fetching captions file for video preview

streamer45:
#### Summary

Adding some logic to save the VTT captions file id into the recording post props so it can be access directly.

I am keeping everything else the same in the call post props even if slightly redundant as this "centralized" data may still be useful if we ever want to retroactively populate calls info whenever we migrate to a proper SQL schema.


cpoile (approved):
Awesome!

PR #583, opened 2023-12-05
[MM-56054] Fix potential races when accessing or updating call post

streamer45:
#### Summary

We need to take some extra measures to avoid concurrently reading/writing the call post or we can get into an inconsistent state. This is done at two levels:

- Using the global lock on call to synchronize read/writes for a given call post.
- Implementing a dedicated `GetPost` store method that can fetch from the writer db node since `p.GetPost` can suffer from replica lag.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-56054

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,6 +13,7 @@ require (
 require (
 	github.com/Masterminds/semver v1.5.0
 	github.com/gorilla/websocket v1.5.0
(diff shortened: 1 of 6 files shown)


cpoile (approved):
Makes sense, thanks!

PR #585, opened 2023-12-07
MM-56014 - Fix type error

cpoile:
#### Summary
- Fixing a mistake I made in one of the selectors. Oops, apologies.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-56014


streamer45 (approved):
Nice! :yay:

cpoile:
@streamer45 Oh, but I don't think this explains the original "unmuting->cannot hear anyone from that point on" issue...

streamer45:
> @streamer45 Oh, but I don't think this explains the original "unmuting->cannot hear anyone from that point on" issue...

Right, let's fix the type error here and I'll follow up with the reporter to figure out the rest.

PR #587, opened 2023-12-09
Improve E2E tests

streamer45:
#### Summary

PR does a general review of our e2e tests performance, especially focusing on eliminating most artificial waiting and simplifying a few checks for speed.

Results are decent from a tests duration perspective, decreasing the time of execution from ~38 to ~18 minutes. 

We still have a good amount of of flakyness though, mostly concentrated in notifications tests. Will need to take a closer look as at times clients are actually failing to connect so there could be some deeper underlying issues. Maybe the instances would benefit a bump in specs :)


cpoile (approved):
Awesome, I'll look at the notification tests and see if something similar can be done with them.

PR #591, opened 2023-12-12
[MM-56178] Fix failure to update call post (MySQL)

streamer45:
#### Summary

Looks like MySQL needs a special no-op mapper function in order to correctly translate columns into struct fields.

I wish we could add a test for this but we need some work to make proper integrations tests that run against both db engines. I'll take care of it though. 

For now I manually tested this on both db drivers and it's looking good.
 
#### Ticket Link

https://mattermost.atlassian.net/browse/MM-56178


cpoile (approved):
Thank you

PR #592, opened 2023-12-13
[MM-56183] Handle bot clients ws reconnects

streamer45:
#### Summary

During a transcribing job the ws connection briefly dropped, the client immediately reconnected without issues but our server side logic didn't account for this to happen. Upon stopping the job we'd be sending the fake `call_end` event to the outdated connection ID which would cause the client to never receive it and the job to continue indefinitely.

PR addresses this by handling this special case without introducing breaking changes. Upon reconnect we update the bot connection id in the state.

A better approach would be to send an explicit event with the job ID included in the payload so that there's no confusion no matter how many times the bot client reconnected. I am planning to do this in https://mattermost.atlassian.net/browse/MM-56187.

Will need to do some manual testing on HA environment prior to merging.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-56183


server/plugin.go
```diff
@@ -65,7 +65,8 @@ type Plugin struct {
 
 	// A map of callID -> *cluster.Mutex to guarantee atomicity of call state
 	// operations.
-	callsClusterLocks map[string]*cluster.Mutex
+	callsClusterLocks    map[string]*cluster.Mutex
+	callsClusterLocksMut sync.RWMutex
```
cpoile, line 69: We're using a mut for a map of mutexes... 😂

cpoile (approved):

server/plugin.go
streamer45: indeed, better than deadlocking which was the alternative :)

streamer45:
@cpoile Asking for a second review as last commit was significant :sweat_smile: 

While not in love with mocked tests, I found a bug while doing manual testing that a properly mocked unit test would have caught. So added some boilerplate to make it work. Hopefully it will also make it easier to test more stuff going forward and bump our repo stats as a bonus :)

server/websocket_test.go
```diff
@@ -0,0 +1,228 @@
...
+			}).Once()
+
+			state.Call.Transcription.BotConnID = "newTranscribingBotConnID"
+			expectedStateJSON, err := json.Marshal(state)
+			require.NoError(t, err)
+
+			mockAPI.On("KVSet", "channelID", expectedStateJSON).Return(nil).Once()
+
+			err = p.handleBotWSReconnect("newTranscribingBotConnID", "prevTranscribingBotConnID", "originalConnID", channelID)
+			require.NoError(t, err)
```
cpoile, line 225: Could you explain a bit about what this is testing? It's a little unclear to me looking from the outside.

cpoile (commented):
Just a clarifying question

server/websocket_test.go
streamer45: The most important part is the `KVSet` mocked call. We are essentially asserting the state mutation is what's expected given the initial `state` and the `handleBotWSReconnect` method's inputs. Let me add a couple of comments.

streamer45: Done

cpoile (approved):
Ok, that clearer, thank you!

PR #597, opened 2023-12-15
drop darwin from official manifest, build local platform for devs

lieut-data:
#### Summary
The plugin.json manifest has been "lying", in that it lists support for MacOS (`darwin`) even though we don't support or distribute that. It's necessary to allow developers to work locally.

Fix this by detecting if `MM_SERVICESETTINGS_ENABLEDEVELOPER` is set as truthy in the environment, and dynamically rewriting the manifest to clear the `Executables` and instead the platform-agnostic `Executable`. The server won't know what platform is supported, but will assume it's the right one and "just run it", working fine in development while leaving CI and production builds unchanged.

Then unlist darwin from the canonical manifest.

This complements https://github.com/mattermost/matterbuild/pull/97.

#### Ticket Link
None.


cpoile (approved):
If we're no longer including darwin in the manifest, there's no point building them in the makefile for non-`"MM_SERVICESETTINGS_ENABLEDEVELOPER` situations, right? Probably we should remove those lines. (I can do that for you if you want)

Otherwise, this should work fine, and clean up the manifest. 🎉

lieut-data:
Oh, yes! Weird, I didn't even think to check there. If you could help remove, I'd be grateful, @cpoile!

cpoile:
Yep, done!

streamer45 (approved):
Thank you :+1:

PR #601, opened 2023-12-21
[MM-56048] Make access to global window properties safer

streamer45:
#### Summary

PR adds some try/catch when accessing properties on the global `window` object to avoid failing with an uncaught error in the unlikely case the object not being accessible/defined (e.g. cross-origin like the bug report).

While reproducing this issue, I realized that the `channelIDForCurrentCall` selector was getting called hundreds of times on channel load. We are now using `createSelector` which should cache the result which seems to be drastically reducing the number of calls. Please let me know if it makes sense.

**Before**

image

**After:**

image

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-56048

### DIFF
--- a/webapp/src/utils.ts
+++ b/webapp/src/utils.ts
@@ -86,6 +86,18 @@ export function getCallsClient(): CallsClient | undefined {
     return callsClient;
 }
 
+export function getCallsClientChannelID(): string {
+    return getCallsClient()?.channelID || '';
+}
+
+export function getCallsClientSessionID(): string {
+    return getCallsClient()?.getSessionID() || '';
+}
+
+export function getCallsClientInitTime(): number {
+    return getCallsClient()?.initTime || 0;
+}
+
 export function shouldRenderCallsIncoming() {
     try {
         const win = window.opener ? window.opener : window;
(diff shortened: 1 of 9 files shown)


streamer45:
Memoizing the client object was a bit too extreme given we add certain properties after creation. I added some getter utils so that we cache the actual values we need rather than the whole thing.

webapp/src/utils.ts
```diff
@@ -77,18 +77,41 @@ export function getChannelURL(state: GlobalState, channel: Channel, teamId: stri
...
+        const nonChannels = window.location.pathname.startsWith('/boards') || window.location.pathname.startsWith('/playbooks') || window.location.pathname.includes(`${pluginId}/expanded/`);
+        if (win.desktop && nonChannels) {
         // don't render when we're in desktop, or in boards or playbooks, or in the expanded view.
         // (can be simplified, but this is clearer)
+            return false;
+        }
+        return true;
+    } catch (err) {
+        logErr(err);
         return false;
```
cpoile, line 113: Should we default to returning true like before?

cpoile (approved):
Looks good, nice improvements thank you!

webapp/src/utils.ts
streamer45: Would that make a difference? With the try block in place I am not sure that would be a possible code path, hence why the linter isn't complaining I guess.

PR #606, opened 2024-01-02
[MM-56055] Improve access to Calls posts props

streamer45:
#### Summary

PR improves access to props in Calls posts to avoid getting JS errors and blank pages if these are somehow missing (e.g. patching a post manually through API).

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-56055


webapp/src/utils.ts
```diff
@@ -510,3 +511,25 @@ export function notificationsStopRinging() {
         window.e2eNotificationsSoundStoppedAt.push(Date.now());
     }
 }
+
+export function getCallPropsFromPost(post: Post): CallPostProps {
+    return {
+        title: post.props?.title,
+        start_at: post.props?.start_at,
+        end_at: post.props?.end_at,
```
cpoile, line 519: It occurred to me that these maybe should have been namespaced (something like `post.props.call.title`, etc) to prevent being clobbered by another plugin thinking "props.title seems like a good place to put something"... But maybe that's being too paranoid.

cpoile (approved):

webapp/src/utils.ts
streamer45: Technically they are scoped by the custom post type. Unless a plugin starts messing with calls posts but that's likely at their own peril.

PR #609, opened 2024-01-04
[MM-53455] Broadcast certain call events to participants only

streamer45:
#### Summary

The original idea was to minimize websocket events we are sending out since most of them are broadcast to the entire channel even though it's likely that only a subset of members are in the call and plenty of these events are unneeded by clients not connected to the call (e.g. mute/unmute and other session state changes).

However, to optimize this we need to have the list of participants which in certain cases means an additional call to the database to fetch the call state. Also `model.WebsocketBroadcast` doesn't support a slice of user ids so we need to loop on `PublishWebSocketEvent` from the plugin side. 

Even though the reasoning behind this PR makes total sense from a logical perspective (avoid sending out unnecessary data), the above practical reasons make me doubt a little on whether the changes are actually worth the tradeoffs from a performance standpoint. Will need to load-test a bit and how this behaves at scale.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-53455


cpoile (approved):
Should be interesting to see the numbers; code looks good.

streamer45:
@cpoile I've pretty much redone this from scratch to work on top of the new changes, will need a new review.

streamer45:
@cpoile Just a reminder this is ready for re-review.

cpoile (approved):
Nice, thank you! I guess the only question I would have is: if we're worried about the performance implications, would it be better to add an option to the server's own method to broadcast to a slice of userIDs?  (in the future of course)

streamer45:
> Nice, thank you! I guess the only question I would have is: if we're worried about the performance implications, would it be better to add an option to the server's own method to broadcast to a slice of userIDs? (in the future of course)

Yes, ideally we'd do that to avoid some of the penalty of having to loop through RPC calls.

PR #610, opened 2024-01-05
[MM-56187] Implement explicit event to stop jobs

streamer45:
#### Summary

PR implements an explicit `job_stop` websocket event to signal bots they need to exit without needing to track connection ids (see https://github.com/mattermost/mattermost-plugin-calls/pull/592 for context).

We currently keep backwards compatibility by leaving the existing (now deprecated) code relying on the call end event, although since the breaking change only affects the transcriber job we may decide to rip it out sooner rather than later.

#### Related PRs

https://github.com/mattermost/calls-common/pull/24
https://github.com/mattermost/rtcd/pull/127
https://github.com/mattermost/calls-transcriber/pull/16

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-56187


webapp/src/types/types.ts
```diff
@@ -180,3 +180,8 @@ export type RealNewPostMessageProps = {
     set_online: boolean;
     team_id: string;
 }
+
+export type StopJobData = {
+    channelID: string;
+    jobID: string;
+};
```
cpoile, line 187: Remind me why we're duplicating the type here?

cpoile (approved):

webapp/src/types/types.ts
streamer45: No reason, that should have been removed. I initially implemented it on this side only but then moved it to `calls-common` since it's a public client event and also removed the redundant channel ID which is not necessary.

PR #613, opened 2024-01-10
[MM-46339] Handle user leaving channel with active call

streamer45:
#### Summary

PR implements the logic to handle the case of a user leaving the channel while in a call. Since they are effectively losing permissions to the call, we immediately disconnect the user and show a dialog so that they are aware of what happened (see screenshots).

Note: we may want to port the client-side changes to mobile to be consistent.

#### Screenshots

**User left**

image

**User was removed**

image

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-46339


webapp/src/components/call_error_modal.tsx
```diff
@@ -125,6 +127,26 @@ export const CallErrorModal = (props: Props) => {
...
+    case userLeftChannelErr.message:
+        headerMsg = (
+            <span>{formatMessage({defaultMessage: 'You left the channel'})}</span>
+        );
+        msg = (
+            <span>
+                {formatMessage({defaultMessage: 'You were disconnected from the call as you left the channel.'})}
+            </span>
+        );
+        break;
```
streamer45, line 149: @cwarnermm This is the part I am asking for editor review as we are adding a couple of new strings. Thanks!

server/plugin.go
```diff
@@ -390,3 +390,40 @@ func (p *Plugin) MessageWillBeUpdated(c *plugin.Context, newPost, oldPost *model
...
+		p.LogWarn("UserHasLeftChannel: unexpected nil channel member")
+		return
+	}
+
+	state, err := p.kvGetChannelState(cm.ChannelId, false)
+	if err != nil {
+		p.LogError("UserHasLeftChannel: failed to get call state", "err", err.Error(), "channelID", cm.ChannelId)
+		return
+	} else if state == nil || state.Call == nil {
+		p.LogDebug("UserHasLeftChannel: no call ongoing", "channelID", cm.ChannelId)
```
cpoile, line 405: Do we need this? I'm worried it will spam the logs...

webapp/src/websocket_handlers.ts
```diff
@@ -363,3 +368,14 @@ export function handleUserDismissedNotification(store: Store, ev: WebSocketMessa
...
 }
+
+export function handleUserRemovedFromChannel(store: Store, ev: WebSocketMessage<UserRemovedData>) {
+    const channelID = ev.data.channel_id || ev.broadcast.channel_id;
+    const currentUserID = getCurrentUserId(store.getState());
+    const removedUserID = ev.data.user_id || ev.broadcast.user_id;
+    const removerUserID = ev.data.remover_id;
+
+    if (removedUserID === currentUserID && channelID === channelIDForCurrentCall(store.getState())) {
+        getCallsClient()?.disconnect(removerUserID === currentUserID ? userLeftChannelErr : userRemovedFromChannelErr);
```
cpoile, line 379: nice and simple

cpoile (approved):

server/plugin.go
streamer45: That's the question, we need it when something suddenly breaks and we are clueless :)

I think that's what debug level logs are in my mind. It's a spam which you should be able to filter. It's more the times I had to debug something in production and wished I had these sorts of logs than the times I suffered through spam. But I agree with you, likely nothing too critical happening here so can probably be removed.

cwarnermm (commented):
I propose the following copy updates visible to end users:

**You left the channel**
You have left the channel, and have been disconnected from the call.
[OK]

**You were removed from channel**
You have been removed from the channel, and have been disconnected from the call.
[OK]

esarafianou (approved):

abhijit-singh (commented):
Thanks @streamer45, looks good! I was wondering if we can make the styling consistent with other modals. We're moving towards left aligned text and right aligned buttons in all the modals. Here's a figma link for reference - [Figma file](https://www.figma.com/file/G1xHnuLDxW4YPacSY4BW08/Calls%3A-UX-Nits?type=design&node-id=1528-138216&mode=design&t=9nsOMlE0IIFHl9os-4)
image

streamer45:
Thanks @abhijit-singh! Should be improved now:

image

image

abhijit-singh (approved):
Looks good, thanks @streamer45!

PR #614, opened 2024-01-12
Make our load-test client speak :speaking_head: :magic_wand: 

streamer45:
#### Summary

PR adds a couple of functionalities to our load-test client:

- A new `-speech-file` flag to pass any valid audio Opus encoded file. I am including two generated speech samples with their text source. 
  - This could be useful to measure transcription quality and performance of well known samples.
- A way to generate speech from text on the fly and essentially script conversations. See [lt/cmd/speech/main.go](https://github.com/mattermost/mattermost-plugin-calls/blob/lt-speech/lt/cmd/speech/main.go).
  - This allows for a more flexible way to test realistic conversations between multiple users.

Had to restructure things a bit to make this work so updating the docs as well.


streamer45:
Made `./lt` a separate Golang module because it actually didn't make sense to share dependencies with the plugin, even more so now that we have some CGO stuff in there (opus).

cpoile (approved):
Approving because it works and I've made enough changes/additions on my side that to make any here would cause some bad conflicts.
Thank you again for this, truly helpful for the live captions testing.

streamer45:
@cpoile It's the flaky one again, will merge.

PR #620, opened 2024-01-16
[MM-54359] Implement session validation check

streamer45:
#### Summary

PR implements a session validation check to forcefully disconnect any calls client that has either their session revoked or expired.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-54359

### DIFF
--- a/server/websocket.go
+++ b/server/websocket.go
@@ -412,7 +412,7 @@ func (p *Plugin) wsReader(us *session, authSessionID, handlerID string) {
 				continue
 			}
 
-			if s, appErr := p.API.GetSession(authSessionID); appErr != nil || time.Now().UnixMilli() >= s.ExpiresAt {
+			if s, appErr := p.API.GetSession(authSessionID); appErr != nil || (s.ExpiresAt != 0 && time.Now().UnixMilli() >= s.ExpiresAt) {
 				fields := []any{
 					"channelID",
 					us.channelID,
--- a/server/websocket.go
+++ b/server/websocket.go
@@ -48,6 +48,10 @@ const (
 	wsReconnectionTimeout            = 10 * time.Second
 )
 
+var (
+	sessionAuthCheckInterval = 10 * time.Second
+)
+
 type CallsClientJoinData struct {
 	ChannelID string
 	Title     string
@@ -379,7 +383,10 @@ func (p *Plugin) OnWebSocketDisconnect(connID, userID string) {
 	}
 }
 
-func (p *Plugin) wsReader(us *session, handlerID string) {
+func (p *Plugin) wsReader(us *session, authSessionID, handlerID string) {
+	sessionAuthTicker := time.NewTicker(sessionAuthCheckInterval)
+	defer sessionAuthTicker.Stop()
+
 	for {
 		select {
 		case msg, ok := <-us.wsMsgCh:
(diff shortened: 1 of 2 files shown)


server/websocket.go
```diff
@@ -48,6 +48,10 @@ const (
 	wsReconnectionTimeout            = 10 * time.Second
 )
 
+var (
+	sessionAuthCheckInterval = 10 * time.Second
```
streamer45, line 52: @esarafianou For performance reasons I wouldn't go lower than this, let me know if it's reasonable.

server/websocket.go
```diff
@@ -397,6 +404,39 @@ func (p *Plugin) wsReader(us *session, handlerID string) {
...
 			return
+		case <-sessionAuthTicker.C:
+			// Server versions prior to MM v9.5 won't have the session ID set so we
+			// cannot go ahead with this check.
+			// Should be removed as soon as we bump the minimum supported version.
+			if authSessionID == "" {
+				continue
+			}
+
+			if s, appErr := p.API.GetSession(authSessionID); appErr != nil || time.Now().UnixMilli() >= s.ExpiresAt {
```
cpoile, line 415: Do you have a sense of what kind of pressure this is going to put on the system? 
Is there a batched version of getSession?

cpoile (approved):
Thanks. Just a non-blocking question.

server/websocket.go
streamer45: > Do you have a sense of what kind of pressure this is going to put on the system?

Performance should be fairly good since we cache sessions so normally that call shouldn't be hitting the database but yeah that's a concern I had as well. 

> Is there a batched version of getSession?

I don't believe so, definitely not exposed to the plugin side anyway.

streamer45: @cpoile To add some context, in general I felt this to be the safest approach because otherwise we'd have to implement a plugin hook that triggers on session invalidation which can definitely improve performance but it's also much easier to mess up security wise.

If performance ever becomes a problem I'd probably go down that route.

streamer45:
Sent a fix e2e tests caught. We should also support session with a zero `ExpiresAt` which means no expiration.

server/websocket.go
esarafianou: Totally reasonable.

server/websocket.go
```diff
@@ -397,6 +404,39 @@ func (p *Plugin) wsReader(us *session, handlerID string) {
...
 			return
+		case <-sessionAuthTicker.C:
+			// Server versions prior to MM v9.5 won't have the session ID set so we
+			// cannot go ahead with this check.
+			// Should be removed as soon as we bump the minimum supported version.
+			if authSessionID == "" {
+				continue
+			}
+
+			if s, appErr := p.API.GetSession(authSessionID); appErr != nil || (s.ExpiresAt != 0 && time.Now().UnixMilli() >= s.ExpiresAt) {
```
esarafianou, line 415: When is the session's `ExpiresAt` set to 0 that we need to take into account here?

esarafianou: I just saw your comment above that _We should also support sessions with a zero ExpiresAt which means no expiration._

A bit irrelevant to the PR, but when is there no session expiration?

server/websocket.go
```diff
@@ -379,7 +383,10 @@ func (p *Plugin) OnWebSocketDisconnect(connID, userID string) {
 	}
 }
 
-func (p *Plugin) wsReader(us *session, handlerID string) {
+func (p *Plugin) wsReader(us *session, authSessionID, handlerID string) {
```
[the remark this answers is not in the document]
streamer45: Indeed, that's the one I mentioned in the internal thread: https://github.com/mattermost/mattermost/pull/25928

server/websocket.go
esarafianou: Thanks for explaining!

esarafianou (approved):
LGTM!

PR #624, opened 2024-01-18
Move AI summarization button to the calls transcription post. Use the generated transcription.

crspeller:
#### Summary

Moves the AI summarization button to the to the call transcription post. 

image

When used the button will now use the WebVTT generated transcript "attached" to that post.

As before, the versions of the calls and AI plugin must both support this for the button to show up.

### DIFF
--- a/server/bot_api.go
+++ b/server/bot_api.go
@@ -347,7 +347,7 @@ func (p *Plugin) handleBotPostTranscriptions(w http.ResponseWriter, r *http.Requ
 	}
 	transcriptionPost.AddProp("call_post_id", info.PostID)
 	transcriptionPost.AddProp("transcription_id", info.JobID)
-	transcriptionPost.AddProp("web_vtt_file_id", info.Transcriptions[0].FileIDs[0])
+	transcriptionPost.AddProp("captions", info.Transcriptions.ToClientCaptions())
 	trPost, appErr := p.API.CreatePost(transcriptionPost)
 	if appErr != nil {
 		res.Err = "failed to create post: " + appErr.Error()
--- a/server/bot_api.go
+++ b/server/bot_api.go
@@ -341,12 +341,13 @@ func (p *Plugin) handleBotPostTranscriptions(w http.ResponseWriter, r *http.Requ
 		UserId:    p.getBotID(),
 		ChannelId: callID,
 		Message:   postMsg,
-		Type:      "custom_calls_transcription",
+		Type:      callTranscriptionType,
 		RootId:    threadID,
 		FileIds:   []string{info.Transcriptions[0].FileIDs[1]},
 	}
 	transcriptionPost.AddProp("call_post_id", info.PostID)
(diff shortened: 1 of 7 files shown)


webapp/src/components/custom_post_types/post_type_transcription.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+export const PostTypeTranscription = (props: Props) => {
+    const aiAvailable = useAIAvailable();
+    const callsPostButtonClicked = useCallsPostButtonClicked();
+
+    const createMeetingSummary = () => {
+        callsPostButtonClicked?.(props.post);
+    };
+
+    const msg = props.post.message;
```
crspeller, line 64: Did this directly to preserve the titling behavior from the server. This should probably be translated like the other one though.

server/bot_api.go
```diff
@@ -341,12 +341,13 @@ func (p *Plugin) handleBotPostTranscriptions(w http.ResponseWriter, r *http.Requ
...
 		ChannelId: callID,
 		Message:   postMsg,
-		Type:      "custom_calls_transcription",
+		Type:      callTranscriptionType,
 		RootId:    threadID,
 		FileIds:   []string{info.Transcriptions[0].FileIDs[1]},
 	}
 	transcriptionPost.AddProp("call_post_id", info.PostID)
 	transcriptionPost.AddProp("transcription_id", info.JobID)
+	transcriptionPost.AddProp("web_vtt_file_id", info.Transcriptions[0].FileIDs[0])
```
crspeller, line 350: Added because it was easier than hunting it down with the other props. 0/5 if there is a another preferred way.

streamer45: Is there a fundamental difference in whether we feed to the model the VTT captions file vs the text format? Data should be pretty much the same. Is the former easier to deal with prompting wise perhaps?

crspeller: I have some code that parses the VTT from whisper and formats it better for the LLM. I didn't want to have to parse the custom text format to replicate.

streamer45: Fair enough. In that case would you mind adding the whole `captions` prop instead, same as we do for the recording post?

crspeller: @streamer45 Done. I guess the plan with that prop is to support captions in multiple languages?

streamer45: Thanks. And yes eventually we could be doing automated translations so I kept it open to that. For now it will be just one, in whatever language was detected while transcribing.

streamer45 (approved):
Looks good! :tada:

cpoile (approved):
Nice!

streamer45:
@crspeller Looks like the custom transcription post component may have broken an [e2e test expectation](https://mattermost-cypress-report.s3.us-east-1.amazonaws.com/calls-pw/2024_01_19__22_04_30/index.html#?testId=8e65504a01e80415eb14-d0028768af8c05fea30a). Could you have a look?

PR #632, opened 2024-01-25
Script-driven tts loadtesting, using Polly

cpoile:
#### Summary
- Added script-driven load testing
- Uses AWS Polly, which assumes you have your credentials in `~/.aws/credentials` and `~/.aws/config`. Needs to use us-east-1 or another that has the neural tts.
- Still wip, needs documentation updates.

### DIFF
--- a/lt/client/client.go
+++ b/lt/client/client.go
@@ -445,8 +445,6 @@ func (u *User) transmitSpeech() {
 				var err error
 				if u.pollySession != nil {
 					rd, rate, err = u.pollyToSpeech(text)
-				} else {
-					rd, rate, err = textToSpeech(text)
 				}
 				if err != nil {
 					log.Printf("%s: textToSpeech failed: %s", u.cfg.Username, err.Error())
--- a/lt/client/client.go
+++ b/lt/client/client.go
@@ -84,7 +84,7 @@ type Config struct {
 	Setup         bool
 	SpeechFile    string
 	PollySession  *polly.Polly
-	PollyVoiceId  *string
+	PollyVoiceID  *string
 }
 
 type User struct {
@@ -100,7 +100,7 @@ type User struct {
 	isHost      bool
 
 	pollySession   *polly.Polly
-	pollyVoiceId   *string
+	pollyVoiceID   *string
 	speechTextCh   chan string
 	doneSpeakingCh chan struct{}
 
@@ -127,7 +127,7 @@ func NewUser(cfg Config) *User {
 		speechTextCh:   make(chan string, 8),
 		doneSpeakingCh: make(chan struct{}),
 		pollySession:   cfg.PollySession,
-		pollyVoiceId:   cfg.PollyVoiceId,
+		pollyVoiceID:   cfg.PollyVoiceID,
 	}
 }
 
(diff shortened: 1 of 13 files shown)


lt/cmd/speech/main.go
```diff
@@ -83,3 +112,84 @@ func main() {
...
+			defer userWg.Done()
+			if err := user.Connect(stopCh); err != nil {
+				log.Fatalf("connectUser failed: %s", err.Error())
+			}
+		}()
+	}
+
+	// "Conversation" logic
+	go func() {
+		time.Sleep(2 * time.Second) // time to take a sip of coffee before we talk talk talk
```
streamer45, line 163: Can we also send unmute events? Otherwise it feels a bit weird to hear sound when the UI is not matching.

lt/client/client.go
```diff
@@ -423,57 +434,72 @@ func (u *User) transmitSpeech() {
...
 
-			audioSamplesDataBuf := bytes.NewBuffer([]byte{})
-			if _, err := audioSamplesDataBuf.ReadFrom(rd); err != nil {
-				log.Printf("%s: failed to read samples data: %s", u.cfg.Username, err.Error())
-				continue
-			}
+				audioSamplesDataBuf := bytes.NewBuffer([]byte{})
+				if _, err := audioSamplesDataBuf.ReadFrom(rd); err != nil {
+					log.Printf("%s: failed to read samples data: %s", u.cfg.Username, err.Error())
+					return
```
streamer45, line 461: Was this change meant (returning vs continuing)?

streamer45 (commented):
Brilliant work, want to see a demo of this some day :tada:

Left a couple of suggestions but nothing blocking. Also since this will merge in my PR I can take care of some cleanup later on if needed.

lt/cmd/speech/main.go
```diff
@@ -83,3 +112,84 @@ func main() {
 
 	wg.Wait()
 }
+
+func performScript(filename string) error {
+	awsSess := session.Must(session.NewSessionWithOptions(session.Options{
+		SharedConfigState: session.SharedConfigEnable,
+	}))
```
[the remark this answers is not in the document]
cpoile: I added a profile var, you can set `profile=mm-loadtest` now.

lt/client/client.go
cpoile: Yep, because it's in a func now.

lt/cmd/speech/main.go
```diff
@@ -11,17 +15,43 @@ import (
...
 
+const (
+	duration = 10 * time.Minute
+)
+
+var script, siteURL, wsURL, channelID, teamID, userPassword, profile string
+var setup bool
+
 func main() {
+	flag.StringVar(&script, "script", "script.txt", "Script for the tts")
```
streamer45, line 26: This is not working by default because it's looking in the current directory. I'd probably move the scripts to a dedicated folder, maybe on the root of the load-tester.

streamer45 (approved):
Thanks Chris, just one comment. We'll likely review this again as part of ceiling tests and send improvements where needed.

streamer45:
@cpoile One more thing, if you could merge with `main` to see if the server changes go away.

streamer45:
@cpoile Apologies, I just realized we had to merge https://github.com/mattermost/mattermost-plugin-calls/pull/614 to fix this unless we rebase against main.

PR #633, opened 2024-01-26
MM-56540 - Live captions 🎥 

cpoile:
#### Summary
- WIP
- Needs tests and documentation
- UI is largely ready, most of it is ready, just needs some polish.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-56540


webapp/package.json
```diff
@@ -76,6 +76,7 @@
   },
   "dependencies": {
     "@calls/common": "github:mattermost/calls-common#25619914bcfe6347101b29204b2976518742e81e",
+    "@mattermost/compass-icons": "0.1.40",
```
streamer45, line 79: I worked so hard on removing this dependency back then, do we really need to add it back? :p

server/client_message.go
```diff
@@ -8,8 +8,9 @@ import (
 )
 
 type clientMessage struct {
-	Type string          `json:"type"`
-	Data json.RawMessage `json:"data,omitempty"`
+	Type    string          `json:"type"`
+	Data    json.RawMessage `json:"data,omitempty"`
+	mapData map[string]any
```
streamer45, line 13: I assume this can be removed?

webapp/src/types/types.ts
```diff
@@ -180,3 +180,7 @@ export type RealNewPostMessageProps = {
     set_online: boolean;
     team_id: string;
 }
+
+export type LiveCaptions = {
+    [sesionID: string]: LiveCaption;
```
streamer45, line 185: small typo

streamer45 (commented):
Gave a first quick pass, looking great 💯 , only left a couple of minor comments.

webapp/package.json
cpoile: Nope, don't need it at all -- it doesn't seem to work anyway (we only get the one that's included in server), so it's mostly useless to us, considering we add new icons on almost every UI update. :)

server/client_message.go
cpoile: Yep, don't know why that was still there (part of an experiment).

server/websocket.go
```diff
@@ -987,6 +987,24 @@ func (p *Plugin) WebSocketMessageHasBeenPosted(connID, userID string, req *model
 			return
 		}
 		msg.Data = []byte(msgData)
+	case clientMessageTypeCaption:
```
streamer45, line 990: Unless there's a good reason to do it here, it would be best to handle the message in `handleClientMsg` like for most other client initiated events.

webapp/src/components/expanded_view/component.tsx
```diff
@@ -1355,10 +1388,18 @@ const CloseButton = styled.button`
...
     position: absolute;
     bottom: 96px;
     display: flex;
     flex-direction: column;
     gap: 12px;
 `;
+
+const LiveCaptionsOverlay = styled.div`
+    position: absolute;
+    width: 100vw;
```
streamer45, line 1401: Does this mean it will go on top of other elements like participants list, call thread?

webapp/src/constants.ts
```diff
@@ -10,6 +10,7 @@ export const DEFAULT_RING_SOUND = 'Calm';
 export const CALL_START_POST_TYPE = 'custom_calls';
 export const CALL_RECORDING_POST_TYPE = 'custom_calls_recording';
 export const CALL_TRANSCRIPTION_POST_TYPE = 'custom_calls_transcription';
+export const LIVE_CAPTION_TIMEOUT = 5000;
```
streamer45, line 13: Do you think this is generally enough? I suppose it depends on what's the maximum output length of the transcriber.

server/websocket.go
cpoile: I would say the good reason is: it avoids an unnecessary data->json string->data conversion, which isn't needed. I realize we do that from client to backend, but I can't think of a reason to do it from one backend to another. Plus, it makes sense to treat this differently than a typical message from a regular client, since it's not really a regular client.

webapp/src/components/expanded_view/component.tsx
cpoile: It goes on top of reactions and the participant list, but not over the thread view. I can _probably_ figure out how to put it on top of the thread view if needed, but I thought we might not want to do that, because having captions block the thread might be undesirable. I'm fine either way, so maybe @matthewbirtch can let us know which is best?

webapp/src/constants.ts
cpoile: This timer only affects a caption that hasn't had any changes -- so the caption will always be there if there's updates or new lines etc, and only once the participant stops speaking will the timeout remove the caption (after 5 seconds of not speaking). I've found when watching the load test that this leaves a natural amount of time to read the caption. We can tune it to whatever of course.

webapp/src/components/expanded_view/component.tsx
matthewbirtch: @cpoile @streamer45 the caption should definitely not overlap on top of the thread RHS. The captions should stay within the bounds of the main call room area. 
image

cpoile: @matthewbirtch Thank you. How about the case when the participant list is active?

cpoile: After experimenting with both, I think it looks better if the captions are centred in the expanded view main area (with the blue background). That is: (we can adjust the max width etc.)
image

matthewbirtch: Yep, what you've done here makes the most sense to me - centered within the main area. Although, would we have this much text on the screen at one time? I assume we would cap it at 2 lines of text at a time.

For a max-width, I would play with something like 600px?

cpoile: @matthewbirtch As for the amount of text -- these are three people talking at once, so I don't think it makes sense to artificially limit the number of captions, do you? This is probably not going to happen often. Generally there is a 5 second wait until the last caption gets removed, so unless there's overlapping conversations, we usually will only see one or two captions at the same time.

600 px sounds good -- I'll play around it some more during testing and post some screenshots.

matthewbirtch: > these are three people talking at once, so I don't think it makes sense to artificially limit the number of captions, do you

Oh, I see. I didn't make that connection at first - I should have looked closer. How many people are we tracking at one time? If we go much beyond 3, the screen would just fill with captions :) As you mentioned, not a likely scenario, but we may want to come up with some kind of limit to prevent that. Probably not a high priority fix right now though

server/websocket.go
streamer45: Fair enough. We should still call `p.metrics.IncWebSocketEvent` as appropriate though.

server/performance/metrics.go
```diff
@@ -15,23 +15,27 @@ import (
...
 const (
-	metricsNamespace        = "mattermost_plugin_calls"
-	metricsSubSystemWS      = "websocket"
-	metricsSubSystemCluster = "cluster"
-	metricsSubSystemStore   = "store"
+	metricsNamespace             = "mattermost_plugin_calls"
+	metricsSubSystemWS           = "websocket"
+	metricsSubSystemCluster      = "cluster"
+	metricsSubSystemStore        = "store"
+	metricsSubSystemLiveCaptions = "live_captions"
```
streamer45, line 22: I feel the subsystem should be jobs here so that it can be easily extended to more stuff in the future. Then there could be a job type label added to identify what we are talking about or a simple prefix to the metric name.

server/websocket.go
```diff
@@ -987,6 +988,39 @@ func (p *Plugin) WebSocketMessageHasBeenPosted(connID, userID string, req *model
...
+		}
+		newAudioLenMs, ok := req.Data["new_audio_len_ms"].(float64)
+		if !ok {
+			p.LogError("invalid or missing new_audio_len_ms in caption ws message")
+			return
+		}
+		p.handleCaptionMessage(us.channelID, sessionID, userID, text, newAudioLenMs)
+		return
+	case clientMessageTypeMetric:
+		// Sent from the transcriber.
```
streamer45, line 1016: I think this is great and could be sent by any client eventually to track more stuff (e.g. WebRTC call quality as seen from the client side or more client side performance data).

server/websocket.go
```diff
@@ -1062,3 +1096,23 @@ func (p *Plugin) handleBotWSReconnect(connID, prevConnID, originalConnID, channe
...
+	}, &model.WebsocketBroadcast{ChannelId: channelID, ReliableClusterSend: true})
+
+	p.metrics.ObserveLiveCaptionsAudioLen(newAudioLenMs)
+}
+
+func (p *Plugin) handleMetricMessage(metricName public.MetricName) {
+	switch metricName {
+	case public.MetricPressureReleased:
+		p.metrics.IncLiveCaptionsPressureReleased()
+	case public.MetricTranscriberBufFull:
```
streamer45, line 1115: Can we prefix the types so that they match the methods? `MetricPressureReleased` sounds very generic if you don't have the context in mind. I know it becomes verbose but it's a tradeoff I'd take.

streamer45 (commented):
Looking great, left a couple more comments.

cpoile:
@streamer45 FYI: I'm working on the load test doc: https://docs.google.com/document/d/12-hDDtK12YXFneoF8OJ5vn8UkMrGbdtj80KJ_qeFJNg/edit?usp=sharing
And I'm still working on ways of relieving the pressure on the transcribers.

server/websocket.go
cpoile: Yep, sounds good. I'm not averse to long names (I usually think my names are too long and I try to cut them down).

server/performance/metrics.go
cpoile: Good idea, done.

webapp/src/components/expanded_view/component.tsx
cpoile: There's no limit to the number of captions we display at one time, but naturally there won't be too many. They disappear after 5 seconds of inactivity. I think this is the kind of thing we should experiment with with real-world usage and then we'll have a better sense of sense of what limits make sense. With the test scripts (with up to 3 people talking simultaneously, or with more natural one-after-another conversations) the number of captions is never overwhelming.

cpoile:
@streamer45 @matthewbirtch I think I'm pretty much done development on our first version of this. I'd like to put it on Community on Tuesday and let it soak until I head off for PTO on Friday. The load test doc has a couple suggestions for tweaks to make on the after-call transcriber (reduce its threads to 2 or 1).
If Community's transcriber machine is a c6i.2xlarge, we can use 2 threads per live transcriber and 1 live transcriber per pool, which will give us good results with up to four calls. See: https://docs.google.com/document/d/12-hDDtK12YXFneoF8OJ5vn8UkMrGbdtj80KJ_qeFJNg/edit

cpoile:
@matthewbirtch Thanks for the review!
1. Done.
2. 👍 
3. Right now it's set to max of 50% of width. I can set it to max 600px, and otherwise 50% (or whatever % you'd like, let me know).
4-6. These are limitations of the transcription model. You get them with live captions because it's using smaller chunks of audio, and you don't get them with the end-of-call transcription because it has a lot more context. We have to balance the amount of context it uses with the amount of "real-time-ness", since the more context we give the longer the transcription will take. We can tweak those parameters. Right now we're using up to 8 seconds of audio.

matthewbirtch:
> * I'd like to update the `cc` icon because the stroke weight and size don't quite seem to match the other icons - that's my fault. I'll aim to get that to you this week
> *

I've updated the compass icons repo with a new release that includes the updated `cc` icon.

> 3\. Right now it's set to max of 50% of width. I can set it to max 600px, and otherwise 50% (or whatever % you'd like, let me know).

I'd say we could increase it to 70% width up to a max of 600px. When the window is smaller, 50% seems to narrow, so this should be good.

cpoile:
@matthewbirtch Is there an svg version of the cc icon? (perhaps in a figma somewhere? that's where I got it last time). We're unable to use recent versions of the compass-icons because webapp interferes, so we use svgs until everything equalizes.

I've made the change on the caption width -- looks much better. (the test server has the updated plugin so you can test it if you'd like).

matthewbirtch:
> @matthewbirtch Is there an svg version of the cc icon? (perhaps in a figma somewhere? that's where I got it last time). We're unable to use recent versions of the compass-icons because webapp interferes, so we use svgs until everything equalizes.
> 
> I've made the change on the caption width -- looks much better. (the test server has the updated plugin so you can test it if you'd like).

Yep, you can find the SVG for cc icon in the compass-icons repo here: https://github.com/mattermost/compass-icons/blob/master/svgs/closed-caption-outline_F0DBD.svg

For future, all icon SVGS are in the compass-icons repo here: https://github.com/mattermost/compass-icons/blob/master/svgs/

cpoile:
@matthewbirtch Top is the new, bottom is old, does this look good?

image

matthewbirtch:
> @matthewbirtch Top is the new, bottom is old, does this look good?
> 
> 
> 
> image
> 
> 

Hmmm, it still looks too small in comparison to the other buttons. Is there something different we're doing on the cc button from the rest?

cpoile:
@matthewbirtch Nope, I think it's just a smaller icon. The one above was the icon's normal size, this is 28px: 
image
This is 30px:
image
This is 32px:
image
etc.
Which would you prefer?

matthewbirtch:
> @matthewbirtch Nope, I think it's just a smaller icon. The one above was the icon's normal size

> This is 32px:
> 
> image

I think 32 is closest to the intended size. There is something weird though if the icons aren't sizing correctly since the SVGs we've created are all consistently sized to a 24x24 bounding box. I'll see if I can inspect this and see what's going on

cpoile:
@matthewbirtch Ok, new version is up on the test server.

matthewbirtch (approved):

webapp/src/components/expanded_view/component.tsx
```diff
@@ -1027,13 +1040,16 @@ export default class ExpandedView extends React.PureComponent<Props, State> {
...
+        const liveCaptionsText = this.state.showLiveCaptions ? formatMessage({defaultMessage: 'Turn off live captions'}) : formatMessage({defaultMessage: 'Turn on live captions'});
+
         const globalRhsSupported = Boolean(this.props.selectRhsPost);
 
         const isChatUnread = Boolean(this.props.threadUnreadReplies);
 
         const isHost = this.props.callHostID === this.props.currentUserID;
 
         const isRecording = isHost && this.props.isRecording;
+        const showCCButton = this.props.transcriptionsEnabled && this.props.isRecording && Boolean(this.props.callRecording?.start_at);
```
streamer45, line 1052: Adding a comment here since I noticed it on Community: we should hide if live captions are disabled.

cpoile: Thanks, fixed in [3b22c07](https://github.com/mattermost/mattermost-plugin-calls/pull/633/commits/3b22c075324ebeb5777cc81e736252fcb34061e3)

cpoile:
Added https://mattermost.atlassian.net/browse/MM-57408 LiveCaptionsLanguage support

server/bot_api.go
```diff
@@ -477,6 +483,9 @@ func (p *Plugin) handleBotPostJobsStatus(w http.ResponseWriter, r *http.Request,
 			if err := p.stopTranscribingJob(state, callID); err != nil {
 				p.LogError("failed to stop transcribing job", "callID", callID, "err", err.Error())
 			}
+			if lcState != nil {
+				lcState.EndAt = time.Now().UnixMilli()
+			}
```
streamer45, line 488: This is redundant right? I'd expect `stopTranscribingJob` to take care of it.

server/transcription_api.go
```diff
@@ -59,15 +72,25 @@ func (p *Plugin) transcriptionJobTimeoutChecker(callID, jobID string) {
...
-				"recState": clientState.toMap(),
+				"recState": recClientState.toMap(),
 			}, &model.WebsocketBroadcast{ChannelId: callID, ReliableClusterSend: true})
 		}
 
-		p.publishWebSocketEvent(wsEventCallTranscriptionState, map[string]interface{}{
-			"callID":  callID,
-			"trState": clientState.toMap(),
+		// We need to send it as a "recording" because we don't handle the "transcription" type on the client.
+		// MM-57265: However, we don't show this message yet.
```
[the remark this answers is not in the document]
cpoile: I don't remember what I was getting at with the first sentence. The second one with the ticket is still good -- we receive a delayed ws event and don't handle it on the client.

plugin.json
```diff
@@ -288,9 +288,9 @@
     ]
   },
   "props": {
-    "min_rtcd_version": "v0.12.0",
+    "min_rtcd_version": "v0.14.0",
     "min_offloader_version": "v0.7.0",
```
streamer45, line 292: Two things: 

- We don't need to bump `rtcd`. I actually wouldn't do that.
- May want to bump offloader though given config validation will be outdated otherwise.

cpoile: @streamer45 Ok, but why not update rtcd?

streamer45: Mostly because there's no need to. I see updating rtcd here same as updating `min_server_version`. We can do it but keeping some compatibility is a bit easier on upgrades otherwise we can get into sticky situations like when upgrading Community and Hub nothing works unless rtcd is upgraded before the plugin.

cpoile: Ok, sounds good

PR #640, opened 2024-02-19
[MM-56920] Allow passing SiteURL override for jobs

streamer45:
#### Summary

By default our jobs use the SiteURL from config in order to connect. To allow connection from within a local network in production we need to add some overrides.

#### Related PR

https://github.com/mattermost/calls-recorder/pull/64

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-56920


standalone/src/init.ts
```diff
@@ -169,14 +166,6 @@ export default async function init(cfg: InitConfig) {
     }
     RestClient.setToken(getToken());
 
-    // initialize some basic state.
-    await Promise.all([
-        getMe()(store.dispatch, store.getState),
-        getMyPreferences()(store.dispatch, store.getState),
-        getMyTeams()(store.dispatch, store.getState),
-        getMyTeamMembers()(store.dispatch, store.getState),
-    ]);
```
streamer45, line 178: Slightly unrelated but when investigating the flow of requests I figured these were not needed for the recorder. In fact they would all fail as unauthorized since the bot wouldn't be logged in. So we are moving them on the global widget logic .

standalone/src/widget/index.tsx
```diff
@@ -73,7 +75,14 @@ async function initWidget(store: Store) {
 }
 
 async function initStoreWidget(store: Store, channelID: string) {
-    await store.dispatch(getChannelAction(channelID));
+    // initialize some basic state.
+    await Promise.all([
+        getMe()(store.dispatch, store.getState),
+        getMyPreferences()(store.dispatch, store.getState),
+        getMyTeams()(store.dispatch, store.getState),
+        getMyTeamMembers()(store.dispatch, store.getState),
```
cpoile, line 83: nit: I think all of these can be done like in like 84, e.g. `store.dispatch(getMe())`. But no big deal either way.

cpoile (approved):

standalone/src/widget/index.tsx
streamer45: oh yeah, after years I am still confused by actions. I'll update.

PR #646, opened 2024-02-29
Fix fetching captions

streamer45:
#### Summary

Since https://github.com/mattermost/mattermost/pull/26239 permissions checks on files require a channel ID to be set so it's not sufficient to use the post ID.

### DIFF
--- a/server/bot_api.go
+++ b/server/bot_api.go
@@ -327,7 +327,7 @@ func (p *Plugin) handleBotPostTranscriptions(w http.ResponseWriter, r *http.Requ
 	// Updating the file to point to the existing call post solves this problem
 	// without requiring us to expose a dedicated API nor attach the file which
 	// we don't want to show.
-	if err := p.updateFileInfoPostID(info.Transcriptions[0].FileIDs[0], info.PostID); err != nil {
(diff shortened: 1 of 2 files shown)


server/bot_api.go
```diff
@@ -327,7 +327,7 @@ func (p *Plugin) handleBotPostTranscriptions(w http.ResponseWriter, r *http.Requ
 	// Updating the file to point to the existing call post solves this problem
 	// without requiring us to expose a dedicated API nor attach the file which
 	// we don't want to show.
-	if err := p.updateFileInfoPostID(info.Transcriptions[0].FileIDs[0], info.PostID); err != nil {
+	if err := p.updateFileInfoPostID(info.Transcriptions[0].FileIDs[0], callID, info.PostID); err != nil {
```
cpoile, line 330: arg, this `callID` == `channelID` is going to confuse me forever. :)

cpoile (approved):

server/bot_api.go
streamer45: @cpoile I hear you, that's why we have https://mattermost.atlassian.net/browse/MM-54253

PR #650, opened 2024-03-04
[MM-51852] Update server version to min version that has MM-51852

cpoile:
#### Summary
- https://github.com/mattermost/mattermost/pull/24731 has a breaking change, and the earliest server branch with that PR is 9.2.0, so make 9.2.0 our new min server version.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-56659

### DIFF
--- a/plugin.json
+++ b/plugin.json
@@ -4,7 +4,7 @@
   "description": "Integrates real-time voice communication in Mattermost",
   "homepage_url": "https://github.com/mattermost/mattermost-plugin-calls/",
   "support_url": "https://github.com/mattermost/mattermost-plugin-calls/issues",
-  "min_server_version": "9.2.0",
+  "min_server_version": "9.5.0",
   "server": {
     "executables": {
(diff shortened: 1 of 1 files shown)


streamer45:
Wondering if we should bump to latest ESR (v9.5) since we are at it. Most customers would jump to that one when they reach v9.

cpoile:
@streamer45 I thought it was more of a "this is the bare minimum you need" kind of thing, not a "we recommend you be on this ESR for various other reasons." But if you prefer, I don't mind either way.

streamer45:
> @streamer45 I thought it was more of a "this is the bare minimum you need" kind of thing, not a "we recommend you be on this ESR for various other reasons." But if you prefer, I don't mind either way.

That's correct but also have a somewhat important security fix we merged recently (https://github.com/mattermost/mattermost-plugin-calls/pull/620) which doesn't really work before 9.5 either. So maybe there's a chance to align with ESRs going forward.

streamer45 (approved):
Thanks!

Issue #651, opened 2024-03-05
The plugin does not apply the language set by the user


wensimin:
mattermost version : mattermost-team-edition:release-9.6
com.mattermost.calls version:  0.24.0
I see the following error message in the console

![1709630440460](https://github.com/mattermost/mattermost-plugin-calls/assets/11828594/fe7e6777-01a1-42b0-818b-b94e23d0ef2b)

Did I get something wrong?

streamer45:
Hi @wensimin, it looks like we don't have translations for that specific language code. I am wondering if this is a simple issue of language identifiers. Would any of the following translation files apply to you?

https://github.com/mattermost/mattermost-plugin-calls/blob/main/webapp/i18n/zh_Hans.json

https://github.com/mattermost/mattermost-plugin-calls/blob/main/webapp/i18n/zh_Hant.json

wensimin:
Yes, I think it is zh_Hans.json @streamer45 

But I can't find this in the packaged file

Is there a way for me to change this code manually?

Thank you

wensimin:
Thanks for your work！

Claudio Costa ***@***.***> 于 2024年3月6日周三 上午10:29写道：

PR #653, opened 2024-03-06
Fix Chinese language code mapping

streamer45:
#### Summary

After some discussion https://community.mattermost.com/core/pl/b8cans4gy7b6fyw9jkstyfp7wr we re-map certain codes to the language files we provide instead of duplicating them.

Fixes https://github.com/mattermost/mattermost-plugin-calls/issues/651


cpoile (approved):
Thank you!

PR #657, opened 2024-03-12
[MM-57211] Fix issues with multiple transcription jobs per call

streamer45:
#### Summary

In case of multiple transcription jobs per call we were not writing the metadata correctly which would cause failure to attach and render the captions. 

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-57211

### DIFF
--- a/server/job_metadata.go
+++ b/server/job_metadata.go
@@ -109,7 +109,7 @@ func (p *Plugin) saveRecordingMetadata(postID, recID, trID string) error {
 				trID: tm.toMap(),
 			}
 		} else {
-			recordings[trID] = tm.toMap()
+			transcriptions[trID] = tm.toMap()
 		}
 		post.AddProp("transcriptions", transcriptions)
 	}
--- a/server/recording_api.go
+++ b/server/recording_api.go
@@ -17,21 +17,21 @@ import (
 
 var callRecordingActionRE = regexp.MustCompile(`^\/calls\/([a-z0-9]+)/recording/(start|stop|publish)$`)
 
-const recordingJobStartTimeout = 2 * time.Minute
+const recordingJobStartTimeout = time.Minute
 
 func (p *Plugin) recJobTimeoutChecker(callID, jobID string) {
 	time.Sleep(recordingJobStartTimeout)
 
 	state, err := p.lockCall(callID)
 	if err != nil {
-		p.LogError("failed to lock call", "err", err.Error())
+		p.LogError("failed to lock call", "err", err.Error(), "callID", callID, "jobID", jobID)
 		return
 	}
 	defer p.unlockCall(callID)
 
 	recState, err := state.getRecording()
 	if err != nil {
-		p.LogError("failed to get recording state", "error", err.Error())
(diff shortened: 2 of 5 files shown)


server/job_metadata.go
```diff
@@ -109,7 +109,7 @@ func (p *Plugin) saveRecordingMetadata(postID, recID, trID string) error {
 				trID: tm.toMap(),
 			}
 		} else {
-			recordings[trID] = tm.toMap()
```
streamer45, line 112: This is the bug, an innocent typo.

server/recording_api.go
```diff
@@ -17,21 +17,21 @@ import (
 
 var callRecordingActionRE = regexp.MustCompile(`^\/calls\/([a-z0-9]+)/recording/(start|stop|publish)$`)
 
-const recordingJobStartTimeout = 2 * time.Minute
```
streamer45, line 20: Lowering this a bit since 2 minutes is just too much for a user to be waiting.

server/transcription_api.go
```diff
@@ -51,9 +51,18 @@ func (p *Plugin) transcriptionJobTimeoutChecker(callID, jobID string) {
...
+				p.LogError("failed to stop recording job", "err", err.Error(), "callID", callID, "jobID", jobID)
 			}
+
+			// This is needed as we don't yet handle wsEventCallTranscriptionState on
+			// the client since jobs are coupled.
+			recClientState.Err = "failed to start transcriber job: timed out waiting for bot to join call"
+			p.publishWebSocketEvent(wsEventCallRecordingState, map[string]interface{}{
+				"callID":   callID,
+				"recState": clientState.toMap(),
+			}, &model.WebsocketBroadcast{ChannelId: callID, ReliableClusterSend: true})
```
streamer45, line 65: This is something I forgot to add originally. If the transcribing job fails first then we wouldn't notify the host as we are not handling the below `wsEventCallTranscriptionState` event.

server/transcription_api.go
```diff
@@ -26,7 +26,7 @@ func (p *Plugin) transcriptionJobTimeoutChecker(callID, jobID string) {
 
 	trState, err := state.getTranscription()
 	if err != nil {
-		p.LogError("failed to get transcription state", "error", err.Error())
+		p.LogWarn("failed to get transcription state", "err", err.Error(), "callID", callID, "jobID", jobID)
```
cpoile, line 29: Not asking for a change, but just curious why the change to LogWarn? If we return, it seems like an error makes sense?

cpoile (approved):

server/transcription_api.go
streamer45: Good question. The reason is that in most cases that's not really a sign of a problem. For example it will happen in any of these cases:

- If host ends a call before the bot gets a chance to join.
- If host stops and restart a recording before the first one started.
- If any of the two coupled jobs (transcriptions and recording) fails as we also stop the other but don't explicitly cancel the timeout checker.

cpoile: Ahh, okay, that makes sense. Thanks.

PR #658, opened 2024-03-12
[MM-42464] SQL store: `calls` & `calls_channels`

streamer45:
#### Summary

PR adds the basic store implementation for `calls` and `calls_channels`.

#### Ticket Link

#### Parent PR

https://github.com/mattermost/mattermost-plugin-calls/pull/647


server/interfaces/interfaces.go
```diff
@@ -25,3 +27,11 @@ type Metrics interface {
...
 	IncStoreOp(op string)
 }
+
+type Store interface {
+	Migrate(direction models.Direction, dryRun bool) error
+	Close() error
+	GetPost(postID string) (*model.Post, error)
+	UpdateFileInfoPostID(fileID, channelID, postID string) error
+	KVGet(pluginID, key string, fromWriter bool) ([]byte, error)
+}
```
streamer45, line 37: Hopefully we won't need this for much longer. It's currently used to allow mocking.

server/db/call_store.go
```diff
@@ -0,0 +1,227 @@
...
+	if err := s.dbXFromGetOpts(opts).Get(&call, q, args...); err == sql.ErrNoRows {
+		return nil, fmt.Errorf("call not found")
+	} else if err != nil {
+		return nil, fmt.Errorf("failed to get call: %w", err)
+	}
+
+	return &call, nil
+}
+
+func (s *Store) GetCallByChannelID(channelID string, opts GetCallOpts) (*public.Call, error) {
```
streamer45, line 207: There's a bit of redundancy here given the `CallID` vs `ChannelID` conundrum. We could combine by doing a simple `OR` but for now I am keeping them separate as we may not need them both, depending on how we implement things.

server/public/call.go
```diff
@@ -0,0 +1,21 @@
+package public
+
+type Call struct {
+	ID           string      `json:"id"`
+	ChannelID    string      `json:"channel_id"`
+	StartAt      int64       `json:"start_at"`
+	EndAt        int64       `json:"end_at"`
+	CreateAt     int64       `json:"create_at"`
+	DeleteAt     int64       `json:"delete_at"`
```
streamer45, line 9: Do we need an `UpdateAt`?

server/db/migrations/mysql/000002_create_calls.down.sql
```diff
@@ -0,0 +1,16 @@
+SET @preparedStatement = (SELECT IF(
```
cpoile, line 1: When do we officially drop support for mysql?

server/db/misc_store.go
```diff
@@ -0,0 +1,80 @@
...
+import (
+	"database/sql"
+	"fmt"
+
+	"github.com/mattermost/mattermost/server/public/model"
+
+	sq "github.com/mattermost/squirrel"
+)
+
+func (s *Store) KVGet(pluginID, key string, fromWriter bool) ([]byte, error) {
```
cpoile, line 12: Can we have a comment explaining why this KVGet (like you have for GetPost)?

cpoile (approved):
Very clean, thanks Claudio!

server/public/call.go
streamer45: Not really, I think we can leave it off for now.

server/db/call_store.go
streamer45: Sure, difference is simple though. `CallID` is unique, hence why it will be primary key here, `ChannelID` on the other hand is shared by all of the calls that will ever happen in a channel. However, since we don't support multiple concurrent calls in the same channel, `channelID` is what's usually used by clients and APIs to identify the current (active) call.

server/db/misc_store.go
streamer45: I think this is going away as soon as everything else is implemented.

streamer45:
Merging to keep things tidy. Can address feedback in subsequent PRs.

PR #659, opened 2024-03-13
[MM-42464] SQL store: `calls_sessions`

streamer45:
#### Summary

This is the last bit in terms of new tables (for now). 

As discussed, I am leaving jobs stuff behind for the time being while we figure out if we need them or if we can still rely on post props.

I'll also need to likely tweak some things such as adding indexes but I prefer to defer that to the upcoming PR which will be updating the business logic as it will become clearer how and when we fetch data.

#### Ticket Link

#### Parent PR

https://github.com/mattermost/mattermost-plugin-calls/pull/647


cpoile (approved):
Looks good, thanks!

PR #663, opened 2024-03-18
[MM-57327] Set basename on default client as well if present

### LINKED ISSUE #660: Setting site url is not used in mattermost desktop
I have a mattermost service at https://mydomain/mattermost/
But calls requested the following API in Windows Desktop
image
The correct request should be
`:3000/mattermost/api/v4/channels/c8h6b3jy4ibni86jjzry53r1uc`
not `:3000/api/v4/channels/c8h6b3jy4ibni86jjzry53r1uc`
My setting
image

streamer45:
#### Summary

The default client used by redux action wasn't accounting for the `basename` any longer so it would break any use of Calls from desktop against installations served under a subpath. Explicitly setting the url seems to fix this.

I think this may be a regression introduced in https://github.com/mattermost/mattermost-plugin-calls/pull/555/files#diff-2aac6bf65edf6e3c7580f45c0aee032bcdded5c0adb7a5ef9929ee31b74e0ad3. 

@cpoile Please let me know if you think this should be addressed differently. It doesn't look great but at the same time I didn't want to go as far as implementing our own basic actions to load MM data in Redux.

Fixes #660 

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-57327

### DIFF
--- a/standalone/src/init.ts
+++ b/standalone/src/init.ts
@@ -41,6 +41,7 @@ import {
 import {pluginId} from 'plugin/manifest';
 import reducer from 'plugin/reducers';
 import RestClient from 'plugin/rest_client';
+import {Client4} from 'mattermost-redux/client';
 import {callsConfig, iceServers, needsTURNCredentials} from 'plugin/selectors';
 import {DesktopNotificationArgs, Store, WebAppUtils} from 'plugin/types/mattermost-webapp';
 import {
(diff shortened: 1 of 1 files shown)


streamer45:
I think we should consider a dot release for this given the impact.

cpoile (approved):
A couple thoughts:
- Should we do the same for ~~the standalone~~ webapp?
- If we restrict ourselves to using the RestClient, then would this be an issue? I don't think we need to use Client4 at all, instead use RestClient. But perhaps Client4 is being used somewhere else even if we never touch it? Maybe I can try it if you want to send a new ticket to me.

PR #666, opened 2024-03-27
[MM-42464] SQL store: Business logic

streamer45:
#### Summary

This is the most substantial PR as it includes all the business logic changes needed after moving from KV to SQL store. Most of the complexity was changing from the JSON blob format for the call state to having separate entities in their own tables, namely calls, channels, sessions and jobs.

I tried to simplify where possible but without making any changes to the client side. More work will be needed to clean up properly and provide a consistent API but I'll defer that since this is more than enough for now :)

Given we are touching pretty much all the backend, there's a very good chance some regressions may have creeped in. My plan is to add more tests to cover as many edge cases as possible and get this on Community as soon as possible once e2e and manual QA are all good.

@cpoile If you could take care of some smoke testing on mobile I'd be grateful. No need to go too deep but at least testing the basics would be a good start. PR should not have any breaking changes so if you spot anything funny please let me know.

#### Ticket Link


server/state.go
```diff
@@ -255,133 +167,174 @@ func (cs *callState) getClientState(botID, userID string) *CallStateClient {
...
+	jobs, err := p.store.GetCallJobs(call.ID, db.GetCallJobOpts{
+		FromWriter: fromWriter,
+	})
 	if err != nil {
-		return fmt.Errorf("failed to marshal channel state: %w", err)
+		return nil, fmt.Errorf("failed to get call jobs: %w", err)
+	}
+	for _, job := range jobs {
+		if job.Type == public.JobTypeRecording && state.Recording == nil {
+			state.Recording = job
```
cpoile, line 239: What's the purpose of the `state.Recording == nil` and `state.Transcription == nil` checks?

server/sync.go
```diff
@@ -11,9 +11,8 @@ import (
 	"github.com/mattermost/mattermost-plugin-calls/server/cluster"
 )
 
-// lockCall locks the global (cluster) mutex for the given channelID and
-// returns the current state.
-func (p *Plugin) lockCall(channelID string) (*channelState, error) {
+// lockCallSimple locks the global (cluster) mutex for the given channelID.
+func (p *Plugin) lockCallSimple(channelID string) error {
```
cpoile, line 15: I think I understand what you mean, but maybe there's a better name for this and the next? like `lockCall` and `lockCallAndReturn`? But it's a matter of taste, so up to you.

cpoile (commented):
This is a lot of work, wow. PR num 666 is right.

I ran through the basics on mobile, joining leaving raising talking reacting recording, looks like it's working well.

Not anything serious, just mostly questions really.

Heads up that I'm getting the following error after upgrading and after starting and ending a couple calls (successfully):
```
error [2024-03-28 17:41:23.692 -04:00] failed to get all calls channels              caller="app/plugin_api.go:1003" plugin_id=com.mattermost.calls origin="main.(*Plugin).handleGetAllCallChannelStates api.go:146" err="failed to get calls channels: sql: Scan error on column index 2, name "props": unsupported source type <nil>"
```

server/api_router.go
```diff
@@ -0,0 +1,157 @@
...
+
+func (p *Plugin) newAPIRouter() *mux.Router {
+	router := mux.NewRouter()
+
+	// Public API endpoints (no auth required)
+
+	versionRoute := router.HandleFunc("/version", p.handleGetVersion).Methods("GET")
+	var metricsRoute *mux.Route
+	if p.metrics != nil {
+		// NOTE: deprecated in favor of the ServeMetrics hook. Consider removing in v1.0.
```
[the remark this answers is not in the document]
streamer45: I think we are still relying on it in production as we speak as it's not being that long. It's also still documented -> https://docs.mattermost.com/configure/calls-deployment.html#calls-plugin-metrics

Agreed we need a ticket to get rid of this.

server/db/calls_channels_store.go
```diff
@@ -52,25 +48,46 @@ func (s *Store) GetCallsChannel(channelID string, opts GetCallsChannelOpts) (*pu
...
-		return fmt.Errorf("channel should not be nil")
+	if err := channel.IsValid(); err != nil {
+		return fmt.Errorf("invalid channel: %w", err)
 	}
 
 	qb := getQueryBuilder(s.driverName).
 		Update("calls_channels").
 		Set("Enabled", channel.Enabled).
-		Set("Props", channel.Props).
+		Set("Props", s.newJSONValueWrapper(channel.Props)).
```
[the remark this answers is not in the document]
streamer45: Before I tried to have the `CallsChannelProps` type implement [`driver.Valuer`](https://pkg.go.dev/database/sql/driver#Valuer) (see https://github.com/mattermost/mattermost-plugin-calls/pull/647/files#diff-da74529495f2cec1febd6e6f6d7da6e13fd603cfda02ea1866effd10a7c34427) but that broke completely when I realized we could be using Postgres with binary parameters enabled and depending on that we need to marshal the JSON as either bytes (binary) or string (text) so we need store level context to make that decision, hence the wrapper.

server/db/calls_store.go
```diff
@@ -28,86 +28,50 @@ var callsColumns = []string{
...
 		Set("EndAt", call.EndAt).
 		Set("DeleteAt", call.DeleteAt).
-		Set("Participants", call.Participants).
-		Set("Stats", call.Stats).
-		Set("Props", call.Props).
-		Where(
-			sq.Eq{"ID": call.ID},
-			sq.Eq{"ChannelID": call.ChannelID},
-		)
+		Set("ThreadID", call.ThreadID).
```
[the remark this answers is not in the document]
streamer45: If we ever have that use case we'll modify the query. I tried to only update what we currently use.

server/db/store.go
```diff
@@ -66,6 +74,14 @@ func NewStore(settings model.SqlSettings, wConnector, rConnector driver.Connecto
 		log:        log,
 	}
 
+	if *settings.DriverName == model.DatabaseDriverPostgres {
+		binaryParams, err := hasBinaryParams(*settings.DataSource)
+		if err != nil {
+			return nil, fmt.Errorf("failed to check binary parameters")
+		}
+		st.binaryParams = binaryParams
+	}
```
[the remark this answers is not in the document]
streamer45: I am testing because during development everything was working until I tried it on a real instance. See [the above comment](https://github.com/mattermost/mattermost-plugin-calls/pull/666#discussion_r1543780482) for context. Just want to make sure we test the store in both configurations.

server/public/adapters.go
```diff
@@ -61,3 +44,12 @@ func (cp *CallProps) Scan(src any) error {
...
 	return json.Unmarshal(data, cp)
 }
+
+func (jp *CallJobProps) Scan(src any) error {
+	data, ok := src.([]byte)
+	if !ok {
+		return fmt.Errorf("unsupported source type %T", src)
+	}
+
+	return json.Unmarshal(data, jp)
```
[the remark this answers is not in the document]
streamer45: The json package is well known for its inconsistencies :p Still waiting for v2 :)

server/state.go
streamer45: That's a very good question. I meant to leave a comment. The reason is that the returned jobs are sorted by `StartAt` and the `nil` check is because we want to extract only the most recent one because in theory there's no DB constraint so there could be more than one "unfinished". If we were returning a single job type we could `LIMIT` but in this case we want to get all types in one go.

server/sync.go
streamer45: Yeah, this is what a real `lockCall` was supposed to look like. We can rename.

streamer45:
> Heads up that I'm getting the following error after upgrading and after starting and ending a couple calls (successfully):
> 
> ```
> error [2024-03-28 17:41:23.692 -04:00] failed to get all calls channels              caller="app/plugin_api.go:1003" plugin_id=com.mattermost.calls origin="main.(*Plugin).handleGetAllCallChannelStates api.go:146" err="failed to get calls channels: sql: Scan error on column index 2, name "props": unsupported source type <nil>"
> ```

Thanks. I am still thinking whether we should just have a `NOT NULL` constraint for JSON based columns, or add a `{}` default or just deal with the fact they could be `NULL`. Will make sure to figure it out.

server/api_router.go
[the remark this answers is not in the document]
streamer45: Created https://mattermost.atlassian.net/browse/MM-57549, will add a comment.

server/db/migrations/mysql/000001_create_calls_channels.up.sql
```diff
@@ -1,7 +1,7 @@
 CREATE TABLE IF NOT EXISTS calls_channels (
     ChannelID varchar(26) PRIMARY KEY,
     Enabled BOOLEAN,
-    Props JSON
+    Props JSON NOT NULL
```
streamer45, line 4: @cpoile I fixed the migration, went with `NOT NULL` but didn't enforce a default at the schema level due to some [MySQL quirks](https://www.percona.com/blog/cant-we-assign-a-default-value-to-the-blob-text-geometry-and-json-data-types/) I'd rather not deal with. A default value is still enforced by store logic as during insert a `nil` map will be serialized to `null` so we should be good.

The only caveat is that this is not a backwards compatible change in terms of your local state but I wanted to avoid the less clean `COALESCE` (or adding a new migration) since we haven't released any of this yet so to fix your db you'll need to manually run something like:

```sql
UPDATE calls_channels SET Props = 'null' WHERE Props IS NULL;
```

cpoile: Of course 'null' is NOT NULL. 😝

cpoile (approved):
Thanks, this is a lot of good work!

server/db/migrations/mysql/000001_create_calls_channels.up.sql
streamer45: Yeah, `null` JSON literal :)

PR #668, opened 2024-03-27
MM-54243 - Slash command for host change

cpoile:
#### Summary
- rebase and cherry pick of #508 
- host controls are behind the professional license

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-54243
- Fixes https://mattermost.atlassian.net/browse/MM-54313


server/enterprise/license.go
```diff
@@ -29,6 +29,11 @@ func (e *LicenseChecker) isAtLeastE20Licensed() bool {
 	return license.IsE20LicensedOrDevelopment(e.api.GetConfig(), e.api.GetLicense())
 }
 
+// isAtLeastE10Licensed returns true when the server either has at least an E20 license or is configured for development.
```
streamer45, line 32:
```suggestion
// isAtLeastE10Licensed returns true when the server either has at least an E10 license or is configured for development.
```

server/state.go
```diff
@@ -205,8 +206,19 @@ func (us *userState) getClientState(sessionID string) UserStateClient {
 }
 
 func (cs *callState) getHostID(botID string) string {
-	var host userState
+	if cs.HostLocked {
+		return cs.HostID
+	}
 
+	// if current host is still in the call, keep them as the host
+	for _, state := range cs.Sessions {
```
streamer45, line 214: Do we absolutely need to loop twice over sessions?

server/session.go
```diff
@@ -217,9 +215,7 @@ func (p *Plugin) removeUserSession(state *channelState, userID, connID, channelI
 
 	delete(state.Call.Sessions, connID)
 
-	if state.Call.HostID == userID && len(state.Call.Sessions) > 0 {
-		state.Call.HostID = state.Call.getHostID(p.getBotID())
-	}
```
streamer45, line 222: A little unsure on why this was removed. If the current host is still in the call why bother with recalculating everything and loop through sessions?

cpoile: If I remember, that was in response to your comment that we should move all the logic into the `getHostID`...
This happens so rarely (comparatively) that I liked the idea of keeping the logic in one place, but your call.

server/state.go
cpoile: No, but this was done for clarity. I made the change, but I think it's worse for understandability...

streamer45: With the comment I think it's more than reasonable. It's a rather simple rule after all.

server/session.go
cpoile: Ok, as you like.

server/state.go
cpoile: Ok

streamer45:
@cpoile One non-blocking thought that comes to mind. If an assigned host leaves, the call will be host-less until they either come back or a new host is assigned by an admin. I am wondering if this is always desired/expected. I wonder if it would be acceptable to promote someone else to host according to the usual rule and if the assigned one comes back, change again.

streamer45 (approved):

cpoile:
@streamer45 I'm trying something here, let me know what you think. The tricky part is (unless I'm missing something) we need to remember who the assigned host was, while allowing another person to be host.

streamer45:
> @streamer45 I'm trying something here, let me know what you think. The tricky part is (unless I'm missing something) we need to remember who the assigned host was, while allowing another person to be host.

That sounds about right. Also, I was merely asking, I assume it makes sense to you as well? :p

streamer45:
One great thing (can be deferred) would be some e2e tests so that when I go and convert this to the new data store I know if I messed it all up :p

cpoile:
@streamer45 Definitely, https://mattermost.atlassian.net/browse/MM-57529

Also, just added something we were missing: hosts should be allowed to change hosts.

cpoile:
@streamer45 ok to merge?

PR #670, opened 2024-04-01
[MM-42464] Merge in latest changes

streamer45:
#### Summary

Tried my best but can't guarantee it will be bug free as I had to manually touch lots of files.


cpoile (approved):
Thanks for putting this together, that's not an insignificant amount of work. 🎉

PR #673, opened 2024-04-03
[MM-57264] Close emoji picker and bar when a reaction is chosen

streamer45:
#### Summary

PR implements some UX improvements to how the reaction bar and picker behave. We now close them automatically in all of the following cases:

- User reacts
- User presses ESC key
- User clicks outside the element

@matthewbirtch I left raising hand out of this logic since it wasn't specified but let me know.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-57264

### DIFF
--- a/e2e/tests/popout.spec.ts
+++ b/e2e/tests/popout.spec.ts
@@ -240,26 +240,6 @@ test.describe('popout window', () => {
         await expect(page.locator('#calls-widget')).toBeHidden();
     });
 
-    test('raising hand', async ({page, context}) => {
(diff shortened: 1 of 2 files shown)


cpoile (approved):
Nice!

However I'm not sure I want this, but I see it's what Zoom does... It removes the ability to react multiple times quickly.

streamer45:
> Nice!
> 
> However I'm not sure I want this, but I see it's what Zoom does... It removes the ability to react multiple times quickly.

Yeah, it totally prevents emoji spamming :(

PR #674, opened 2024-04-03
SQL store: add `idx_calls_end_at`

streamer45:
#### Summary

I reviewed our stores and on the side of not over-optimizing the only index worth adding as things stand is on `calls.endat` given the table will grow indefinitely and fetching the active calls relies on that particular column.


cpoile (approved):
Cool, sounds good.

PR #675, opened 2024-04-03
[MM-57266] UX improvements to call card (post component)

streamer45:
#### Summary

PR implements improvements as suggested in the [design file](https://www.figma.com/file/QwtIHBpTPtcDy7ccZDRtoD/MM--Call-Card-UI-Updates?type=design&node-id=1335-418726&mode=design)

@matthewbirtch One bit missing here is the row divider in "narrow" view. This is because it's a bit tricky to change layout dynamically as we'd need to show an element that is not there naturally. It can be done but it may end up being hacky as CSS would handle any wrapping naturally when necessary. But let me know how important this is and whether we should reconsider.

@cpoile On the above, let me know if you have any ideas but the only solution I could think of so far would be to hook up a [`ResizeObserver`](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver/ResizeObserver) and dynamically check on div width to render the divider but it's probably going to be a perf nightmare.

#### Screenshots

image

image

image

image

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-57266


cpoile:
Would css container queries work? I'm not sure what browsers we have to support (it's newer) (https://caniuse.com/css-container-queries).
Since we can assume the card will wrap at a certain viewport width, could we use a media query event?
This is useful: https://tigeroakes.com/posts/resize-event-alternatives/

streamer45:
> Would css container queries work? I'm not sure what browsers we have to support (it's newer) (https://caniuse.com/css-container-queries). 

Thanks! Let me look into that.

> Since we can assume the card will wrap at a certain viewport width, could we use a media query event? This is useful: https://tigeroakes.com/posts/resize-event-alternatives/


It's not really the viewport the main issue, but the fact the component renders in RHS and RHS is expansible so it's a bit more tricky. Unless of course we wrapped when in RHS always but that would be a big generalization.

streamer45:
@cpoile Container queries seem to be exactly what we require here! The only problem now is that we need to bump styled components to make them work (from v5 to v6) which may turn out to be a bit of a pain, getting all sorts of errors. Will keep you posted.

webapp/src/components/custom_post_types/post_type/component.tsx
```diff
@@ -248,15 +258,19 @@ const Left = styled.div`
 const Right = styled.div`
     display: flex;
     flex-grow: 1;
+    gap: 12px;
```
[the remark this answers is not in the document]
streamer45: Thanks, I didn't realize that was causing unwanted space. We can probably avoid rendering altogether if call ended and no recordings are available but I actually like the CSS solution as it's implementation agnostic.

webapp/src/components/custom_post_types/post_type/component.tsx
```diff
@@ -268,12 +282,19 @@ const MessageWrapper = styled.div`
...
 `;
 
 const SubMessage = styled.div`
     white-space: normal;
+    font-size: 12px;
+    line-height: 16px;
+    color: rgba(var(--center-channel-color-rgb), 0.72);
 `;
 
 const Profiles = styled.div`
```
matthewbirtch, line 300: Add here as well
```
&:empty {
    display: none;
}
```

matthewbirtch (changes requested):
This is looking great. Thanks for the work on it @streamer45. A few minor things from me:

1. In the "Call ended" state, in the RHS there is extra space at the bottom of the card. It's likely because the right content container is still present, but empty and the gap is added.
    image

2. For the divider, it would be great if we can add it somehow as it helps make it look more organized when it wraps. If we are able to do `@container` queries so that the border and gap shows only when the component is below 400px (roughly) that could work.

3. Can we reduce the gap between the avatars and the button?
    Screenshot 2024-04-05 at 12 22 53 PM

streamer45:
Rebased and applied the changes (using `@container` queries now). Should look better now.

webapp/src/components/custom_post_types/post_type/component.tsx
```diff
@@ -245,6 +248,17 @@ const SubMain = styled.div`
     width: 100%;
     flex-wrap: wrap;
     row-gap: 20px;
```
matthewbirtch, line 250: update to `row-gap: 12px`

matthewbirtch:
Thanks @streamer45. Looks great now. Just one minor comment. Should be good to go once that's updated.

matthewbirtch (approved):
Thanks @streamer45

webapp/src/components/custom_post_types/post_type/component.tsx
```diff
@@ -248,15 +272,42 @@ const Left = styled.div`
...
+
+const RowDivider = styled.hr`
+    display: none;
+
+    @container main (inline-size < 566px) {
+        &&&& {
+          display: block;
+          width: 100%;
+          margin: 0;
+          border: 1px solid rgba(var(--center-channel-color-rgb), 0.08);
```
matthewbirtch, line 296: One more minor nit: this should be 
```
border-top: 1px solid rgba(var(--center-channel-color-rgb), 0.08);
```
Otherwise it ends up being a 2px border on the `hr` element (since it adds the border to all sides).

cpoile (approved):

PR #678, opened 2024-04-04
[MM-57471] Host controls: Three dot menus

cpoile:
#### Summary
- Adds three dot menus to the expanded view, participant list rhs, and widget participant list
- Only `Make host` is showing and implemented
- Some rules: if you are an admin you get the host controls; if you are admin you get controls over the host; if you are host you don't have controls over yourself
- Designs: https://www.figma.com/file/8C0ojMc778Zx3a0nm66Edw/MM-56584-Host-Controls?node-id=1335%3A418726&mode=dev

There's usually these states for the host controls button:
- hovering in the user's row
- hovering over the three-dot button
- clicking the three-dot button but not hovering over the menu item (in the widget participant list in the webapp (not desktop) the menu item is auto-highlighted for some reason)
- and hovering over the menu item

The desktop is a bit tricky, for now it fits but we might have to expand the window when we get more menu items. I'll wait till we get to that point.

Webapp

|   |   |   |
|--|--|--|
| image | image | image |
| image| image | image|
| image | | |
| image | image | image |
| image | | |

Desktop
|  |  |  |
|---|---|---|
| image | image | image |
| image | | |

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-57471


cpoile:
I'll update snapshots when we get approval on the design changes, then the e2e tests will pass.

server/api_router.go
```diff
@@ -93,6 +93,10 @@ func (p *Plugin) newAPIRouter() *mux.Router {
 	router.HandleFunc("/calls/{channel_id:[a-z0-9]{26}}/dismiss-notification", p.handleDismissNotification).Methods("POST")
 	router.HandleFunc("/calls/{call_id:[a-z0-9]{26}}/recording/{action}", p.handleRecordingAction).Methods("POST")
 
+	// Host Controls
+	hostCtrlRouter := router.PathPrefix("/host").Subrouter()
+	hostCtrlRouter.HandleFunc("/{call_id:[a-z0-9]{26}}/make_host/{new_host_id:[a-z0-9]{26}}", p.handleMakeHost).Methods("POST")
```
streamer45, line 98: Given hosts only make sense in the context of calls I think this endpoint should go under the `/calls` root.

server/host_controls_api.go
```diff
@@ -0,0 +1,25 @@
+package main
+
+import (
+	"github.com/gorilla/mux"
+	"net/http"
+)
+
+func (p *Plugin) handleMakeHost(w http.ResponseWriter, r *http.Request) {
+	var res httpResponse
+	defer p.httpResponseHandler(&res, w)
```
streamer45, line 10: We should use `p.httpAudit()`.

standalone/src/recording/components/recording_view/index.tsx
```diff
@@ -14,14 +14,19 @@ import {useSelector} from 'react-redux';
...
+} from 'src/selectors';
 
 const MaxParticipantsPerRow = 10;
 
 const RecordingView = () => {
     const {formatMessage} = useIntl();
-    const [screenPlayerNode, setScreenPlayerNode] = useState<HTMLVideoElement|null>(null);
-    const [screenStream, setScreenStream] = useState<MediaStream|null>(null);
+    const [screenPlayerNode, setScreenPlayerNode] = useState<HTMLVideoElement | null>(null);
+    const [screenStream, setScreenStream] = useState<MediaStream | null>(null);
```
streamer45, line 29: More custom linting, is this your IDE's revenge against me? :p

webapp/src/components/call_widget/participant.tsx
```diff
@@ -0,0 +1,161 @@
...
+    );
+};
+
+const StyledDotMenu = styled(DotMenu)`
+    margin-right: -4px;
+`;
+
+const StyledThreeDotsButton = styled(ThreeDotsButton)`
+    fill: rgba(var(--center-channel-color-rgb), 0.56);
+`;
```
streamer45, line 161: Here and in other files, do you mind if we start using [transient props](https://styled-components.com/docs/faqs#transient-props-since-5.1) for styled components so we don't get DOM warnings? Prefixing passed props with `$` should be enough, same as I did in https://github.com/mattermost/mattermost-plugin-calls/pull/679 for the rest of them.

streamer45:
@cpoile Something weird is happening on my side. If I make a participant host from the popout and then leave the call (again from the popout) everything seems to crash and I am left in a weird state.

webapp/src/components/call_widget/component.tsx
```diff
@@ -985,139 +961,6 @@ export default class CallWidget extends React.PureComponent<Props, State> {
         );
     };
 
-    renderParticipantsList = () => {
```
streamer45, line 988: Bit by bit we'll get there :)

webapp/src/components/dot_menu/dot_menu.tsx
```diff
@@ -0,0 +1,222 @@
...
+        // @ts-ignore
+        <MenuButton
+            title={title}
+            isActive={(isActive ?? false) || isOpen}
+            onClick={(e: MouseEvent) => {
+                e.preventDefault();
+                e.stopPropagation();
+                toggleOpen();
+            }}
+            onKeyDown={(e: KeyboardEvent) => {
```
streamer45, line 98: Any reason for not using key up? Just feels more natural to me and it avoids spamming.

webapp/src/components/dot_menu/dropdown.tsx
```diff
@@ -0,0 +1,102 @@
...
+    flip?: Parameters<typeof flip>[0];
+    shift?: Parameters<typeof shift>[0];
+    focusManager?: boolean | Omit<ComponentProps<typeof FloatingFocusManager>, 'context' | 'children'>;
+    portal?: boolean;
+    strategy?: Strategy;
+    isOpen: boolean;
+    onOpenChange?: ((open: boolean) => void);
+};
+
+const Dropdown = (props: DropdownProps) => {
```
streamer45, line 40: Looks like a lot of work for a "mere" dropdown :(

webapp/src/components/dot_menu/dropdown.tsx
```diff
@@ -0,0 +1,102 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {Strategy} from '@floating-ui/react';
```
streamer45, line 4: I assume this is the new "proper" way of doing things?

streamer45 (commented):
Nice! Gave it a very quick pass and left some comments but still need to go through most of the new UI components.

matthewbirtch (changes requested):
Nice work @cpoile. A few things I noticed:

1. Seems like there might be too much padding at the top and bottom of the new menu. Is it 8px? It looks like it might be 10px. If so, can we change to 8?
2. When I open the host controls from the ••• button, if I move my mouse off the menu, it closes the menu and makes it so it is not very forgiving with mouse movements. It should stay open until I click away from the menu, click the ••• again, or click one of the menu items.
   

4. When hovering over each row in the participant list, can we add a hover state to the entire row if there are host controls available? Like this:
    image
3. When changing hosts from the popout window, the window in the background goes blank.

https://github.com/mattermost/mattermost-plugin-calls/assets/2040554/778cbfee-5c87-43a3-930a-b8c6930b5604

standalone/src/recording/components/recording_view/index.tsx
cpoile: Must be :)

webapp/src/components/call_widget/participant.tsx
cpoile: K, sounds good, done.

webapp/src/components/dot_menu/dropdown.tsx
cpoile: Yep. I've removed some of what we don't need, but I kept some that we aren't using right now in anticipation that we'll need it later.

webapp/src/components/dot_menu/dropdown.tsx
cpoile: Yep

streamer45 (commented):
Looks good, just some open questions on the API design.

server/api_router.go
```diff
@@ -93,6 +93,10 @@ func (p *Plugin) newAPIRouter() *mux.Router {
 	router.HandleFunc("/calls/{channel_id:[a-z0-9]{26}}/dismiss-notification", p.handleDismissNotification).Methods("POST")
 	router.HandleFunc("/calls/{call_id:[a-z0-9]{26}}/recording/{action}", p.handleRecordingAction).Methods("POST")
 
+	// Host Controls
+	hostCtrlRouter := router.PathPrefix("/calls/host").Subrouter()
+	hostCtrlRouter.HandleFunc("/{call_id:[a-z0-9]{26}}/make_host/{new_host_id:[a-z0-9]{26}}", p.handleMakeHost).Methods("POST")
```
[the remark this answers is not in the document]
cpoile: For this I think I would like to have a host subpath, just to make the subcommands make sense (eg mute, unmute).

server/host_controls_api.go
```diff
@@ -0,0 +1,25 @@
...
+	"net/http"
+)
+
+func (p *Plugin) handleMakeHost(w http.ResponseWriter, r *http.Request) {
+	var res httpResponse
+	defer p.httpAudit("handleMakeHost", &res, w, r)
+
+	userID := r.Header.Get("Mattermost-User-Id")
+	callID := mux.Vars(r)["call_id"]
+	newHostID := mux.Vars(r)["new_host_id"]
```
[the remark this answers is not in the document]
cpoile: Maybe we should chat more in 1:1 but I'm not sure how it's not really REST anymore... I would think the fact that it's a POST and I'm using url params is worse :). This could be a GET, except for the symantic that we're changing something, so maybe the reason to put it in JSON is simply because that's what you're "supposed" to do for POSTs? I dunno, the REST conventions are so arbitrary I can't keep track of it all. tbh, it's just simpler to put the params in the url here. 😝

cpoile:
@matthewbirtch Thanks for the review, I think I've found a clear and simple way to keep the menu around, and I've tried to match the button hover/active states, but there was a bit of ambiguity in the designs. Can you do some experimentation and let me know? 
Something I noticed, we should include some rules about name lengths and participant card sizes in the expanded view (they can get pretty big at the moment), and then add that to the UI cleanup work for 1.0.

server/api_router.go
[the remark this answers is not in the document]
streamer45: That's fine but then `make_host` would be redundant. Could we just use the verb?

How about `POST/PUT /calls/$CALL_ID/host/make/$HOST_ID`, `POST /calls/$CALL_ID/host/mute/$USER_ID` and so on? Still I feel the higher level grouping should be calls.

matthewbirtch:
> @matthewbirtch Thanks for the review, I think I've found a clear and simple way to keep the menu around, and I've tried to match the button hover/active states, but there was a bit of ambiguity in the designs. Can you do some experimentation and let me know? 

This is working much better now @cpoile, nice work! 

> Something I noticed, we should include some rules about name lengths and participant card sizes in the expanded view (they can get pretty big at the moment), and then add that to the UI cleanup work for 1.0.

Created a ticket here: https://mattermost.atlassian.net/browse/MM-57822

Feel free to add any extra details in the description

server/host_controls_api.go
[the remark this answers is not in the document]
streamer45: Apologies, I mistyped before. I meant to write `extensible` because having the payload makes it way easier to make non-breaking changes in the future. Agreed the URL looks cleaner this way though so not a big deal.

matthewbirtch (approved):

server/api_router.go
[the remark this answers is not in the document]
cpoile: Ah, I didn't understand what you meant. Yes, that works great, good idea.

server/host_controls_api.go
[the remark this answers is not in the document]
cpoile: Ahah, yes, that makes sense. :)

server/host_controls_api.go
```diff
@@ -0,0 +1,34 @@
...
+func (p *Plugin) handleMakeHost(w http.ResponseWriter, r *http.Request) {
+	var res httpResponse
+	defer p.httpAudit("handleMakeHost", &res, w, r)
+
+	userID := r.Header.Get("Mattermost-User-Id")
+	callID := mux.Vars(r)["call_id"]
+
+	var params struct {
+		NewHostID string `json:"new_host_id"`
+	}
```
streamer45, line 18: Fine for now but eventually should go into `public` so that it can be used by potentially automated or custom clients as well.

streamer45 (approved):
Looks great! Just one nit:

image

The hover portion is overflowing into the avatar. I'd expect some spacing there.

cpoile:
@matthewbirtch The overflow issue noted by Claudio is due to our not having min (or max) widths on the participant cards. Can you add that to the list of UI fixes for the card? (I'm not sure what we should have for min and max, if any).

matthewbirtch:
> The overflow issue noted by Claudio is due to our not having min (or max) widths on the participant cards. Can you add that to the list of UI fixes for the card? (I'm not sure what we should have for min and max, if any).

@cpoile I updated [this jira ticket](https://mattermost.atlassian.net/browse/MM-57822) with these details. Thanks for flagging.

PR #679, opened 2024-04-05
Migrate to styled components v6

streamer45:
#### Summary

PR migrates styled components from v5 to v6 to support container queries (to properly address https://github.com/mattermost/mattermost-plugin-calls/pull/675).

It was a bit more painful than ideal and some things just broke without warning (e.g. button hover states). Hopefully I caught them all but there totally could be something more I missed.


streamer45:
@cpoile That e2e test is starting to fail a bit too often. Could you have a look when you get a chance? May be a simple timeout issue.

cpoile (approved):

cpoile:
I've made a ticket for the flaky tests: https://mattermost.atlassian.net/browse/MM-57753

They don't seem to just be timeouts; they're failing in different ways each time. So I suspect something is not working reliably on the client during the tests.

PR #681, opened 2024-04-05
[MM-57737] Improve client side call state consistency

streamer45:
#### Summary

PR should fix most of the remaining call state data races originally mentioned in https://github.com/mattermost/mattermost-plugin-calls/pull/512, specifically when loading the pop out window and in case of a websocket reconnect.

This is done by implementing a websocket request handler that returns the calls state under lock. Using the websocket channel as opposed to making HTTP request should guarantee a valid sequencing of potentially concurrent events (e.g. users joining, leaving, etc).

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-57737

### DIFF
--- a/server/websocket.go
+++ b/server/websocket.go
@@ -888,6 +888,40 @@ func (p *Plugin) handleReconnect(userID, connID, channelID, originalConnID, prev
 	return nil
 }
 
+func (p *Plugin) handleCallStateRequest(channelID, userID, connID string) error {
+	// We should go through only if the user has permissions to the requested channel
+	// or if the user is the Calls bot.
+	if !(p.isBot(userID) || p.API.HasPermissionToChannel(userID, channelID, model.PermissionReadChannel)) {
+		return fmt.Errorf("forbidden")
+	}
+
+	// Locking is not ideal but it's the only way to guarantee a race free
+	// sequence and a consistent state.
+	// On the client we should make sure to make this request only when strictly
+	// necessary (i.e first load, joining call, reconnecting).
+	state, err := p.lockCallReturnState(channelID)
+	if err != nil {
+		return fmt.Errorf("failed to lock call: %w", err)
+	}
+	defer p.unlockCall(channelID)
+
+	if state == nil {
+		return fmt.Errorf("no call ongoing")
+	}
+
+	clientStateData, err := json.Marshal(state.getClientState(p.getBotID(), userID))
+	if err != nil {
+		return fmt.Errorf("failed to marshal client state: %w", err)
+	}
+
+	p.publishWebSocketEvent(wsEventCallState, map[string]interface{}{
+		"channel_id": channelID,
+		"call":       string(clientStateData),
+	}, &model.WebsocketBroadcast{ConnectionId: connID, ReliableClusterSend: true})
+
+	return nil
+}
+
 func (p *Plugin) WebSocketMessageHasBeenPosted(connID, userID string, req *model.WebSocketRequest) {
 	var msg clientMessage
 	msg.Type = strings.TrimPrefix(req.Action, wsActionPrefix)
@@ -896,10 +930,14 @@ func (p *Plugin) WebSocketMessageHasBeenPosted(connID, userID string, req *model
 	us := p.sessions[connID]
 	p.mut.RUnlock()
 
-	if msg.Type != clientMessageTypeJoin &&
-		msg.Type != clientMessageTypeLeave &&
-		msg.Type != clientMessageTypeReconnect && us == nil {
-		return
+	if us == nil {
+		// Only a few events don't require a user session to exist. For anything else
+		// we should return.
+		switch msg.Type {
+		case clientMessageTypeJoin, clientMessageTypeLeave, clientMessageTypeReconnect, clientMessageTypeCallState:
+		default:
+			return
+		}
 	}
 
 	if us != nil && !us.wsMsgLimiter.Allow() {
@@ -985,6 +1023,19 @@ func (p *Plugin) WebSocketMessageHasBeenPosted(connID, userID string, req *model
 			p.LogError(err.Error())
 		}
 
+		return
+	case clientMessageTypeCallState:
+		p.metrics.IncWebSocketEvent("in", "call_state")
+
+		channelID, _ := req.Data["channelID"].(string)
+		if channelID == "" {
+			p.LogError("missing channelID")
+			return
+		}
+
+		if err := p.handleCallStateRequest(channelID, userID, connID); err != nil {
+			p.LogError("handleCallStateRequest failed", "err", err.Error(), "userID", userID, "connID", connID)
+		}
 		return
 	case clientMessageTypeSDP:
 		msgData, ok := req.Data["data"].([]byte)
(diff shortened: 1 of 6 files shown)


webapp/src/index.tsx
```diff
@@ -714,59 +715,14 @@ export default class Plugin {
             }
         };
 
-        const fetchChannelData = async (channelID: string) => {
```
streamer45, line 717: We can finally get rid of this, it was totally redundant as we'd be fetching the same again in `fetchChannels`.

webapp/src/index.tsx
```diff
@@ -635,12 +631,17 @@ export default class Plugin {
...
+        const fetchChannels = async (skipChannelID?: string): Promise<AnyAction[]> => {
             const actions = [];
             try {
                 const data = await RestClient.fetch<CallChannelState[]>(`${getPluginPath()}/channels`, {method: 'get'});
 
                 for (let i = 0; i < data.length; i++) {
+                    if (skipChannelID === data[i].channel_id) {
+                        logDebug('skipping channel from state loading', skipChannelID);
+                        continue;
+                    }
```
streamer45, line 643: This is important as it will avoid messing the state for the current call which should only come from websocket from now on.

server/websocket.go
```diff
@@ -888,6 +888,40 @@ func (p *Plugin) handleReconnect(userID, connID, channelID, originalConnID, prev
...
+
+	// Locking is not ideal but it's the only way to guarantee a race free
+	// sequence and a consistent state.
+	// On the client we should make sure to make this request only when strictly
+	// necessary (i.e first load, joining call, reconnecting).
+	state, err := p.lockCallReturnState(channelID)
+	if err != nil {
+		return fmt.Errorf("failed to lock call: %w", err)
+	}
+	defer p.unlockCall(channelID)
```
cpoile, line 906: Would it make much of a difference if we immediately unocked? Looks like there's no reason no to, and it might reduce some of the non-idealness (and prevent the publish ws event from delaying the unlock).

server/websocket.go
```diff
@@ -896,10 +930,14 @@ func (p *Plugin) WebSocketMessageHasBeenPosted(connID, userID string, req *model
...
-		msg.Type != clientMessageTypeLeave &&
-		msg.Type != clientMessageTypeReconnect && us == nil {
-		return
+	if us == nil {
+		// Only a few events don't require a user session to exist. For anything else
+		// we should return.
+		switch msg.Type {
+		case clientMessageTypeJoin, clientMessageTypeLeave, clientMessageTypeReconnect, clientMessageTypeCallState:
+		default:
+			return
```
cpoile, line 939: As an aside, this is just bad language design. (I mean all c-derivatives here, not just go). Imagine you are a non-programmer (or a programmer who's used to fallthrough, or not used to fallthrough, really), is this quick to read through and know immediately what's happening? sheesh.

webapp/src/components/expanded_view/component.tsx
```diff
@@ -163,6 +163,8 @@ export default class ExpandedView extends React.PureComponent<Props, State> {
     private pushToTalk = false;
     private screenPlayer: HTMLVideoElement | null = null;
 
+    static contextType = window.ProductApi.WebSocketProvider;
```
cpoile, line 166: Huh, didn't know about this.

webapp/src/components/expanded_view/component.tsx
```diff
@@ -592,9 +594,16 @@ export default class ExpandedView extends React.PureComponent<Props, State> {
...
         if (!callsClient) {
+            logErr('callsClient should be defined');
             return;
         }
 
+        // On mount we request the call state through websocket. This avoids
+        // making a potentially racy HTTP call and should guarantee
+        // a consistent state.
+        logDebug('requesting call state through ws');
+        this.context.sendMessage('custom_com.mattermost.calls_call_state', {channelID: callsClient.channelID});
```
cpoile, line 605: I wonder if the websocket message isn't a bit confusing. Are you sending the call's state, or requesting it...?

webapp/src/index.tsx
cpoile: Can we add a comment in the code to remind us about the what and why? I can imagine forgetting and then spending time trying to reason through it.

webapp/src/index.tsx
cpoile: Awesome

webapp/src/index.tsx
```diff
@@ -776,18 +732,36 @@ export default class Plugin {
...
+            // from the ExpandedView component itself.
+            if (isCallsPopOut()) {
+                return;
             }
 
+            const currentCallChannelID = channelIDForCurrentCall(store.getState());
+
+            // We pass currentCallChannelID so that we
+            // can skip loading its state as a result of the HTTP calls in
+            // fetchChannels since it would be racy.
```
cpoile, line 746: ah, here it is. Maybe can we add a similar comment to the function also?

webapp/src/index.tsx
```diff
@@ -800,17 +774,29 @@ export default class Plugin {
...
 
-        this.registerWebSocketEvents(registry, store);
-        this.registerReconnectHandler(registry, store, () => {
-            logDebug('websocket reconnect handler');
-            if (!window.callsClient) {
-                logDebug('resetting state');
-                store.dispatch({
-                    type: UNINIT,
+        // A dummy React component so we can access webapp's
+        // WebSocket client through the provided hook. Just lovely.
```
cpoile, line 778: 😂

cpoile: Will this cause us to depend on a new min server version?

cpoile (commented):
This is kind of a weird one isn't it. But if it works 🤷  :)

webapp/src/index.tsx
streamer45: This has been there for a long time (before monorepo) so I think we are good. But of course I'd like to fix it properly one day (pass the client object directly to the init function) in which case we'll have to add some backwards compatibility check.

webapp/src/components/expanded_view/component.tsx
streamer45: Just being consistent with other events there. We are using the direction (from/to client) to implicitly define whether it's request or response. I know you don't love it :p

server/websocket.go
streamer45: We have ways to get the call state without locking but the point here is that we need to queue the event to be sent through WS before unlocking otherwise we are subject to a race again.

streamer45:
> This is kind of a weird one isn't it.

Why weird? It's a rather simple race condition, our redux state is consistent only if events are received and dispatched in the order they are generated on the server side. With a mixture of HTTP calls and websockets both affecting state we make it racy by design :p

webapp/src/components/expanded_view/component.tsx
streamer45: I guess HTTP fixes the confusion through the Method. Here we use a bit of context as it doesn't make any sense for the client to ever send the call state. Please let me know if this is blocking :)

server/websocket.go
cpoile: Ah, of course 🤦

server/websocket.go
cpoile: Nah, no big deal, just complaining.

webapp/src/components/expanded_view/component.tsx
cpoile: Nope, not blocking :)

cpoile (approved):
Thanks!

PR #684, opened 2024-04-09
[MM-57250] Add spinner animation to start/join call button

streamer45:
#### Summary

PR implements a spinner animation in the start/join button in channel header when connecting to a call. I also updated the call widget to be consistent with this behaviour.

@abhijit-singh One detail to note is that I have not implemented a different message based on whether the user is starting or joining a call. This is because it's tricky to track that state properly. If we were to do it, the button text would end up flickering from `Starting call…` to `Joining call…` because in reality there are two connections behind the process and the one controlling the messaging is the quickest to be made (websocket). So I defaulted to the more generic `Joining call…`. Please let me know if there are concerns of if you find it to be a good enough compromise.

Designs at https://www.figma.com/file/UUuEi45kKku8hq5JQA7tDC/MM-57250-Provide-feedback-after-user-clicks-start-call?type=design&node-id=1-7&mode=design&t=MXQydycXqdZnW7Iw-0

#### Video

https://github.com/mattermost/mattermost-plugin-calls/assets/1832946/4f802888-0f9b-4c67-ba16-90dc7a04212f

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-57250


cpoile (commented):
This is awesome, thanks! 

I put as a comment because I tried on desktop, but it doesn't seem to be working...? (maybe it's something on my end?)

streamer45:
> This is awesome, thanks!
> 
> I put as a comment because I tried on desktop, but it doesn't seem to be working...? (maybe it's something on my end?)

Let me check. I had to port the changes to global widget but only tested it from a browser.

streamer45:
Oh I guess it's the button that needs addressing on Desktop because it won't be in sync.

streamer45:
Thanks @cpoile , a clear oversight. Please give it a try again.

cpoile (approved):
Awesome, works perfectly.
I like the `Joining call...` even when starting. Makes it feel like the call was started on the server and now we're joining it.

abhijit-singh (approved):
Looks good @streamer45! 

Would it be feasible to make the error message a little more specific in case the connection times out and the call fails? Here's what it could be:
image

streamer45:
@abhijit-singh Yes that should be doable although I'd still retain the link to the troubleshooting section.

streamer45:
Added a new error case (and message) to show in case the connection timed out.

image

PR #694, opened 2024-04-11
Make update and delete store operations idempotent

streamer45:
#### Summary

Relying on [`RowsAffected()`](https://pkg.go.dev/database/sql#Result) is just not going to work. Making the operations idempotent so that it doesn't matter if the state is the same. We can look into some optimizations later on.

### DIFF
--- a/server/db/calls_channels_store_test.go
+++ b/server/db/calls_channels_store_test.go
@@ -64,13 +64,6 @@ func testUpdateCallsChannel(t *testing.T, store *Store) {
 		require.EqualError(t, err, "invalid channel: should not be nil")
 	})
 
-	t.Run("missing", func(t *testing.T) {
-		err := store.UpdateCallsChannel(&public.CallsChannel{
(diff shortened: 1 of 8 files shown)


server/db/calls_channels_store_test.go
```diff
@@ -64,13 +64,6 @@ func testUpdateCallsChannel(t *testing.T, store *Store) {
 		require.EqualError(t, err, "invalid channel: should not be nil")
 	})
 
-	t.Run("missing", func(t *testing.T) {
-		err := store.UpdateCallsChannel(&public.CallsChannel{
-			ChannelID: "channelID",
-		})
-		require.EqualError(t, err, "failed to update calls channel")
-	})
```
cpoile, line 72: Why are we removing the "missing" tests here and elsewhere?

cpoile (commented):
Yep, sounds good.

cpoile (approved):

PR #698, opened 2024-04-17
feat: Initial commit for e2e testing migration

toninis:
#### Summary

Complete migration of e2e testing on Github

- Executed upon PR
- Created check `e2e/report-summary` with the S3 report URL redirection
- Made parallelism configurable
  - It's controlled by the env var in the pipeline
  - Modified the node merge tool to dynamically import results
- Added some more verbosity to the execution 
- Easily accessible env vars for all the components required

Once merged I will remove the related Gitlab repositories. 
No more waiting 😄 🚀 

#### Ticket Link

https://mattermost.atlassian.net/browse/CLD-6654


saturninoabril (approved):
Wow, thanks @toninis and all good to me!

One thing I observed is that this still triggered GitLab CI, I know you mentioned that related GitLab repo will be removed but I wonder if [.gitlab-ci.yml](https://github.com/mattermost/mattermost-plugin-calls/blob/68c9a9348ac79e81a7f4784d7bb09f43b42e556d/.gitlab-ci.yml) should be removed together in this PR.
Screenshot 2024-04-19 at 8 19 34 PM

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,275 @@
...
+          mkdir -p ${{ github.workspace }}/config
+          mkdir -p ${{ github.workspace }}/dotenv
+
+          ## Navigate to mattermost server build folder to spawn the environment
+          cd mattermost/server/build/
+          ${{ github.workspace }}/e2e/scripts/prepare-server.sh
+          cd -
+
+          ## Add extra environment variables for mattermost server
+          echo "MM_LICENSE=${{ secrets.MM_PLUGIN_CALLS_TEST_LICENSE }}" >> dotenv/app.private.env
```
streamer45, line 138: For my own understanding, where are these stored? Do we have a private repo somewhere?

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,275 @@
...
+
+          # Spawn mattermost server
+          echo "Spawning mattermost server ... "
+          docker run -d --quiet --name ${CONTAINER_SERVER} \
+          --net ${DOCKER_NETWORK} \
+          --net-alias mm-server \
+          --user mattermost:mattermost \
+          --env-file="${{ github.workspace }}/dotenv/app.private.env" \
+          -v ${{ github.workspace }}/config:/mattermost/config:rw \
+          -v ${{ github.workspace }}/logs:/mattermost/logs:rw \
```
streamer45, line 153: Is there an easy way to access server logs as an artifact? Just asking since it's one of the first thing I used to look at in case of failure.

e2e/scripts/prepare-server.sh
```diff
@@ -0,0 +1,42 @@
...
+  --env "LOGGER_CONSOLELEVEL=DEBUG" \
+  --env "DEV_MODE=true" \
+  --env "DOCKER_NETWORK=${DOCKER_NETWORK}" \
+  --network-alias=calls-offloader ${IMAGE_CALLS_OFFLOADER}
+
+# Check that calls-offloader is up and ready
+docker run --rm --quiet --name "${COMPOSE_PROJECT_NAME}_curl_callsoffloader" --net ${DOCKER_NETWORK} ${IMAGE_CURL} sh -c "until curl -fs http://calls-offloader:4545/version; do echo Waiting for calls-offloader; sleep 5; done; echo calls-offloader is up"
+
+# Check that elasticsearch is ready
+docker run --rm --quiet --name "${COMPOSE_PROJECT_NAME}_curl_elasticsearch" --net ${DOCKER_NETWORK} ${IMAGE_CURL} sh -c "until curl --max-time 5 --output - http://elasticsearch:9200; do echo Waiting for elasticsearch; sleep 5; done; echo elasticsearch is up"
```
streamer45, line 42: Now I wonder why this was in the Gitlab pipeline in the first place. We certainly don't need elastic search to run Calls tests.

e2e/scripts/run.sh
```diff
@@ -0,0 +1,46 @@
...
+echo "Installing calls ..."
+docker exec \
+  ${CONTAINER_SERVER} \
+  sh -c "/mattermost/bin/mmctl --local plugin add bin/calls"
+
+# Generates a sysadmin that Playwright can use
+echo "Generating sample data with mmctl ..."
+docker exec \
+  ${CONTAINER_SERVER} \
+  sh -c "/mattermost/bin/mmctl --local sampledata"
```
streamer45, line 25: Note to self: I think we just need the admin as the tests themselves will create any team/channel/user as needed so generating the whole sample data is just more work for not much benefit I think.

streamer45 (approved):
Excellent work @toninis :100: Ship it!

e2e/scripts/prepare-server.sh
saturninoabril: I believe it's patterned from the server/webapp E2E setup (cypress-ui-automation repo), and yes, this can be removed safely.

One option is the setup made by @mvitale1989 in mattermost repo where `ENABLED_DOCKER_SERVICES` is used to spin up only the services needed by the tests ([link](https://github.com/mattermost/mattermost/blob/master/e2e-tests/.ci/server.generate.sh) to script).

.github/workflows/e2e.yml
toninis: All our secrets are stored as organisation secrets. Unfortunately you do not have access there . 
Once we migrate to our secret management solution everything will be easier to handle

.github/workflows/e2e.yml
toninis: I will create an extra step to upload also server logs in case of failure 👍🏼

e2e/scripts/prepare-server.sh
toninis: I guess the reason is that you spawn the docker-compose with ES in place. If we do not need ES at all we can skipp this check or change the docker-compose completely to our needs .

PR #699, opened 2024-04-17
[MM-54335] Instrument code to monitor app and store performance

streamer45:
#### Summary

PR instruments the code with a couple more metrics to track both app and store performance. Store should be covered completely while for the app side I am currently tracking only a few key methods. May likely add more as I assess overall performance at scale.


streamer45:
This is how it could look in practice:

image

cpoile (approved):
Very cool! A little worried about createCall taking infinite years, but I guess we can work on that now that we know.

PR #700, opened 2024-04-17
[MM-57323] Update node/npm to match webapp

cpoile:
#### Summary
- Update node/npm
- Update webapp to March 18 (when the update was done on that side)
- Update some npm packages to keep pace with mattermost
- Fix webpack errors (yay)
- Clean up types

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-57323


cpoile:
@streamer45 See, stats change our behaviour and our reactions. 
I vote we dump stats altogether.

streamer45:
Your performance is defined by those stats so better keep them if you want to be around for longer :p

webapp/src/actions.ts
```diff
@@ -369,7 +372,7 @@ export function incomingCallOnChannel(channelID: string, callID: string, callerI
             await dispatch(getProfilesByIdsAction([callerID]));
         }
 
-        await dispatch({
```
streamer45, line 372: So dispatching is always synchronous or there's more to it?

webapp/webpack.config.js
```diff
@@ -17,7 +17,7 @@ if (NPM_TARGET === 'debug' || NPM_TARGET === 'debug:watch') {
 
 const plugins = [
     new webpack.ProvidePlugin({
-        process: 'process/browser',
+        process: 'process/browser.js',
```
streamer45, line 20: Hope this wasn't too painful to figure out :grimacing:

streamer45 (approved):
Thanks, apologies for the delay on this.

streamer45:
@cpoile One thing I noticed is that my local env is generating a different `package-lock.json` when installing this.

```
github.com/mattermost/mattermost-plugin-calls » node -v
v20.11.1
github.com/mattermost/mattermost-plugin-calls » npm -v
10.2.4
```

webapp/src/actions.ts
cpoile: Nope, dispatching is async, it's just that putting await dispatch at the return is useless/redundant (since the function returns the promise immediately, so awaiting it does nothing).

webapp/webpack.config.js
cpoile: It was touch and go there for a bit (my sanity)

cpoile:
@streamer45 Try now -- I might have been running into an issue where I was running `npm i` with the old webapp, then the new webapp gets installed, and then I should have run it again (since there's new code now for np meto run against). I did that and found it was then failing on typescript incompatibilities. Upgraded that, and then react-intl is showing some type errors... But it's still working (and we're using it exactly as the docs show it -- and even those docs examples throw this new error), so looks like the typings are broken at the moment (or there's an interaction with the typescript version. Anyway, I ignored them for now because I couldn't find a fix after working on it for way too long.

streamer45 (approved):
Looks good, let's get it merged because all this moving from one version to another locally is driving me nuts :p

PR #702, opened 2024-04-18
[MM-57907] Mute participant

cpoile:
#### Summary
- Implements the mute participant host control
- Pretty straightforward, doesn't show if yourself or if the participant is muted already.

image

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-57907

### DIFF
--- a/server/host_controls.go
+++ b/server/host_controls.go
@@ -58,11 +58,10 @@ func (p *Plugin) changeHost(requesterID, channelID, newHostID string) error {
 }
 
 func (p *Plugin) muteSession(requesterID, channelID, sessionID string) error {
-	state, err := p.lockCallReturnState(channelID)
+	state, err := p.getCallState(channelID, false)
 	if err != nil {
-		return fmt.Errorf("failed to lock call: %w", err)
+		return err
 	}
-	defer p.unlockCall(channelID)
 
 	if state == nil {
 		return errors.New("no call ongoing")
--- a/server/host_controls.go
+++ b/server/host_controls.go
@@ -69,7 +69,7 @@ func (p *Plugin) muteSession(requesterID, channelID, sessionID string) error {
 
 	if requesterID != state.Call.GetHostID() {
 		if isAdmin := p.API.HasPermissionTo(requesterID, model.PermissionManageSystem); !isAdmin {
-			return errors.New("no permissions to change host")
+			return errors.New("no permissions to mute session")
 		}
 	}
 
--- a/server/host_controls.go
+++ b/server/host_controls.go
@@ -78,10 +78,6 @@ func (p *Plugin) muteSession(requesterID, channelID, sessionID string) error {
 		return errors.New("session is not in the call")
 	}
 
-	if p.isBot(ust.UserID) {
-		return errors.New("cannot mute the bot")
-	}
-
 	if !ust.Unmuted {
 		return nil
 	}
--- a/server/host_controls.go
+++ b/server/host_controls.go
@@ -83,7 +83,7 @@ func (p *Plugin) muteSession(requesterID, channelID, sessionID string) error {
 	}
 
 	if !ust.Unmuted {
-		return errors.New("user is already muted")
+		return nil
 	}
 
 	p.publishWebSocketEvent(wsEventHostMute, map[string]interface{}{
--- a/server/host_controls.go
+++ b/server/host_controls.go
@@ -56,3 +56,42 @@ func (p *Plugin) changeHost(requesterID, channelID, newHostID string) error {
 
 	return nil
 }
+
+func (p *Plugin) muteSession(requesterID, channelID, sessionID string) error {
+	state, err := p.lockCallReturnState(channelID)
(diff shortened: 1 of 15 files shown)


server/host_controls.go
```diff
@@ -56,3 +56,41 @@ func (p *Plugin) changeHost(requesterID, channelID, newHostID string) error {
...
+
+	ust, ok := state.sessions[sessionID]
+
+	if !ok {
+		return errors.New("session is not in the call")
+	}
+
+	if p.isBot(ust.UserID) {
+		return errors.New("cannot mute the bot")
+	}
```
streamer45, line 83: Is this reflected in the client side interface as well?

server/host_controls.go
```diff
@@ -56,3 +56,40 @@ func (p *Plugin) changeHost(requesterID, channelID, newHostID string) error {
...
+
+	if !ust.Unmuted {
+		return errors.New("user is already muted")
+	}
```
streamer45, line 87: Not sure, maybe I'd keep it idempotent and just return nil. We are not doing this under lock so it's possible this would happen under normal circumstances.

webapp/src/index.tsx
```diff
@@ -243,6 +244,10 @@ export default class Plugin {
         registry.registerWebSocketEventHandler(`custom_${pluginId}_caption`, (ev) => {
             handleCaption(store, ev);
         });
+
+        registry.registerWebSocketEventHandler(`custom_${pluginId}_host_mute`, (ev) => {
+            handleHostMute(store, ev);
+        });
```
streamer45, line 250: Do we need to update the standalone side as well?

streamer45 (commented):
Nice! Left some comments.

server/host_controls.go
```diff
@@ -56,3 +56,41 @@ func (p *Plugin) changeHost(requesterID, channelID, newHostID string) error {
 
 	return nil
 }
+
+func (p *Plugin) muteSession(requesterID, channelID, sessionID string) error {
+	state, err := p.getCallState(channelID, false)
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
cpoile: Ok, let me add a ticket for optimizing these calls after the controls are done -- there could be a "host controls subset" call. https://mattermost.atlassian.net/browse/MM-57922

server/host_controls_api.go
```diff
@@ -33,3 +33,30 @@ func (p *Plugin) handleMakeHost(w http.ResponseWriter, r *http.Request) {
 	res.Code = http.StatusOK
 	res.Msg = "success"
 }
+
+func (p *Plugin) handleMuteSession(w http.ResponseWriter, r *http.Request) {
+	var res httpResponse
+	defer p.httpAudit("handleMuteSession", &res, w, r)
+
+	userID := r.Header.Get("Mattermost-User-Id")
+	callID := mux.Vars(r)["call_id"]
```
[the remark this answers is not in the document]
cpoile: It's not the typing, it's the ugliness 😛 But seriously, I agree it's better.

webapp/src/websocket_handlers.ts
```diff
@@ -443,3 +444,18 @@ export function handleCaption(store: Store, ev: WebSocketMessage<LiveCaptionData
...
 
+// TODO: MM-57919, refactor wsmsg data to calls-common
+export function handleHostMute(store: Store, ev: WebSocketMessage<{ channel_id: string, session_id: string }>) {
+    const channelID = ev.data.channel_id;
+    const client = getCallsClient();
+    if (!client || client?.channelID !== channelID) {
+        return;
+    }
+
+    const sessionID = getCallsClientSessionID();
```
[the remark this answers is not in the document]
cpoile: 🤦

server/host_controls.go
cpoile: I don't think it needs to be -- we don't see the bots in the call, do we?

server/host_controls.go
cpoile: Ah good point, done.

webapp/src/index.tsx
cpoile: I don't think so -- the bot won't ever receive the message, right?

server/host_controls.go
streamer45: Right, so why the check? :p They are unable to speak at the moment anyway ehe

webapp/src/index.tsx
streamer45: Global widget would though

server/host_controls.go
cpoile: This was one of those "I may as well, or else I'll get called on it" defensive coding things. :P I'll take it out, np.

webapp/src/index.tsx
cpoile: 😂 Can't believe I missed that. Thanks.

streamer45 (approved):
Nicely done! Would love to see a playwright test eventually :)

PR #704, opened 2024-04-18
[MM-54335] Batching of call join/leave operations

streamer45:
#### Summary

Preliminary performance tests are showing significant struggle to handle high frequencies of join/leave events in larger calls (100+ participants). This is mainly due to a couple of reasons:

1. High lock contention, both local (process) and global (cluster) as all these events need to go through serially.
2. Having to fetch the entire call state as part of 1.

PR proposes a way to reduce the number of these expensive operations through batching. This means queuing the operations and running them asynchronously in batches, at a controlled interval rate instead of as soon as they are received. 

Initial tests are showing good results, especially when considering the worst (but possible) case scenario of all participants leaving the call around the same time.


server/session.go
```diff
@@ -168,16 +173,31 @@ func (p *Plugin) addUserSession(state *callState, callsEnabled *bool, userID, co
...
+			if retErr == nil {
+				p.publishWebSocketEvent(wsEventCallHostChanged, map[string]interface{}{
+					"hostID": newHostID,
+				}, &WebSocketBroadcast{
+					ChannelID:           channelID,
+					ReliableClusterSend: true,
+					UserIDs:             getUserIDsFromSessions(state.sessions),
+				})
+			} else {
+				// Undo in case of error
```
cpoile, line 197: So this means we have to hang on to err and undo all state changes now? 
It seems like adding yet another complication... Is there any way around it? Maybe a "change state or rollback" to make it simpler?

server/session.go
```diff
@@ -466,21 +504,87 @@ func (p *Plugin) removeSession(us *session) error {
...
+
+		if err := p.removeUserSession(state, us.userID, us.originalConnID, us.connID, us.channelID); err != nil {
+			p.LogError("failed to remove user session ", "originalConnID", us.originalConnID, "err", err.Error())
+		}
+	}
 
 	p.mut.Lock()
-	delete(p.sessions, us.connID)
+	batcher := p.removeSessionsBatchers[us.channelID]
+	shouldBatch := batcher != nil || sessionsCount >= minMembersCountForBatching
```
cpoile, line 526: So only batch when we're past the `minMembersCountForBatching` threshold, but I guess there's no way to tell if you will actually need the batching ahead of time, right?  (I can't think of a way, other than adding an 'end call' button... But even then it wouldn't detect the mass leaving just before.) Just thinking out loud.

server/websocket.go
```diff
@@ -667,190 +685,278 @@ func (p *Plugin) handleJoin(userID, connID, authSessionID string, joinData Calls
...
-		}, &WebSocketBroadcast{
-			ChannelID:           channelID,
-			ReliableClusterSend: true,
-			UserIDs:             getUserIDsFromSessions(state.sessions),
-		})
+		return nil
 	}
 
-	clientStateData, err := json.Marshal(state.getClientState(p.getBotID(), userID))
+	// Non-batching case
```
cpoile, line 944: From the community data and load tests do you have any idea how much a problem quick-joining is? I can see quick-ending  being a legit problem, but I don't think I've seen a call where people joined all at once. 
That's two separate questions I guess: 1) is quick-joining a serious technical problem, and 2) is quick-joining a real-life problem.

cpoile (approved):
This is a neat solution, I really like it technically. Nice work! I just have some non-blocking questions.

server/batching/batcher.go
```diff
@@ -0,0 +1,113 @@
...
+	stopCh  chan struct{}
+	doneCh  chan struct{}
+	batches int
+}
+
+type Config struct {
+	// The frequency at which batches should be executed.
+	Interval time.Duration
+	// The maximum size of the queue of items.
+	Size int
```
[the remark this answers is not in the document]
streamer45: Well, channels must have a size, unless unbuffered which wouldn't work here as we need a queue. The practical use is to have a bound on resources we allocate. If we are unable to push because the channel is full it likely means we are under-provisioning and something is clearly wrong.

server/session.go
streamer45: I know but the simpler alternative would be to deep clone the whole state and return the original in case of failure. Do we really want to do that every time just to cover an error case that shouldn't happen under normal circumstances? Happy discuss different solutions of course, just couldn't think of anything better.

server/session.go
streamer45: The leaving case is actually easier since we know how many sessions we have connected and can assume eventually they will disconnect roughly at the same time as that's just the nature of large (planned) calls, likely within a tight schedule.

When joining the call it's harder because we have no idea, see the comment I left:

https://github.com/mattermost/mattermost-plugin-calls/blob/c2623a7b86f0aeeda629bf969613b6e68b7ce82e/server/websocket.go#L892-L895

server/websocket.go
streamer45: I think it is as soon as you scale up due to the nature of meetings. They are planned in advance, have a fixed date/time and people will naturally join around that time. If you have enough people (e.g. hundreds) they will likely ramp up very quickly in the first couple of minutes.

server/session.go
cpoile: It's up to you of course, and I'm not sure of the feasibility and cost of doing that. But I guess I would say as devil's advocate: consider a) the mental cost of having to double check if this is a place where we need to catch errors and revert changes + do that manually, vs b) the cost of auto clone + reverting.

streamer45: It's a fair point. I'll run some benchmarks as there's a good chance cloning performance to be a blip in the whole context of why we are doing this.

streamer45: @cpoile Take a look. I made the change as state cloning performance shouldn't be a concern.

```
goos: linux
goarch: amd64
pkg: github.com/mattermost/mattermost-plugin-calls/server
cpu: 11th Gen Intel(R) Core(TM) i7-1185G7 @ 3.00GHz
BenchmarkCallStateClone
BenchmarkCallStateClone/10_sessions
BenchmarkCallStateClone/10_sessions-8         	  568357	      1998 ns/op	    2020 B/op	      16 allocs/op
BenchmarkCallStateClone/100_sessions
BenchmarkCallStateClone/100_sessions-8        	   77806	     15612 ns/op	   17016 B/op	     107 allocs/op
BenchmarkCallStateClone/1000_sessions
BenchmarkCallStateClone/1000_sessions-8       	    8863	    127430 ns/op	  138168 B/op	    1007 allocs/op
BenchmarkCallStateClone/10000_sessions
BenchmarkCallStateClone/10000_sessions-8      	     800	   1358737 ns/op	 1259576 B/op	   10007 allocs/op
BenchmarkCallStateClone/0_sessions
BenchmarkCallStateClone/0_sessions-8          	 2141230	       586.3 ns/op	     800 B/op	       5 allocs/op
PASS
ok  	github.com/mattermost/mattermost-plugin-calls/server	73.514s
```

PR #705, opened 2024-04-19
[MM-57464] Host controls: Lower hand; host controls notifications

cpoile:
#### Summary
- This one is a bit more complicated because I'm adding a host controls notification to the expanded view and the widget (widget wasn't in the designs but I figured it would be wanted). I chose it to stay for 5s (changeable if you want), and added animation for the expanded view.
- I'm getting the display name in the handler so that we avoid the problem of a host doing something and then leaving the call and their name becoming "someone".
- They stack (as you see from me playing around in the screenshots), for that edge-case. The host control message takes precedence over others, since it seems time-sensitive and more important.

| the menu command | fooling around with stacking | raising hand again after it being lowered |
|--|--|--|
| image | image | image |

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-57464


matthewbirtch (commented):
This is awesome @cpoile. Nice work. The only thing I noticed is that in the popout window, the 'user raised hand' notice jumps around strangely when it gets cleared and the 'host lowered your hand' notice animates in (see video below). This seems to happen in both the desktop app and browser. Can we make that cleaner?

https://github.com/mattermost/mattermost-plugin-calls/assets/2040554/b399c819-1b8c-45ce-a699-924bfe06b171

cpoile:
Thanks @matthewbirtch , I've fixed it, let me know if this works for you. I can delay the "host lowered your hand" chip even more if you want there to be a moment before it shows.

matthewbirtch (approved):
Looks great to me @cpoile!

webapp/src/websocket_handlers.ts
```diff
@@ -473,3 +476,53 @@ export function handleHostScreenOff(store: Store, ev: WebSocketMessage<{ channel
...
+
+    // Put the notification on the end of the event loop so that unraiseHand can be processed before
+    // we continue. This prevents the "raised hand" and "host has lowered your hand" reaction chips
+    // from being shown at the same time.
+    setTimeout(() => {
+        store.dispatch({
+            type: HOST_CONTROL_NOTIFICATION,
+            data: hostNotification,
+        });
+    }, 0);
```
streamer45, line 517: Not sure I fully understood. Is this conflicting with a future websocket event?

cpoile: No, it just that if you send it right now there will be two reaction chips at the same time: the "you are raising your hand" + the hostControlNotification. That was causing a jump because the hostControlNotifcation would animate in.

server/websocket.go
```diff
@@ -51,6 +51,7 @@ const (
 	wsEventJobStop                   = "job_stop"
 	wsEventHostMute                  = "host_mute"
 	wsEventHostScreenOff             = "host_screen_off"
+	wsEventHostUnraiseHand           = "host_unraise_hand"
```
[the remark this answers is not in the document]
cpoile: Oh thank goodness. :)

webapp/src/selectors.ts
```diff
@@ -259,6 +261,18 @@ export const recordingForCurrentCall: (state: GlobalState) => CallJobReduxState
...
+};
+
+export const hostControlNotificationsForCurrentCall: (state: GlobalState) => HostControlNotification[] =
+    createSelector(
+        'hostControlNotificationsForCurrentCall',
+        hostControlNotificationsForCalls,
+        idForCurrentCall,
+        (notifications, id) => (id ? notifications[id] || [] : []),
+    );
+
```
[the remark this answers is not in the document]
streamer45: Sorry, I meant in general, the type name, component names, and everything related to this feature. These are not really notifications in the strict sense (e.g. desktop/browser/push).

server/api_router.go
```diff
@@ -98,6 +98,7 @@ func (p *Plugin) newAPIRouter() *mux.Router {
 	hostCtrlRouter.HandleFunc("/make", p.handleMakeHost).Methods("POST")
 	hostCtrlRouter.HandleFunc("/mute", p.handleMuteSession).Methods("POST")
 	hostCtrlRouter.HandleFunc("/screen-off", p.handleScreenOff).Methods("POST")
+	hostCtrlRouter.HandleFunc("/unraise-hand", p.handleUnraiseHand).Methods("POST")
```
streamer45, line 101: I'd also expect this to become `lower-hand`, and in the new methods as well.

webapp/src/components/expanded_view/component.tsx
```diff
@@ -1350,6 +1350,7 @@ const ReactionOverlay = styled.div`
     display: flex;
     flex-direction: column;
     gap: 12px;
+    pointer-events: none;
```
streamer45, line 1353: I guess this is what you mentioned before. Will events pass through without altering the stack order?

server/api_router.go
cpoile: 🤦

webapp/src/selectors.ts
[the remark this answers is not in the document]
cpoile: talked about in channel -- can't think of a more appropriate name, so vive la confusion.

webapp/src/components/expanded_view/component.tsx
cpoile: Yep, which is what we want -- we don't want to mess with the z-index of the reactions or notifications or error message boxes, those have to be on top, but we don't want the div they're on to intercept the pointer events.

streamer45: Cool, will need to merge this into my work as well, thanks.

webapp/src/actions.ts
```diff
@@ -596,3 +596,13 @@ export const stopScreenshare = async (callID: string, sessionID: string) => {
         },
     );
 };
+
+export const unraiseHand = async (callID: string, sessionID: string) => {
```
streamer45, line 600: -> `lowerHand`, you asked for it :)

streamer45 (approved):
Nice work!

PR #707, opened 2024-04-22
[MM-57965] fix Websocket reconnect handler

cpoile:
#### Summary
- The `websocket reconnect handler` was causing the state to be reset every minute or so when the window is in the background. So if you're on a call and you put the expanded view in the background, it would be blank when you come back after a reset.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-57965


streamer45 (approved):
Thanks, my bad. Pushed an e2e test to cover this.

cpoile:
Awesome, thanks @streamer45!

PR #709, opened 2024-04-22
[MM-57967] Add HostControlsAllowed

cpoile:
#### Summary
- Mostly for mobile, but using it for webapp too.
- I'm adding the field outside the configuration file because we need the license to get the value. (And because of that, it doesn't seem to fit in the configuration file, which seems to be more for the configuration that can be set/changed etc.).

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-57967


server/api.go
```diff
@@ -507,11 +507,13 @@ func (p *Plugin) handleConfig(w http.ResponseWriter) error {
...
-		SkuShortName string `json:"sku_short_name"`
+		SkuShortName        string `json:"sku_short_name"`
+		HostControlsAllowed bool
 	}
 	ret := config{
-		clientConfig: p.getConfiguration().getClientConfig(),
-		SkuShortName: skuShortName,
+		clientConfig:        p.getConfiguration().getClientConfig(),
+		SkuShortName:        skuShortName,
+		HostControlsAllowed: p.licenseChecker.HostControlsAllowed(),
```
streamer45, line 516: Maybe we could consider moving all of this into `getClientConfig` instead of having yet another wrapping object.

streamer45 (approved):

server/api.go
cpoile: @streamer45 Let me know what you think of doing it this way

server/api.go
```diff
@@ -499,22 +499,7 @@ func (p *Plugin) handleGetTURNCredentials(w http.ResponseWriter, r *http.Request
...
-		clientConfig
-		SkuShortName        string `json:"sku_short_name"`
-		HostControlsAllowed bool
-	}
-	ret := config{
-		clientConfig:        p.getConfiguration().getClientConfig(),
-		SkuShortName:        skuShortName,
-		HostControlsAllowed: p.licenseChecker.HostControlsAllowed(),
-	}
+	ret := p.getClientConfig()
```
streamer45, line 502: No need to instantiate a variable I suppose, we could just feed it to the encoder?

server/api.go
streamer45: Ah forgot we needed the API to properly calculate it (hence the wrapper). But yeah, making it a plugin method works fine, thanks!

streamer45 (approved):

PR #710, opened 2024-04-22
Run e2e tests on tags and main branch

streamer45:
#### Summary

Just noticed we were not running these on merging to `main`. I think we should, and on tags (releases) as well.


cpoile (approved):
Awesome, yes agreed.

PR #711, opened 2024-04-23
[MM-57954] Use direct db connection for writer

streamer45:
#### Summary

PR makes the store open a connection directly to the database, bypassing the RPC layer. This is necessary as we identified a significant performance issue in fetching the call state (sessions specifically) due to the overhead of RPC calls. 

I am currently only changing the writer connection for two main reasons:

1. The existing bottleneck is when fetching the state under lock, which should only happen on the writer node.
2. Implementing proper replicas management on this side is more complex as it requires both load balancing and keeping health checks in case they go down, all things that are nicely implemented by the server already.

#### Screenshots

*Before*

image

*After*

image

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-57954


server/db/utils.go
```diff
@@ -25,10 +27,22 @@ func (s *Store) setupConnForMigrations(dsn string) (*sql.DB, error) {
 		return nil, fmt.Errorf("failed to ping DB: %w", err)
 	}
 
-	db.SetMaxIdleConns(*s.settings.MaxIdleConns)
-	db.SetMaxOpenConns(*s.settings.MaxOpenConns)
-	db.SetConnMaxLifetime(time.Duration(*s.settings.ConnMaxLifetimeMilliseconds) * time.Millisecond)
-	db.SetConnMaxIdleTime(time.Duration(*s.settings.ConnMaxIdleTimeMilliseconds) * time.Millisecond)
+	maxIdleConns := max(*s.settings.MaxIdleConns/10, 5)
+	maxOpenConns := max(*s.settings.MaxOpenConns/10, 10)
```
streamer45, line 31: May need to review these as we run performance tests. Keeping a conservative approach for now to avoid having the plugin potentially overload the db,

cpoile (approved):
Nice improvement. 🚀

PR #713, opened 2024-04-24
[MM-58006] System console live captions and transcriber settings visibility

cpoile:
#### Summary
- This is a short-term fix until we do the full redesign, but at least now things are visible only if their prerequisites are enabled. Before, some were and some weren't; it was a bit incorrect (noticed when I was doing the docs).

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-58006


streamer45:
Is it expected that if a parent setting (e.g. recording) is disabled but a child setting (e.g. transcription) is enabled then it will still show? 

image

cpoile:
@streamer45 That's literally what this PR is supposed to fix. 😄 Can you try reinstalling and/or refreshing? It's works correctly for me.

streamer45 (approved):
Yeah, it was installed locally but I picked the wrong tab (call test server :man_facepalming:)

Looks good!

PR #714, opened 2024-04-24
[MM-58002] Ensure call state mutating operations are executed in the right order

streamer45:
#### Summary

Performing redux dispatches asynchronously can produce an inconsistent state as the order of arrival of events would not be respected. PR makes sure all the most important state mutating operations happen in the right order.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-58002


webapp/src/actions.ts
```diff
@@ -473,7 +474,22 @@ export const stopRingingForCall = (callID: string): ActionFuncAsync => {
...
+    return async (dispatch: DispatchFunc, getState: GetStateFunc) => {
+        const missingIds = [];
+        for (const id of ids) {
+            if (!getState().entities.users.profiles[id]) {
+                missingIds.push(id);
+            }
+        }
+        if (missingIds.length > 0) {
+            console.log('dispatching profiles');
+            dispatch({type: UserTypes.RECEIVED_PROFILES, data: await RestClient.getProfilesByIds(missingIds)});
```
streamer45, line 487: Any missing profile will go to the existing webapp reducer.

streamer45:
Updated the e2e which passes locally. Will need to merge https://github.com/mattermost/mattermost/pull/26867 first in order to make it pass on this side.

webapp/src/reducers.ts
```diff
@@ -1006,12 +939,6 @@ const clientConnecting = (state = false, action: { type: string, data: boolean }
 export default combineReducers({
     channels,
     clientStateReducer,
-    profiles,
-
-    // DEPRECATED - Needed to keep compatibility with older MM server
-    // version.
-    voiceConnectedProfiles: profiles,
-
```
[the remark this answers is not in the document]
cpoile: 👋

webapp/src/utils.ts
```diff
@@ -293,21 +293,16 @@ export async function getProfilesByIds(state: GlobalState, ids: string[]): Promi
...
+export function getUserIDsForSessions(sessions: SessionState[]) {
+    const idsMap: {[id: string]: boolean} = {};
+    for (const session of sessions) {
+        idsMap[session.user_id] = true;
+    }
+    return Object.keys(idsMap);
+}
+
+export function getSessionsMapFromSessions(sessions: SessionState[]) {
+    return sessions.reduce((a, v) => ({...a, [v.session_id]: v}), {});
```
cpoile, line 305: nit: This is overly wasteful, no need to spread the accumulator, since you're creating as the input you can mutate it.

webapp/src/websocket_handlers.ts
```diff
@@ -182,6 +196,8 @@ export async function handleUserJoined(store: Store, ev: WebSocketMessage<UserJo
         notificationsStopRinging(); // And stop ringing for _any_ incoming call.
     }
 
+    store.dispatch(loadProfilesByIdsIfMissing([userID]));
```
cpoile, line 199: Given the comments we're adding about everything being synchronous, maybe we should add a comment that this will dispatch will be async?

cpoile (approved):
Awesome, thank you. Just some non-blocking questions.

standalone/src/recording/index.tsx
```diff
@@ -23,19 +27,19 @@ import {
...
-async function fetchProfileImages(profiles: UserProfile[]) {
+async function fetchProfileImages(ids: string[]) {
     const profileImages: {[userID: string]: string} = {};
     const promises = [];
-    for (const profile of profiles) {
+    for (const id of ids) {
         promises.push(
             runWithRetry(() => {
-                return fetch(`${getPluginPath()}/bot/users/${profile.id}/image`, RestClient.getOptions({method: 'get'})).then((res) => {
+                return fetch(`${getPluginPath()}/bot/users/${id}/image`, RestClient.getOptions({method: 'get'})).then((res) => {
```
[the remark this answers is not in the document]
streamer45: @cpoile The recorder client won't have any user images so it will need to fetch them for all participants. We could optimize in case of a user re-joining but we are talking small things I think.

streamer45:
Merging this as the only failing tests are the one caused by the webapp regression, getting fixed in https://github.com/mattermost/mattermost-plugin-calls/pull/728

PR #718, opened 2024-04-29
Load-testing tweaks for speech; mute/unmute; benchmarking script

cpoile:
#### Summary
- Participants mute/unmute, with updated/fixed rtcd client dependency
- And a few other tweaks
- Run tests in the lt package
- A better benchmarking script, timed to be 10 minutes, with more realistic levels of cross-talk

#### Ticket Link
- none


cpoile:
Will update transcriber hash when https://github.com/mattermost/calls-transcriber/pull/23 is approved and merged.

lt/client/client.go
```diff
@@ -274,10 +274,20 @@ func (u *User) transmitAudio() {
...
+	if err != nil {
 		log.Printf("%s: failed to mute: %s", u.cfg.Username, err.Error())
 	}
+	return err
+}
+
+func (u *User) Unmute(track webrtc.TrackLocal) error {
+	err := u.callsClient.Unmute(track)
+	if err != nil {
+		log.Printf("%s: failed to mute: %s", u.cfg.Username, err.Error())
```
streamer45, line 288: Should be `failed to unmute`.

lt/client/client.go
```diff
@@ -291,17 +301,23 @@ func (u *User) transmitSpeech() {
...
 	}
 
-	if err := u.callsClient.Unmute(track); err != nil {
-		log.Fatalf(err.Error())
-	}
-
 	for text := range u.speechTextCh {
 		func() {
 			defer func() {
+				time.Sleep(100 * time.Millisecond)
```
streamer45, line 307: Is the sleep necessary?

lt/client/client.go
```diff
@@ -291,17 +301,23 @@ func (u *User) transmitSpeech() {
...
+				if err := u.Mute(); err != nil {
+					log.Fatalf(err.Error())
+				}
+				log.Printf("muted")
```
streamer45, line 311: If we need to log we should include the username. Same for the unmuted case below.

lt/cmd/speech/main.go
```diff
@@ -18,7 +18,7 @@ import (
 )
 
 const (
-	duration = 10 * time.Minute
+	duration = 11 * time.Minute
```
streamer45, line 21: What's the context on this change? :p

lt/client/client.go
cpoile: It's aesthetic -- the mute was firing slightly before the voice had actually finished. It was bugging me. :)

lt/client/client.go
cpoile: Totally, thanks.

lt/cmd/speech/main.go
cpoile: Because we're reusing the connect, it disconnects the scripted user after duration, even if the script isn't done. Since the script is around 10 mins, I'm just bumping it above that. But it makes more sense to turn this into a flag.
I should create a cfg and pass that instead of using vars. But I'll leave that for later if we need it.

streamer45: @cpoile I am not 100% convinced by this. I think it would be fair and likely simpler to ignore any duration if we are working with a script. The length of the script will determine how long it'll run. Thoughts?

cpoile: Sure, sounds good. I'd prefer to turn off the duration for scripts, but I didn't want to add a complicated conditional duration ticker in the connect function. So I opted to just set a long max duration for scripts (24 hours).

streamer45: Makes sense :+1:

streamer45 (approved):
Looks good!

PR #719, opened 2024-04-30
[MM-57263] Implement call settings in popout window

streamer45:
#### Summary

PR implements the call settings menu on the popout window.

I'll create a follow-up ticket to see if we could re-use the new component on the widget although that's more work as they are not matching 100% in either functionality or style.

#### Screenshots

image

#### Design

https://www.figma.com/file/UIfE0vN1ESsXLHZYDP5eqD/Updates-to-Control-Bar-and-Header

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-57263


webapp/src/components/expanded_view/call_settings.tsx
```diff
@@ -0,0 +1,477 @@
...
+const AudioDevicesList = ({deviceType, devices, currentDevice, onDeviceClick}: AudioDevicesListProps) => {
+    const {formatMessage} = useIntl();
+
+    // Note: this is system default, not the concept of default that we save in local storage in client.ts
+    const makeDeviceLabel = (device: MediaDeviceInfo) => {
+        if (device.deviceId.startsWith('default') && !device.label.startsWith('Default')) {
+            return formatMessage({defaultMessage: 'Default - {deviceLabel}'}, {deviceLabel: device.label});
+        }
+        return device.label;
+    };
```
cpoile, line 35: I was going to say: This is one of those places where we should start using useCallback (it can't be put outside the component because we need formatMessage, but we don't need it to be recreated every render).
But then you could say "why not put useCallback around the other functions in this file?" And you could... but there's no real need to -- it's not a hot path at all, and it would probably add even more garbage. So nevermind. Just thinking out loud.

webapp/src/components/expanded_view/call_settings.tsx
```diff
@@ -0,0 +1,477 @@
...
+                className='Menu__content dropdown-menu'
+            >
+                {list}
+            </DevicesList>
+        </div>
+    );
+};
+
+const AudioDeviceSelectedIcon = styled.div`
+&&&& {
```
cpoile, line 75: For this and most of the others, are you sure we need this much specificity? I don't think I've needed to add that in awhile...

For the svg on the next line, a `> svg` should be enough?

webapp/src/components/expanded_view/call_settings.tsx
```diff
@@ -0,0 +1,477 @@
...
+            <AudioDevicesList
+                deviceType={deviceType}
+                devices={devices}
+                currentDevice={isInput ? currentAudioInputDevice : currentAudioOutputDevice}
+                onDeviceClick={handleDeviceClick}
+            />
+            }
+            <li
+                className='MenuItem'
+            >
```
cpoile, line 200: nit: this (and others in the file) can be one line and would look cleaner.  But not a big deal.

webapp/src/components/expanded_view/call_settings.tsx
```diff
@@ -0,0 +1,477 @@
...
+
+const DevicesList = styled.ul`
+&&&& {
+  top: 0;
+  left: calc(100% + 4px);
+  overflow: auto;
+  width: 280px;
+  max-height: calc(100% + 36px);
+  border-radius: 8px;
+}
```
cpoile, line 111: nit: Is there a way to tell your formatter to indent the styled-components css? Also is it possible to set it to 4-spaces? If you really prefer 2, I can set mine to 2 then. I just always thought 4 fits the rest of the file better, but that's the only reason I would prefer it.

cpoile (approved):
This is beautiful, nicely done!

Just non-blocking comments/nits.

It made me realize that the widget's selector might need a fix (the right arrow is missing on the input menu?):
image

webapp/src/components/expanded_view/call_settings.tsx
streamer45: We need the linter. Can't rely on people's IDEs :p

webapp/src/components/expanded_view/call_settings.tsx
streamer45: I really struggled with this because most of them actually needed the extra specificity due to how our menu classes are defined. And that's why in-line styling was working so well, but someone introduced styled components :p

streamer45: I can review, it just takes time.

webapp/src/components/expanded_view/call_settings.tsx
streamer45: Good thoughts. To be honest, I mostly rely on you to tell me if something is highly unoptimized on the React side. But in general, I'd expect us to measure what the actual impact is before making any change and in this instance, it seems minimal considering the whole component is unmounted most of the time.

webapp/src/components/expanded_view/call_settings.tsx
streamer45: Remember https://mattermost.atlassian.net/browse/MM-48266?

abhijit-singh (changes requested):
Looks good @streamer45! A few things I noticed that we should probably fix:
1. When you change the audio/microphone settings from the widget, it gets updated in the call pop out window. But if you change the settings in the pop out window, they don't get updated in the widget, even though the setting actually gets changed. So the widget at that point is essentially showing wrong information. It is an edge case but we should try to make them consistent if possible. 
2. The settings submenu currently gets hidden behind the thread sidebar on smaller screens. Can we bring it to the front instead of behind the sidebar? Just like it works with the participants sidebar right now. 
image

3. I know this is consistent with the widget, but the options in the submenu look disabled right now with that 56% opacity. We can just have them use the center-channel-color variable and remove the opacity in both places (popout window sub menu and widget submenu).
image

streamer45:
@abhijit-singh

1. Yeah the widget side was still bugged, also we weren't enforcing true sync. I made some changes that should keep things in sync at all times, even when components are rendering.
2. This was slightly tricky due to how we render the chat thread but it should work as expected now without hopefully breaking something else. 
3. Easy enough, done.

@cpoile Re-asking for dev review since I touched a bunch of stuff.

abhijit-singh (approved):
Awesome! Thank you @streamer45!

cpoile (approved):

PR #721, opened 2024-05-02
[MM-58085] Improve calls load balancing logic

streamer45:
#### Summary

PR implements a (hopefully) more efficient load-balancing logic for calls. Up until now, we'd be using a simple round-robin approach which can work well for lots of smaller calls but it can be quite inefficient in case of larger calls.

The proposed changes will fetch actual system load (CPU) info from the rtcd instances and select the host with the lower load. 
The rationale here is that we know CPU to be the main performance bottleneck. Exposing this information avoids having to calculate the load in more complex ways such as figuring out how many connections and tracks (and their type) we are sending at any given time.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-58085


server/rtcd.go
```diff
@@ -215,42 +215,70 @@ func (m *rtcdClientManager) getHost(ip string) *rtcdHost {
...
+		} else if info.Load.Load1 < minLoad {
+			minLoad = info.Load.Load1
+			hostWithMinLoad = hostsAvailable[i]
+		}
+	}
+
+	// Fallback to random choice if we couldn't get system info.
+	if hostWithMinLoad == nil {
+		hostWithMinLoad = hostsAvailable[rand.Intn(len(hostsAvailable))]
+	}
```
streamer45, line 273: This and the `continue` in case of error above are to make the change backward compatible in which case we'd be using a randomized approach.

cpoile (approved):
Looks great, nicely done!

streamer45:
@cpoile Asking for re-review since I slightly changed the logic after I noticed that the 1 minute average wasn't as reactive as I would have liked. We are now using a 2-second instant load (see https://github.com/mattermost/rtcd/tree/MM-54335-improvements).

cpoile (approved):
Cool, looks great.

Issue #722, opened 2024-05-03
User avatars are not rendered


Saucistophe:
We just tried the plugin after upgrading Mattermost from version 7 to the latest. We also updated the plugin to its latest version.

The plugin works, but somehow the avatars have an incorrect URL ("/api") and are not rendered:

![image](https://github.com/mattermost/mattermost-plugin-calls/assets/7082839/27bbbf6e-feaf-465e-8d88-b82c7c71519a)

Did we miss something during the upgrade?

streamer45:
@Saucistophe Haven't seen that before. Are you seeing the avatars correctly in other components outside of the plugin?

It would be useful to clear the cache and share the output of the Network tab (see example below).

![image](https://github.com/mattermost/mattermost-plugin-calls/assets/1832946/44402885-0ec3-4a4c-b0de-a5be329d4126)

Also, are you running MM under a subpath by any chance?

Saucistophe:
Yes, they are correctly displayed. It comes from the fact that we serve mattermost under a nested route (XXXX.com/mattermost).
The avatars are correctly displayed in the app, with the url:
```
https://<redacted>.com/mattermost/api/v4/users/5f56da3khjnffxiitqscmreaaw/image?_=0
```
But in the plugin they are KO, with url:
```
https://<redacted>.com/api/v4/users/1of93e5337ngjfhg3exeys4hme/image?_=1707915276925
```
notice the missing `/mattermost`

streamer45:
@Saucistophe Could you confirm the Calls plugin version you are running?

streamer45:
Thanks for the report @Saucistophe . A fix for this should be available in the upcoming release (v0.27)

Saucistophe:
Thanks for the quick response!!

PR #723, opened 2024-05-06
[MM-58125] Fix Client4 initiated requests when running under subpath

streamer45:
#### Summary

PR fixes some other issues with fetching assets when MM is running under a subpath. It's the same fix as https://github.com/mattermost/mattermost-plugin-calls/pull/663 but ported to the webapp side.

I am not sure what happened here since you raised this exact point in https://github.com/mattermost/mattermost-plugin-calls/pull/663#pullrequestreview-1946394499 which I thought I addressed as there was no harm in doing it on both sides but in the end I obviously didn't :man_facepalming: 

Fixes https://github.com/mattermost/mattermost-plugin-calls/issues/722

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-58125


cpoile (approved):
👍

PR #724, opened 2024-05-06
[MM-57466] Host controls: Remove from call

cpoile:
#### Summary
- Host remove, send a host_remove ws event (for the removed user's UX only -- the remove happens on the server)
- Had to do a bit of refactoring of the in call notifications and error modals to match the new designs, this will affect all the notifications and modals. 
- Tidied up how we were doing spacing so it's a bit more uniform and predictable. Please take a look @matthewbirtch in case I missed anything.
- Designs: https://www.figma.com/file/8C0ojMc778Zx3a0nm66Edw/MM-56584-Host-Controls?node-id=1335%3A419040&mode=dev

| | | |
|--|--|--|
| image | image | image |
| image | desktop: image | |

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-57466


matthewbirtch (changes requested):
Thanks @cpoile. Looking and working great. Just a few minor discrepancies:

1. On the 'remove participant' confirmation popup in the expanded window:
    - The icon should be a bit larger (font-size should be 24px)
    - The alignment of the icon is a bit off. It should align vertically with the title, so it probably needs to bump up and to the left to accomplish this.
    - Can you adjust the buttons in this popover to the smaller size?
    - use these classes for the primary button`btn btn-sm btn-primary btn-danger`
    - use these classes for the secondary button`btn btn-sm btn-tertiary` 
    image

2. The menu divider looks too dark to me. I think because it is getting another border-color with 20% opacity applied to it, but it should be center channel color with 8% opacity
    image
   
image

cpoile:
@matthewbirtch Should be fixed now, let me know!

matthewbirtch (approved):
> Should be fixed now, let me know! (screencaps in summary updated)

@cpoile could you bump the icon down in this popover notice? It is sitting too high and looks like it needs to move down by 4px.

image




Other than that, I think we're good to go

standalone/src/init.ts
```diff
@@ -282,6 +282,9 @@ export default async function init(cfg: InitConfig) {
         case `custom_${pluginId}_host_lower_hand`:
             handleHostLowerHand(store, ev as WebSocketMessage);
             break;
+        case `custom_${pluginId}_host_removed`:
+            handleHostRemoved(store, ev as WebSocketMessage);
```
streamer45, line 286: Eventually, I'd expect us to update our [Go client implementation](https://github.com/mattermost/rtcd/tree/master/client) to handle these events accordingly.

webapp/src/actions.ts
```diff
@@ -606,3 +607,17 @@ export const lowerHand = async (callID: string, sessionID: string) => {
...
     );
 };
+
+export const hostRemove = async (callID?: string, sessionID?: string) => {
+    if (!callID || !sessionID) {
+        return {};
+    }
+
+    return RestClient.fetch(
+        `${getPluginPath()}/calls/${callID}/host/remove`,
```
streamer45, line 617: Random thought since we've been battling with issues. Would this work properly aginst a server running under a subpath? I see we are handling that case in `getPluginPath()` but now we are also setting the path to the client itself on initialization. Just want to double check the two things are not conflicting.

webapp/src/websocket_handlers.ts
```diff
@@ -526,3 +535,37 @@ export function handleHostLowerHand(store: Store, ev: WebSocketMessage<{
...
+    if (!client || client?.channelID !== channelID) {
+        return;
+    }
+
+    const sessionID = client.getSessionID();
+    if (ev.data.session_id !== sessionID) {
+        return;
+    }
+
+    handleCallEnd(store, {data: {channelID}} as unknown as WebSocketMessage<EmptyData>); // fake the wsMsg
```
streamer45, line 555: Would calling this result in the client thinking the call has actually ended? I don't think that should be the case as they should be able to see it ongoing from outside (profiles and all).

webapp/src/websocket_handlers.ts
```diff
@@ -526,3 +535,37 @@ export function handleHostLowerHand(store: Store, ev: WebSocketMessage<{
...
+
+    if (isDesktopApp()) {
+        if (window.desktopAPI?.sendCallsError) {
+            window.desktopAPI.sendCallsError('client-error', channelID, hostRemovedMsg);
+        } else {
```
streamer45, line 560: I don't love having this in more than one place. Can we think of something else?

streamer45 (commented):
Nice work, left some initial comments.

server/host_controls_api.go
```diff
@@ -114,3 +114,30 @@ func (p *Plugin) handleLowerHand(w http.ResponseWriter, r *http.Request) {
...
+	}
+	if err := json.NewDecoder(http.MaxBytesReader(w, r.Body, requestBodyMaxSizeBytes)).Decode(&payload); err != nil {
+		res.Err = err.Error()
+		res.Code = http.StatusBadRequest
+		return
+	}
+
+	if err := p.hostRemoveSession(userID, callID, payload.SessionID); err != nil {
+		p.LogError("handleRemoveSession: failed to remove", "err", err.Error())
+		res.Code = http.StatusInternalServerError
```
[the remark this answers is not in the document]
cpoile: Ok, I made some err codes for host controls. I suppose I could have done a single type for server error, and everything else would be a bad request, but this feels better (if a bit verbose).

standalone/src/init.ts
cpoile: sure: https://mattermost.atlassian.net/browse/MM-58171

webapp/src/actions.ts
cpoile: Yes it works, just tested.

webapp/src/websocket_handlers.ts
cpoile: What I really wanted was just the disconnect, so I'm keeping that. You're right that we don't want to lose the profiles. 
I noticed that when the session times out the client is technically disconnected, but it takes another 10 or so seconds for the user's own client to finally disconnect, and they are given a "call connection failed" error message. This is much cleaner.

webapp/src/websocket_handlers.ts
cpoile: Good point. Added a util function, next to the other desktop util function.

cpoile:
@matthewbirtch I'm not sure why you're seeing it like that, the one in the screenshots is what I'm seeing... Hmm

matthewbirtch:
> @matthewbirtch I'm not sure why you're seeing it like that, the one in the screenshots is what I'm seeing... Hmm

My apologies. It must have been a cache issue. I just tried again after clearing my cache and it's all good now.

matthewbirtch (approved):
LGTM!

streamer45 (approved):
Looks great, thanks for addressing everything :+1:

cpoile:
@streamer45 Let me know if https://github.com/mattermost/mattermost-plugin-calls/pull/724/commits/352b3c0c6a7a070ffcc58837e51c334d52cb5654 looks good

streamer45:
Thanks @cpoile. Eventually we should review it and optimize it (we only need to know if the session is still there) but I think we have a ticket for that already so all good!

PR #725, opened 2024-05-06
[MM-58033] Fix: Incoming call banner in expanded view not working

cpoile:
#### Summary
- Removed the check for expanded view (not sure why we even had that -- the channel switch does work in safari.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-58033

### DIFF
--- a/e2e/tests/notifications.spec.ts
+++ b/e2e/tests/notifications.spec.ts
@@ -1,4 +1,3 @@
-
 import {expect, test} from '@playwright/test';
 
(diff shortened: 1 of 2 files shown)


streamer45:
The context would be https://github.com/mattermost/mattermost-plugin-calls/pull/473#issuecomment-1635203698. No idea why we disabled clicking altogether though, surely an oversight.

streamer45 (approved):
Nice! Thanks for the test :)

PR #726, opened 2024-05-07
[MM-57470] Host controls: Mute all

cpoile:
#### Summary
- Added host mute all command
- @matthewbirtch It's a bit tough to tell, so maybe I'm wrong, but it looked like there were some small differences in the participants list on widget and the designs, so I made some small spacing changes to the header but didn't touch anything else. I assumed we wanted some hover on the mute all command, so I added that: on the expanded view the height matches the close button in the RHS (and it's blue). On the widget, the mute-all can't be as far right or there won't be enough room for the hover, so I moved it a bit to the right. Have a look and let me know if you'd like anything changed.

| | | | |
|--|--|--|--|
| image | image | image | image |

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-57470


matthewbirtch (changes requested):
Looks good @cpoile. Just one minor detail:

The mute all text seems to be sitting a bit low within the button. It seems to be more misaligned in the expanded view participant RHS context.
Screenshot 2024-05-08 at 11 04 31 AM

Screenshot 2024-05-08 at 11 08 55 AM

cpoile:
@matthewbirtch I'm not sure why you're seeing that either... Here's what I'm seeing:
image
image

matthewbirtch:
> @matthewbirtch I'm not sure why you're seeing that either... Here's what I'm seeing: image image

@cpoile looks like this was a cache issue as well. All good for me after a hard refresh.

The only other thing is what I mentioned over the equivalent mobile PR - `Mute all` should not mute the host who has clicked mute all: https://github.com/mattermost/mattermost-mobile/pull/7945#pullrequestreview-2046539326

cpoile:
@matthewbirtch Let's move that conversation here, as this is the PR that would include that functionality.
Personally, I think clicking "Mute all" should mute all, otherwise it doesn't do what it says it will do. If the host wants to be unmuted after, they can click their own mute button.
On the other hand, I can understand the usefulness of muting everyone else in order to be the only one speaking. But it seems like a rare use case -- that it is the host who is currently talking and wants to "mute others" (since it's only the host who will have this button).
But if it is still desirable, then I think we would need to change the label to something like `Mute others`.
@streamer45 or @esethna thoughts?

matthewbirtch:
> @matthewbirtch Let's move that conversation here, as this is the PR that would include that functionality. Personally, I think clicking "Mute all" should mute all, otherwise it doesn't do what it says it will do. If the host wants to be unmuted after, they can click their own mute button. On the other hand, I can understand the usefulness of muting everyone else in order to be the only one speaking. But it seems like a rare use case -- that it is the host who is currently talking and wants to "mute others" (since it's only the host who will have this button). But if it is still desirable, then I think we would need to change the label to something like `Mute others`. @streamer45 or @esethna thoughts?

I'm open to a label change. As another data point, this seems to be the common behavior in both Teams and Zoom. 'Mute all' mutes everyone except you. (screenshot from teams confirmation modal)
image

streamer45:
> @matthewbirtch Let's move that conversation here, as this is the PR that would include that functionality. Personally, I think clicking "Mute all" should mute all, otherwise it doesn't do what it says it will do. If the host wants to be unmuted after, they can click their own mute button. On the other hand, I can understand the usefulness of muting everyone else in order to be the only one speaking. But it seems like a rare use case -- that it is the host who is currently talking and wants to "mute others" (since it's only the host who will have this button). But if it is still desirable, then I think we would need to change the label to something like `Mute others`. @streamer45 or @esethna thoughts?

Not opinionated but what's the purpose of muting everyone if then you have to unmute right away? Maybe we are thinking of different use cases (e.g. before a call starts vs during a call).

cpoile:
@streamer45 So you'd expect `Mute all` to mute everyone except yourself?  That's different from what I would expect, but if it's what everyone wants I'll make the change, no problem.

streamer45:
> @streamer45 So you'd expect `Mute all` to mute everyone except yourself? That's different from what I would expect, but if it's what everyone wants I'll make the change, no problem.

It's not really about what I expect but trying to understand the usefulness of muting everyone and keeping a call silent.

cpoile:
@streamer45 I can see that too. 
I'm fine with whatever you both think is best. 👍

streamer45:
> @streamer45 I can see that too. I'm fine with whatever you both think is best. 👍

Let's make sure the messaging explains exactly what's going to happen and then I think we can keep it as designed.

cpoile:
@streamer45 So... that's a vote for making it mute everyone else, but keep it as `Mute all`?

streamer45:
> @streamer45 So... that's a vote for making it mute everyone else, but keep it as `Mute all`?

Do we have the extra messaging in a prompt like Matt showed above?

cpoile:
@streamer45 No, it's a one-click as designed.

streamer45:
> @streamer45 No, it's a one-click as designed.

Okay, if we don't add more messaging then I think it would be best to change the label to `Mute others` as proposed above.

cpoile:
Ok, @matthewbirtch sound good?

matthewbirtch:
> Ok, @matthewbirtch sound good?

Yep, let's proceed with that. Alternatively we keep it `Mute all` with a tooltip `This will mute everyone in the call but you`

cpoile:
Done:
image
image

Only shows when someone other than host is unmuted, and only affects sessions other than the host's.

server/host_controls.go
```diff
@@ -97,6 +97,36 @@ func (p *Plugin) muteSession(requesterID, channelID, sessionID string) error {
...
+	// Unmute anyone muted (who is not the host/requester).
+	// If there are no unmuted sessions, return without doing anything.
+	for id, s := range state.sessions {
+		if s.Unmuted && s.UserID != requesterID {
+			p.publishWebSocketEvent(wsEventHostMute, map[string]interface{}{
+				"channel_id": channelID,
+				"session_id": id,
+			}, &model.WebsocketBroadcast{UserId: s.UserID, ReliableClusterSend: true})
+		}
+	}
```
streamer45, line 125: Could we add a TODO so I remember to update this when we finally merge https://github.com/mattermost/mattermost-plugin-calls/pull/609?

streamer45 (approved):
Looks good!

webapp/src/components/call_widget/participants_list.tsx
```diff
@@ -86,11 +99,37 @@ const styles: Record<string, React.CSSProperties> = ({
...
+    line-height: 16px;
+    color: var(--button-bg);
+
+    border: none;
+    background: none;
+    border-radius: 4px;
+
+    &:hover {
+        // thanks style sheets...
+        background: rgba(var(--button-bg-rgb), 0.08) !important;
```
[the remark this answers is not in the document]
cpoile: @streamer45 This wouldn't work no matter how much class specificity I added, presumably because the one it was trying to trump was an id selector. :/

server/host_controls.go
cpoile: @streamer45 Hm, I'm not sure how that would that affect this?

streamer45: It will because we are extending `publishWebSocketEvent` to support a slice of user ids to loop through which is essentially what we are doing here.

matthewbirtch (approved):

server/host_controls.go
cpoile: Ahhhhh, yes. Thanks.

xxfogs:
Is this a paid feature? I seem to be on v0.27, yet I don't have such a button

image

image

streamer45:
@xxfogs Yes, host controls are included in our professional tier, meaning a license is required.

PR #727, opened 2024-05-08
Bump rtcd to latest v0.15.0

streamer45:
#### Summary

Bumping `rtcd` to the latest release.

### DIFF
--- a/server/activate.go
+++ b/server/activate.go
@@ -182,7 +182,7 @@ func (p *Plugin) OnActivate() error {
 		rtcServerConfig.TURNConfig.StaticAuthSecret = cfg.TURNStaticAuthSecret
 	}
 	if cfg.ICEHostPortOverride != nil {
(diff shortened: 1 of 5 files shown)


cpoile (approved):

server/activate.go
```diff
@@ -182,7 +182,7 @@ func (p *Plugin) OnActivate() error {
 		rtcServerConfig.TURNConfig.StaticAuthSecret = cfg.TURNStaticAuthSecret
 	}
 	if cfg.ICEHostPortOverride != nil {
-		rtcServerConfig.ICEHostPortOverride = *cfg.ICEHostPortOverride
+		rtcServerConfig.ICEHostPortOverride = rtc.ICEHostPortOverride(fmt.Sprintf("%d", *cfg.ICEHostPortOverride))
```
streamer45, line 185: I think it's reasonable to keep supporting the basic syntax (single port) only from this side. We can always review later if needed but for anything advanced I'd say they should be using `rtcd`.

PR #728, opened 2024-05-08
Fix style after global changes

streamer45:
#### Summary

PR attempts to fix some issues that are affecting how Calls is rendering certain components after the global changes in https://github.com/mattermost/mattermost/pull/26407 went in.

So far I am addressing the issues found in the pop-out view but there could be more hiding:

**Before**
before
**After**
after

@matthewbirtch There's one additional noticeable change. The widget is now not centered given the sizing changes on the LHS.

image

image


matthewbirtch:
@streamer45, the popout window looks good. 

For the widget width, we could adjust to fit the new LHS default width?
- when team sidebar is hidden: 248px
- when team sidebar is showing: 306px

One other thing I just noticed is the header height of the participants RHS and the Thread RHS are not matching with the new RHS header height. Changing the height from 63px to 56px should do the trick

image

I also see a very defined line between the participants RHS and Thread RHS when both are open now. I'm not sure what's caused this.

streamer45:
Thanks, let me review the above.

streamer45:
> when team sidebar is hidden: 248px
when team sidebar is showing: 306px

So we are expanding by 26px (280 to 306) if the team sidebar is showing and by 32px (216 to 248) if not. @matthewbirtch Could you confirm whether the inconsistency is meant? Because it will affect other things as now we are assuming the relative gap to be equal.

streamer45:
Thanks. I think we can make it work either way, just wanted to make extra sure it wasn't a mistake.

streamer45:
@matthewbirtch The above should be addressed, please have a look.

matthewbirtch (approved):
Thanks @streamer45. Looks good

cpoile (approved):
Thanks for this Claudio!

streamer45:
@cpoile I pushed a fix for the sizing issues, could you double-check check it's not broken any longer on your side?

cpoile:
@streamer45 Yes, fixed, thank you!

PR #730, opened 2024-05-09
MM-58165 - Updating calls not allowed modal

asaadmahmood:
#### Summary
MM-58165 - Updating calls not allowed modal

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-58165

#### Screenshot
CleanShot 2024-05-09 at 14 52 25@2x


streamer45:
Just a thought but we may want to do a deeper review and make sure all the Calls modals are rendering as expected since we have a few.

streamer45:
> @streamer45 Can we assign a QA on this?

Sure, we can have Dylan some checks.

DHaussermann:
@streamer45 I'm happy to put this on a server and test the modals. 

But, something odd is happening when I try building this locally so, obviously I'd rather go grab the build from the pipeline. I'm a bit rusty on how we I can do this now. Maybe things have changed.

image
I see a package that is `.zip` in the artifacts here ☝️  but it does not contain a `.gz`. Even though the structure looks correct, packaging the contents myself as a `.gz` did not seem to work either :)

Can you please point me to where I can grab the binairy from the pipeline?

streamer45:
@DHaussermann To be honest I never tried using a build produced by our pipeline but let me know if this works -> https://github.com/mattermost/mattermost-plugin-calls/actions/runs/9015566919

DHaussermann:
Thank @streamer45 that **is** the tar bundle I'm looking for 👍

DHaussermann (approved):
Tested and passed 
- Confirmed I see the desired change. The modal now take less vertical space. The whitespace and margins for the text have also been adjusted.
- Checked other modals for visual consistency (Can't start a call, Can't start a recording, Switch to a different call)
- Tested with current server version as well as oldest supported `v9.5.4` to ensure no webapp changes affect this

LGTM!

asaadmahmood:
@DHaussermann Any idea why this e2e test is failing?
CleanShot 2024-05-10 at 18 50 52@2x

PR #731, opened 2024-05-10
[MM-57953] Host controls: Add more host notices

cpoile:
#### Summary
- To avoid confusion, changed host notifications to host notices
- Added notices for host changed and host removed
- Designs: https://www.figma.com/file/8C0ojMc778Zx3a0nm66Edw/MM-56584-Host-Controls?node-id=1335%3A418726&mode=dev

| | |
|--|--|
| image | image |

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-57953


webapp/src/components/host_notices.tsx
```diff
@@ -0,0 +1,154 @@
...
+    border-radius: 16px;
+    width: fit-content;
+    font-weight: 600;
+    font-size: 14px;
+    line-height: 20px;
+    background: #FFFFFF;
+
+    ${({$onWidget}) => $onWidget && css`
+        width: 100%;
+        border-radius: 6px;
```
matthewbirtch, line 120: can we adjust to 8px radius to match the main widget radius?

matthewbirtch (changes requested):
Looks great to me @cpoile. 

1. Just a minor tweak on the corner radius (8px instead of 6px) for the notices that appear above the widget.

2. The host icon seems scaled down in comparison to the others.
    image

4. As a follow-up I think we'll probably want to adjust other notices to match style too. This can be done as a separate PR though if you'd rather

    image

    Screenshot 2024-05-10 at 4 44 01 PM

    I've added these in [Figma here](https://www.figma.com/file/8C0ojMc778Zx3a0nm66Edw/MM-56584-Host-Controls?type=design&node-id=1903-145807&mode=design&t=mcLgnA6of3K7aAyf-4), but let me know if there are others beyond these

matthewbirtch (approved):
Looks good. Thanks for the updates @cpoile. I've created a follow-up ticket for the other notices: https://mattermost.atlassian.net/browse/MM-58221

streamer45:
@cpoile Two things:

1. If a host has been assigned through the slash command in case they drop and rejoin they'd be host again but I don't see the notice showing. Is that expected?
2. If the new host is the current user shouldn't we show "you are the host" to be consistent?

webapp/src/websocket_handlers.ts
```diff
@@ -388,8 +388,8 @@ export function handleCallHostChanged(store: Store, ev: WebSocketMessage<CallHos
         },
     });
 
-    const profiles = profilesInCurrentCallMap(store.getState());
-    const hostProfile = profiles[ev.data.hostID];
+    const hostProfile = profilesInCurrentCallMap(store.getState())[ev.data.hostID] ||
```
cpoile, line 391: This happens because we send out the host change before the join msg. I did this here bc it's less complicated than delaying the host change message server side.

streamer45 (approved):
Thanks @cpoile , looks great!

cpoile:
Merging for now since we need to get master onto community -- if we need to bold the `You` then I can do that in one of the UI PRs.

matthewbirtch:
> Merging for now since we need to get master onto community -- if we need to bold the `You` then I can do that in one of the UI PRs.

Apologies for the delay. I agree with bolding the `You` in this notice @cpoile

PR #733, opened 2024-05-10
Fix live captions icon

streamer45:
#### Summary

Since we are keeping https://github.com/mattermost/mattermost-plugin-calls/pull/719 out of the upcoming release, we need to update the CC icon which is still using the previous sizing.


webapp/src/components/icons/cc_icon.tsx
```diff
@@ -12,22 +12,17 @@ type Props = {
 export default function CCIcon({fill, style, className, ...rest}: Props) {
     return (
         <svg
-            width='24'
-            height='24'
-            viewBox='0 0 24 24'
+            width='20'
+            height='20'
+            viewBox='2 3 20 18'
```
cpoile, line 17: Are you getting this from the svg itself (i.e. did it change)? Because I cut and pasted the svg originally, and afaik you don't need to change this in the svg, you just do what you did in the style above.

cpoile (approved):
Cool, thanks!

streamer45:
@cpoile test

PR #735, opened 2024-05-13
[MM-57919] Use new host controls ws msgs

cpoile:
#### Summary
- Update calls-common, refactor to use new host control ws msg definitions

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-57919

### DIFF
--- a/standalone/src/init.ts
+++ b/standalone/src/init.ts
@@ -14,6 +14,9 @@ import {
     CallStateData,
     EmptyData,
     HelloData,
+    HostControlLowerHand,
+    HostControlMsg,
+    HostControlRemoved,
     UserDismissedNotification,
     UserJoinedData,
     UserLeftData,
@@ -55,7 +58,8 @@ import {
     handleCallStart,
     handleCallState,
     handleHostLowerHand,
-    handleHostMute, handleHostRemoved,
+    handleHostMute,
+    handleHostRemoved,
     handleHostScreenOff,
     handleUserDismissedNotification,
     handleUserJoined,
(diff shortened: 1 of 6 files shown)


standalone/src/init.ts
```diff
@@ -274,17 +277,17 @@ export default async function init(cfg: InitConfig) {
...
+            handleHostScreenOff(store, ev as WebSocketMessage<HostControlMsg>);
             break;
         case `custom_${pluginId}_host_lower_hand`:
-            handleHostLowerHand(store, ev as WebSocketMessage);
+            // Note: I'm not sure what's making tsc unhappy about the following without the `as unknown`
+            handleHostLowerHand(store, ev as unknown as WebSocketMessage<HostControlLowerHand>);
             break;
         case `custom_${pluginId}_host_removed`:
-            handleHostRemoved(store, ev as WebSocketMessage);
+            handleHostRemoved(store, ev as unknown as WebSocketMessage<HostControlRemoved>);
```
cpoile, line 290: This was frustrating. I spent a lot of time trying to understand what was going on, and I can't explain it. Changing calls-common doesn't work. Moving the case statement around doesn't work. The typescript compiler is thinking that `ev` includes `LiveCaptionData` in it, or sometimes it thinks it includes `JobStopData`. And since `HostControlLowerHand` or `HostControlRemoved` doesn't overlap with those two, it calls it an error. So, I don't know.

streamer45: @cpoile The event object is of type `WebsocketEventData` which means any "sub-type" should appear in https://github.com/mattermost/calls-common/blob/7e8e818a7c1cd5e4b7a121170c12d208c834b8d9/src/types/types.ts#L128-L147 since we are essentially creation a union.

cpoile: @streamer45 Oh god. So stupid.

streamer45: No worries, it's a bit convoluted. Wish there was a better way, to be honest.

streamer45 (approved):

PR #736, opened 2024-05-14
Fix pop out height

streamer45:
#### Summary

A small UX regression due to the sizing changes in https://github.com/mattermost/mattermost-plugin-calls/pull/715 which we weren't accounting for in case of screen sharing.

#### Screenshots

*Before*

2024-05-14-162608_2560x1080_scrot

*After*

2024-05-14-162456_2560x1080_scrot


cpoile (approved):
Thanks!

PR #737, opened 2024-05-14
[MM-58232] Host notices: fix background and color

cpoile:
#### Summary
- Fix bg and color on dark themes

| old | new |
|--|--|
| image | image |
| image | image |

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-58232

### DIFF
--- a/webapp/src/components/host_notices.tsx
+++ b/webapp/src/components/host_notices.tsx
@@ -119,7 +119,7 @@ const Notice = styled.div<{ $onWidget?: boolean }>`
     font-weight: 600;
     font-size: 14px;
     line-height: 20px;
-    background: #FFFFFF;
+    background: var(--button-color);
 
     ${({$onWidget}) => $onWidget && css`
         width: 100%;
@@ -128,6 +128,7 @@ const Notice = styled.div<{ $onWidget?: boolean }>`
         font-size: 11px;
         font-weight: 400;
         white-space: pre;
+        background: var(--center-channel-bg);
     `}
 `;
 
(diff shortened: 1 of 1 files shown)


webapp/src/components/host_notices.tsx
```diff
@@ -144,17 +145,24 @@ const RedStyledCompassIcon = styled(StyledCompassIcon)`
...
-    margin-left: ${({$onWidget}) => ($onWidget ? 1 : 0)}px;
+    margin-left: 0;
     margin-top: 1px;
-    fill: var(--center-channel-color-64);
-    width: ${({$onWidget}) => ($onWidget ? 14 : 18)}px;
+    fill: rgba(var(--calls-bg-rgb), 0.56);
+    width: 18px;
+
+    ${({$onWidget}) => $onWidget && css`
+        fill: var(--center-channel-color-64);
```
streamer45, line 154: This would be the first time not using the `rgba()` form. Just raising the inconsistency. I admit it looks cleaner though.

streamer45 (approved):
Thanks, looks great!

webapp/src/components/host_notices.tsx
cpoile: Huh, I never even thought about it -- we've been using both in mobile for a long time. If you want I can take it out later in another PR. But personally I don't mind, because it's clear.

cpoile: as an aside -- don't know why I made a ticket for this. I won't include it in my weekly roundup. :p

PR #739, opened 2024-05-15
chore(cypress): Fix user profile e2e test

yasserfaraazkhan:
Fix: https://github.com/mattermost/mattermost/pull/26420#issuecomment-2113202367

### DIFF
--- a/e2e/tests/join_call.spec.ts
+++ b/e2e/tests/join_call.spec.ts
@@ -89,8 +89,8 @@ test.describe('join call', () => {
         await userBPage.page.locator('[data-testid=SendMessageButton]').click();
 
         await userAPage.locator('.post__header').locator('button.user-popover').last().click();
-        await expect(userAPage.locator('#user-profile-popover')).toBeVisible();
-        await userAPage.locator('#user-profile-popover').locator('#startCallButton').click();
+        await expect(userAPage.locator('div.user-profile-popover')).toBeVisible();
+        await userAPage.locator('div.user-profile-popover').locator('#startCallButton').click();
 
         await expect(userAPage.locator('#calls-widget')).toBeVisible();
         await expect(userAPage.locator('#calls-widget-loading-overlay')).toBeHidden();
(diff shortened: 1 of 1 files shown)


e2e/tests/join_call.spec.ts
```diff
@@ -109,8 +112,9 @@ test.describe('join call', () => {
         await userBPage.page.locator('[data-testid=SendMessageButton]').click();
 
         await userAPage.locator('.post__header').locator('button.user-popover').last().click();
-        await expect(userAPage.locator('#user-profile-popover')).toBeVisible();
-        await expect(userAPage.locator('#user-profile-popover').locator('#startCallButton')).toBeDisabled();
+        await expect(userAPage.locator('div.user-profile-popover')).toBeVisible();
+        await expect(userAPage.locator('div.user-profile-popover').locator('#startCallButton')).toBeDisabled();
+        await userAPage.locator('button.closeButtonRelativePosition').click();
```
cpoile, line 117: Just a thought -- will this be robust if another button is added on the page? Could we limit the locator to the `div.user-profile-popover` to make it more future proof?

cpoile (approved):
Thanks @yasserfaraazkhan!

streamer45:
@yasserfaraazkhan Given these are now passing even though the functionality is still not completely back (thanks for https://mattermost.atlassian.net/browse/MM-58271), would it make sense to update the test to cover what's missing? Let me know if you need help with that.

e2e/tests/join_call.spec.ts
yasserfaraazkhan: @cpoile with the new changes for pop over.
The testing tool encounters some kinda overlay when attempting to interacting with other elements while the profile pop over is open. This `button.closeButtonRelativePosition` is just a step to close the pop over and continue with other steps.

Another way was to `foce` click on elements, but I avoided that. This is just with the test tool. In real time, this works fine.

yasserfaraazkhan:
Sure  @streamer45 
I'll make a note in Jira to add a step based on the fix pushed.

Currently, we do have a step that checks if the button is disabled and it is passing.
https://github.com/mattermost/mattermost-plugin-calls/blob/main/e2e/tests/join_call.spec.ts#L129

please let me know, if you'd suggest to merge or put a pin on this until fix is ready.

streamer45:
Thanks. I think we can merge this one to fix the pipeline but I'd like to address that eventually because it makes us vulnerable to future regressions.

streamer45 (approved):

yasserfaraazkhan:
Got it.
Will keep close eye on these tests.

PR #740, opened 2024-05-15
[MM-57959] Host controls e2e tests

cpoile:
#### Summary
- Added tests for each host control in each situation it can be used in (widget, expanded view participant card, expanded view rhs)
- Added snapshots for each to prevent visual regressions ~~(esp useful for the host notices)~~

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-57959


webapp/src/websocket_handlers.ts
```diff
@@ -590,45 +587,38 @@ export function handleHostRemoved(store: Store, ev: WebSocketMessage<HostControl
...
-            store.dispatch({
-                type: HOST_CONTROL_NOTICE_TIMEOUT_EVENT,
-                data: {
-                    callID: ev.data.call_id,
-                    noticeID: hostNotice.noticeID,
-                },
-            });
-        }, HOST_CONTROL_NOTICE_TIMEOUT);
+    if (ev.data.session_id === sessionID) {
+        getCallsClient()?.disconnect(new Error(hostRemovedMsg));
```
cpoile, line 591: 🙉

streamer45: what happened here :stuck_out_tongue:

cpoile: This is the correct way to do it. Should have done it earlier. :)

cpoile:
Notices seem to be tough to screencap -- they're animated and the usernames are not stable. (each would disqualify them, together = impossible, afaict).

e2e/page.ts
```diff
@@ -48,6 +64,19 @@ export default class PlaywrightDevPage {
...
+        await this.sendSlashCommand('/call end');
+        let modal = this.page.locator('#calls-end-call-modal');
+        if (await modal.isVisible()) {
+            await modal.getByRole('button', {name: 'End call'}).click();
+        } else {
+            modal = this.page.locator('.modal-content');
+            if (await modal.isVisible()) {
+                await modal.getByRole('button', {name: 'Understood'}).click();
+            }
+        }
```
streamer45, line 77: Why is the check necessary? Generally I'd expect it to fail if we cannot end the call because our assumption of the state is likely not correct.

e2e/tests/host_controls.spec.ts
```diff
@@ -1,23 +1,35 @@
...
+    joinCall,
+    joinCallAndPopout,
+    startCall,
+    startCallAndPopout,
+} from '../utils';
 
 const userStorages = getUserStoragesForTest();
 const usernames = getUsernamesForTest();
 
+test.setTimeout(400000);
```
streamer45, line 16: This is a lot, can't be avoided I suppose?

streamer45 (approved):
Amazing test work and great refactoring! :bowing_man:

e2e/page.ts
cpoile: Because we're running it in `test.afterEach` we have to handle either case. If we cannot end the call (some weird state or the server is broken) then that's fine. This is just the two happy path cases. 
The purpose is to clean up after tests that fail in the middle and then leave calls running, which screws up next tests and cause more failures. I want the only failure to be the test that failed for real reasons.

e2e/tests/host_controls.spec.ts
cpoile: I noticed that the run time can vary really wildly. So most of the time it's quick, but sometimes it would take 5m for a run. Maybe it's a server load thing? I'd rather it not fail because of load.

e2e/page.ts
streamer45: That's fine. We can always create a separate method to make it stricter.

cpoile: I'm not sure what you mean -- what kind of separate method?

PR #741, opened 2024-05-16
[MM-58195] [MM-58283] Fix banner

cpoile:
#### Summary
- ~~Just a little fix needed, otherwise it's centred.~~ New designs

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-58195
- https://mattermost.atlassian.net/browse/MM-58283


streamer45 (approved):
Thanks, the specific fix is fine but these banners will likely need more polish because they just look off. The icon and font size seem a bit inconsistent and the close button doesn't have the proper hover and active states, plus the icon color is very faded.

matthewbirtch:
@cpoile `canceled` looks to be spelled incorrectly in the banner (American English)

cpoile:
Updated based on new designs: https://www.figma.com/design/TyjWV32kguu5YgPQJjpwVy/Call-Banners?node-id=0-1&m=dev

image
image
image
image
image
image
image
image

webapp/src/constants.ts
```diff
@@ -52,7 +52,7 @@ export const CallAlertConfigs: { [key: string]: CallAlertConfig } = {
         type: CallAlertType.Warning,
         icon: 'alert-outline',
         bannerText: defineMessage({defaultMessage: 'Call quality may be degraded due to unstable network conditions.'}),
-        dismissable: false,
+        dismissable: true,
```
streamer45, line 55: I'd like to understand the effect of changing this because it was made not-dismissable by design given it's a condition affecting the client that could cause a degraded experience. Would the banner reappear at the next `mos` event that goes over the threshold or would it be permanently hidden?

webapp/src/components/expanded_view/global_banner.tsx
```diff
@@ -29,55 +38,74 @@ export default function GlobalBanner(props: Props) {
...
+    width: 100%;
+    height: 40px;
+    background-color: ${({$bgColor}) => $bgColor};
+    font-size: 14px;
+    gap: 4px;
     color: ${({$color}) => $color};
-  }
 
-  i {
-    font-size: 18px;
```
matthewbirtch, line 62: This should stay - the icon font-size should still be 18px

cpoile: @matthewbirtch It is still there, just moved to where it should be.

webapp/src/constants.ts
cpoile: It will go away and then come back on the next `mos` event. I think we should think about the logic here and decide if we even want this banner to be dismissable.

webapp/src/components/expanded_view/global_banner.tsx
matthewbirtch: Hmmm, it doesn't seem like it's getting applied then. I see you have added it on the close button, but this icon font-size should apply in the banner area as well

webapp/src/constants.ts
streamer45: Either we implement a backoff or we keep it on.

cpoile: I would prefer not to add the complexity of a lock and a timer for this...

webapp/src/components/expanded_view/global_banner.tsx
cpoile: Yep you're right, my mistake.

matthewbirtch (approved):
Looks good @cpoile. Thanks for these updates.

The only other thing I'm thinking is that for the warning banner, we maybe shouldn't use a theme color for the background - we should use a hard coded value of `#FFBC1F` (or the rgb equivalent - `rgb(255, 188, 31)`). Just in case someone has a weird color for their away indicator. I was just thinking since we're hard-coding the text color on the warning banner, we should do the same on the background

cpoile:
@matthewbirtch We're using `--away-indicator` and `--dnd-indicator` everywhere in the code already. If we're concerned about remapping of these, then maybe we should change them all in a separate effort?

matthewbirtch:
> @matthewbirtch We're using `--away-indicator` and `--dnd-indicator` everywhere in the code already. If we're concerned about remapping of these, then maybe we should change them all in a separate effort? Edit: but I'll make the change here, and we can decide whether to change them elsewhere later.

ok, we would need to be consistent with the widget call degraded banner

cpoile:
@matthewbirtch by being consistent, do you mean doing what we have for the widget banner now? (using `var(--away-indicator)` bg and `var(--center-channel-color)` fg, or changing widget to the hardcoded values here?

matthewbirtch:
> @matthewbirtch by being consistent, do you mean doing what we have for the widget banner now? (using `var(--away-indicator)` bg and `var(--center-channel-color)` fg, or changing widget to the hardcoded values here?

If we use `var(--center-channel-color)` for the foreground it will result in white text on yellow background in dark themes.
image

So I think we need to go with hard-coded values in that context as well

webapp/src/constants.ts
streamer45: I am on board :)

cpoile: Ok, I didn't want to use a lock and timer, I used a throttle. Here's the behavior: the call banner will appear instantly when it's first called, it will turn on and off based on mos just like always. But if you dismiss the banner, it won't come back at least for 20 seconds, then it will come back if there is still network trouble.  We can lengthen the delay if we want.

cpoile:
@matthewbirtch Makes sense, here are screencaps with a dark theme:
image
image
image
image
image
image
image
image
image
image
image

webapp/src/constants.ts
cpoile: And before you ask, I don't think we need to implement cross-popout/widget dismiss synchronization. 😛

streamer45: I was fine with just making it not dismissible so not going to ask for that :p

webapp/src/constants.ts
```diff
@@ -12,6 +12,7 @@ export const CALL_RECORDING_POST_TYPE = 'custom_calls_recording';
 export const CALL_TRANSCRIPTION_POST_TYPE = 'custom_calls_transcription';
 export const LIVE_CAPTION_TIMEOUT = 5000;
 export const HOST_CONTROL_NOTICE_TIMEOUT = 5000;
+export const DEGRADED_CALL_QUALITY_ALERT_WAIT = 20000;
```
streamer45, line 15: I think we should make this dynamic based on the derivative of the mos change. That way if the connection goes really bad we can show it again without waiting.

&nbsp;
&nbsp;
&nbsp;
&nbsp;
 
I am just kidding :)

cpoile: 😭

streamer45 (approved):
Looks great! Thanks for fixing it all :)

matthewbirtch (approved):
Looks great @cpoile. Thanks for these improvements.

I'm going to approve, but, could we make a couple minor last tweaks on the widget banners? 
image
1. the x icon should be right-aligned
2. can we make the hover state of the x icon have `border-radius: 2px`?

cpoile:
@matthewbirtch Sure thing, here's the revision:
image

@streamer45 Can you have another look at the call quality banner code? I  found an edge case -- you dismiss, call quality gets better, then gets worse, then gets better within the delay time. The banner will appear briefly until the next good mos is received. Not a big deal, but I don't want that flicker. It could be fixed with a tweak to the throttle, but the throttle solution is over complicated and not needed. A lock is simpler. 😅

matthewbirtch:
> @matthewbirtch Sure thing, here's the revision: 
image

Looks great, thanks!

e2e/tests/start_call.spec.ts
```diff
@@ -451,7 +451,7 @@ test.describe('call post', () => {
 
         await page.getByTestId('PostDotMenu-Button-' + postID).click();
 
-        await page.locator('#CENTER_dropdown_' + postID).locator('li', {hasText: 'Edit'}).click();
+        await page.keyboard.type('e');
```
streamer45, line 454: What is this change about?

webapp/src/components/call_widget/component.tsx
```diff
@@ -1502,6 +1508,12 @@ export default class CallWidget extends React.PureComponent<Props, State> {
                             },
                         },
                     });
+                    if (alertID === 'degradedCallQuality') {
+                        this.callQualityBannerLocked = true;
+                        setTimeout(() => {
+                            this.callQualityBannerLocked = false;
+                        }, DEGRADED_CALL_QUALITY_ALERT_WAIT);
```
streamer45, line 1515: I'd expect this to be in the state to guarantee re-renders. I'm wondering if this is any different, though.

streamer45 (approved):
Thanks Chris! Just a couple of questions.

e2e/tests/start_call.spec.ts
cpoile: This was to fix the e2e tests failing -- the tests were getting hung up because it couldn't find the edit button, so I changed it to use the keyboard shortcut.
I'm not sure if they fixed it since then, but that's what I found in the playwright playback above the [67879fd](https://github.com/mattermost/mattermost-plugin-calls/pull/741/commits/67879fd4a0d2f54d507c96c4d209616301d56ab9) fix, and locally.

webapp/src/components/call_widget/component.tsx
cpoile: The lock doesn't need to be in the state because setting it or unsetting it won't affect the UI, it's only the `setState` that it prevents/allows that will. Hope that makes it clearer?

e2e/tests/start_call.spec.ts
streamer45: Cool, could we add a comment? Because the fact that we are typing the message right after makes this extra confusing.

webapp/src/components/call_widget/component.tsx
streamer45: Ah, we keep the lock for the whole wait period. That makes sense, thank you!

PR #742, opened 2024-05-17
[MM-58325] Handle stuck session edge case

streamer45:
#### Summary

Ceiling tests have revealed a race condition while leaving a call before `handleJoin` has finished that could lead to indefinitely stuck sessions. 

A proper fix would be quite challenging so I am opting for a "good enough" workaround. If upon WebSocket disconnects we don't find the session we try again after a few seconds, giving enough time to any potentially concurrent `handleJoin` execution to complete.
 
#### Ticket Link

https://mattermost.atlassian.net/browse/MM-58325


cpoile (approved):
Considering the uniqueness of connID, I don't see any problems. 👍

PR #745, opened 2024-05-21
E2E test to check if call button in user profile popover is disabled after call is started

M-ZubairAhmed:
#### Ticket Link
https://mattermost.atlassian.net/browse/MM-58271


yasserfaraazkhan (approved):
🎉

e2e/tests/join_call.spec.ts
```diff
@@ -90,10 +90,16 @@ test.describe('join call', () => {
...
         await userAPage.locator('.post__header').locator('button.user-popover').last().click();
         await expect(userAPage.locator('div.user-profile-popover')).toBeVisible();
-        await userAPage.locator('div.user-profile-popover').locator('#startCallButton').click();
+
+        // Start the call with the button
+        await expect(userAPage.getByLabel('Start call')).toBeEnabled();
+        await userAPage.getByLabel('Start call').click();
+
+        // Verify that the call button is disabled while the call is started
+        await expect(userAPage.locator('#startCallButton')).toBeDisabled();
```
cpoile, line 99: Would you be able to narrow the selector to the `div.user-profile-popover` as before, just to make sure we don't hit another element that might be added later?
Otherwise, looks great!

M-ZubairAhmed: I improved the locator by using more specific label, please check it out

cpoile (approved):
Thanks Zubair!

PR #747, opened 2024-05-21
[MM-58259] [MM-58260] Move the /call end command to host controls

cpoile:
#### Summary
- Move the `/call end` command to the host controls router
- for [MM-58260]: Adjusted some of the checks in the toast and the channel link icon so that they properly disappear after `/call end` -- before they would stay visible after `/call end`, now they are gone. Let me know if there was a particular reason we were using profile count as a proxy for the call being active -- if so, then maybe should fix that instead.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-58259
- https://mattermost.atlassian.net/browse/MM-58260


streamer45:
> Let me know if there was a particular reason we were using profile count as a proxy for the call being active -- if so, then maybe should fix that instead.

I think there is a reason because I remember fixing related issues. I think this way you could still get the toast to show even when leaving a call (only you in the call). The call technically ends but we are not sending any event to clear the state.

cpoile:
@streamer45 Ah, I see that now. So a couple thoughts, 1) I'll add an e2e test for that, cause that obviously shouldn't be happening when then call is over, 2) What do you think about sending a `wsEventCallEnd` event when we update the call's state `EndAt`? Since call is over, we should make sure the client knows it. Otherwise it feels like the participant counting is a hack.

streamer45:
> @streamer45 Ah, I see that now. So a couple thoughts, 1) I'll add an e2e test for that, cause that obviously shouldn't be happening when then call is over, 2) What do you think about sending a `wsEventCallEnd` event when we update the call's state `EndAt`? Since call is over, we should make sure the client knows it. Otherwise it feels like the participant counting is a hack.

Yes, I think that makes sense. Let's just make sure this is tracked through the unique call ID so that we don't get into funny situations since a new call could start right after.

webapp/src/index.tsx
```diff
@@ -327,7 +326,7 @@ export default class Plugin {
 
                 // following the thread only on join. On call start
                 // this is done in the call_start ws event handler.
-                if (profilesInCallInChannel(store.getState(), channelId).length > 0) {
+                if (channelHasCall(store.getState(), channelId)) {
```
cpoile, line 329: The idea behind these refactorings are to move logic (like length tests) into the selectors, and make the selector's name declare what it does. This means more selectors, but it's clearer and the logic is isolated to one place. 
Then, if we decide that counting participants is more correct, we can make that change in the selector.

webapp/src/actions.ts
```diff
@@ -402,8 +402,14 @@ export const userLeft = (channelID: string, userID: string, sessionID: string) =
             },
         });
 
-        if (ringingEnabled(getState()) && !channelHasCall(getState(), channelID)) {
-            await dispatch(removeIncomingCallNotification(callID));
+        if (numSessionsInCallInChannel(getState(), channelID) === 0) {
```
cpoile, line 405: Noting this because it's a big change. This is what we do in mobile and it seems to simplify things.

webapp/src/websocket_handlers.ts
```diff
@@ -92,25 +92,21 @@ import {
 
 // NOTE: it's important this function is kept synchronous in order to guarantee the order of
 // state mutating operations.
-export function handleCallEnd(store: Store, ev: WebSocketMessage<EmptyData>) {
+export async function handleCallEnd(store: Store, ev: WebSocketMessage<EmptyData>) {
```
streamer45, line 95: Was the comment unclear, or am I missing something? :stuck_out_tongue: 
Either way, something has to go.

cpoile: I was playing around with a different refactor and that snuck through.

streamer45 (commented):
Thanks for taking the time to review this. The approach makes sense to me. 

I am still seeing some weirdness but at this point I can't easily tell if it's a regression or not but still something we could look into as part of this PR.

If a user joins a call and then leaves I think we should show the toast once more. This doesn't seem to happen. What's weirder is that not even refreshing gets it back.

cpoile:
@streamer45 We changed that behavior back in #445: dismissing the toast or joining (via toast or post button) dismisses the banner and it won't come back. If we want to change that behavior, then maybe we should open a new ticket?

streamer45:
> @streamer45 We changed that behavior back in #445: dismissing the toast or joining (via toast or post button) dismisses the banner and it won't come back. If we want to change that behavior, then maybe we should open a new ticket?

Ah, forgot about that. Never mind then if it's working as designed.

webapp/src/actions.ts
streamer45: I implicitly thought we would handle this case server-side and send the unique call ID because this does not protect us against a race, as we still use the channel ID to figure out the locally stored call ID. It's not a huge deal, but it's something to note.

streamer45 (approved):
Looks good, nothing blocking.

streamer45:
Yeah, I haven't checked the whole path, including corner cases like WS reconnects, but if it worked fine on mobile until now, we should be good.

streamer45:
@cpoile Can we merge? So I can prepare a build for Community and start the soaking process.

PR #748, opened 2024-05-22
feat: Introduce new notice-file-generation

toninis:
## Summary

Introduce new packages on NOTICE.txt file 
Related PR: https://github.com/mattermost/delivery-platform/pull/121 

## Ticket Link

https://mattermost.atlassian.net/browse/CLD-6877

### DIFF
--- a/.config/notice-file/config.yaml
+++ b/.config/notice-file/config.yaml
@@ -5,6 +5,8 @@ description: "This document includes a list of open source components used in Ma
 search:
   - "go.mod"
   - "build/go.mod"
+  - "server/public/go.mod"
+  - "standalone/package.json"
   - "webapp/package.json"
 dependencies: []
 devDependencies: []
--- a/.config/notice-file/config.yaml
+++ b/.config/notice-file/config.yaml
@@ -1,12 +1,10 @@
 ---
-
 title: "Mattermost Calls"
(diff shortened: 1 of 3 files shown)


amyblais (approved):

.config/notice-file/config.yaml
```diff
@@ -1,12 +1,10 @@
 ---
-
 title: "Mattermost Calls"
 copyright: "©2015-present Mattermost,Inc. All Rights Reserved. See LICENSE for license information."
 description: "This document includes a list of open source components used in Mattermost Calls plugin, including those that have been modified."
 search:
   - "go.mod"
   - "build/go.mod"
+  - "webapp/package.json"
```
streamer45, line 8: Technically speaking we also have `standalone/package.json` which mostly overlaps with `webapp/package.json` but it's a different thing. We also have `server/public/go.mod` and `lt/go.mod`.

toninis: We mainly need the ones that are actually used on production . Which files depict those dependencies ?

streamer45: All of those, except `lt/go.mod` would be part of our builds.

streamer45: From the ones there already, `build/go.mod` wouldn't really be part of our releases. It's just a dev dependency.

streamer45 (approved):
Thanks!

PR #750, opened 2024-05-23
[MM-54335] Ceiling tests: load-test client improvements

streamer45:
#### Summary

PR implements some improvements that were made necessary while performing v1 of our ceiling tests, namely:

- Using structured logging. This is very helpful as now we are able to identify each log line with the client/user, even those coming from the layers below (rtcd client)
- Added some logic to re-initialize the client in case of a state modification error, something quite likely at scale since Pion doesn't yet support [perfect negotiation](https://developer.mozilla.org/en-US/docs/Web/API/WebRTC_API/Perfect_negotiation) (sdp rollbacks and such).
- Fixed some potential races and made using channels more consistent.

#### Related PR

https://github.com/mattermost/rtcd/pull/144

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-54335


cpoile (approved):
Looks great, thanks!

PR #754, opened 2024-05-27
[MM-53491] Fix rendering of active call icon in sidebar's channel link

streamer45:
#### Summary

PR fixes some outstanding issues with the rendering of the active call icon in the sidebar. This requires some webapp changes to work in all cases (see related PR).

#### Design

https://www.figma.com/design/KAdDampXIcjzQ8oWMs7LpZ/MM-53491-Calls%3A-Rendering-of-LHS-active-call-icon?node-id=1-7&m=dev

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-53491

#### Related PR

https://github.com/mattermost/mattermost/pull/27172

### DIFF
--- a/webapp/src/components/channel_link_label/component.tsx
+++ b/webapp/src/components/channel_link_label/component.tsx
@@ -47,14 +47,17 @@ const ChannelLinkLabel = (props: Props) => {
                     {
                         display: 'flex',
                         alignItems: 'center',
-                        marginLeft: 'auto',
+                        justifyContent: 'center',
                         height: 'auto',
+                        minWidth: '20px',
+                        marginLeft: 'auto',
+                        marginRight: '4px',
                     }
                 }
             >
                 <ActiveCallIcon
                     fill={props.theme.sidebarText}
-                    style={{width: '12px', height: '12px'}}
+                    style={{minWidth: '12px', maxWidth: '12px', minHeight: '12px', maxHeight: '12px'}}
                 />
             </span>
         </OverlayTrigger>


webapp/src/components/channel_link_label/component.tsx
```diff
@@ -47,14 +47,17 @@ const ChannelLinkLabel = (props: Props) => {
...
+                        minWidth: '20px',
+                        marginLeft: 'auto',
+                        marginRight: '4px',
                     }
                 }
             >
                 <ActiveCallIcon
                     fill={props.theme.sidebarText}
-                    style={{width: '12px', height: '12px'}}
+                    style={{minWidth: '12px', maxWidth: '12px', minHeight: '12px', maxHeight: '12px'}}
```
cpoile, line 60: I'm assuming that something is making the width and height not work correctly?

cpoile (approved):
Thanks!

webapp/src/components/channel_link_label/component.tsx
streamer45: @cpoile Not exactly sure since I couldn't reproduce but there was an instance in which the icon would collapse:

image

abhijit-singh (changes requested):
Thanks @streamer45! Definitely better than before, but I wanted to check if we can do a couple of things :
1. ~There's slightly different behaviour if the channel/user name gets truncated or not-
If the text is getting truncated, the call icon shifts on hover and stays visible while the text gets truncated further. This works as expected.~
~If it text is not getting truncated, the call icon gets hidden on hover. Ideally, the call icon should shift in this case as well and the text should get hidden/truncated behind it.~

2. ~This is small but since we're here, can we make sure there's always 4px space between the user name and the calls icon. It gets very close right now. Checked both with and without a custom status.~
Current:
image
image
Ideal:
image
image

streamer45:
@abhijit-singh I am not able to reproduce the above. Are you testing on the matching [webapp branch](https://github.com/mattermost/mattermost/tree/MM-53491)? The channel name should always be truncated if there's not enough space.

abhijit-singh:
Yep, I am on that branch. That state is when the name is shorter or the sidebar width has been increased. If the name doesn't truncate on first load, it doesn't truncate on hover as well.

streamer45:
> If the name doesn't truncate on first load, it doesn't truncate on hover as well.

Mmm, that doesn't sound like the behaviour I implemented on that side. It should truncate, regardless of the initial state. Could you just make sure there's no `truncated` class in the DOM any longer?

abhijit-singh (approved):
Looks good, thanks!

DHaussermann (approved):
Tested and passed 
I tested this change as part of https://github.com/mattermost/mattermost/pull/27172

No issues found.

PR #756, opened 2024-05-28
Fix leaking batchers in HA environment

streamer45:
#### Summary

When implementing the cleanup code in https://github.com/mattermost/mattermost-plugin-calls/pull/704, I failed to realize that this path would only be running on a single HA node, whichever has the lock when removing the last session from the global state.

This caused some of the batchers in other nodes to be leaked. By itself, this wasn't a huge deal until they would be re-used the next time around for a call in the same channel. That would cause a much nastier bug since the [session count](https://github.com/mattermost/mattermost-plugin-calls/blob/b501c6be2becca2fcc7a4403938da8380314dcd1/server/session.go#L536) of the previous call would determine the queue length so any larger call would have had a high chance of causing channel full errors and batched actions to be dropped, leading to an inconsistent state.

We are fixing this but cleaning up batchers once all sessions for the call have been cleared from the local map (`p.sessions`).

### DIFF
--- a/server/session.go
+++ b/server/session.go
@@ -433,24 +433,6 @@ func (p *Plugin) removeUserSession(state *callState, userID, originalConnID, con
 		}
 		setCallEnded(&state.Call)
 
-		p.mut.Lock()
-		if batcher := p.addSessionsBatchers[channelID]; batcher != nil {
-			p.LogDebug("stopping batcher for call", "channelID", channelID)
-			p.addSessionsBatchers[channelID] = nil
-			delete(p.addSessionsBatchers, channelID)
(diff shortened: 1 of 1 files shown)


server/session.go
```diff
@@ -610,3 +623,12 @@ func (p *Plugin) getSessionByOriginalID(sessionID string) *session {
...
 }
+
+func (p *Plugin) hasSessionsForCall(callID string) bool {
+	for _, s := range p.sessions {
+		if s.callID == callID {
+			return true
+		}
+	}
+	return false
+}
```
streamer45, line 634: This is fine for now. At some point, it may be worth mapping sessions to their call ID.

cpoile (approved):
👍

PR #762, opened 2024-06-04
[MM-58017] Calls usage stats

streamer45:
#### Summary

PR implements some basic usage stats for Calls to be rendered in the System Statistics page (`/admin_console/reporting/system_analytics`), namely:

- Total Calls
- Total Active Calls (i.e., ongoing)
- Total Active Sessions
- Calls by day
- Calls by month
- Calls by channel type
- Average call duration
- Average call participants

Note: the last commit is purely about fixing type errors after updating the web app dependency.

#### Screenshots

image
image
image

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-58017


server/db/stats_store.go
```diff
@@ -0,0 +1,329 @@
...
+	ctx, cancel := context.WithTimeout(context.Background(), time.Duration(*s.settings.QueryTimeout)*time.Second)
+	defer cancel()
+	if err := s.rDBx.GetContext(ctx, &count, q, args...); err != nil {
+		return 0, fmt.Errorf("failed to get calls sessions count: %w", err)
+	}
+
+	return count, nil
+}
+
+func (s *Store) GetTotalCalls(active bool) (int64, error) {
```
streamer45, line 117: Keeping it simple for now, but ideally, most of these methods could filter for the last month or so. That way they would perform much better.

server/db/stats_store.go
```diff
@@ -0,0 +1,329 @@
...
+	qb := getQueryBuilder(s.driverName).
+		Select("to_char(to_timestamp(startat / 1000), 'YYYY-MM') AS Month, COUNT(*) AS Count")
+	if s.driverName == model.DatabaseDriverMysql {
+		qb = getQueryBuilder(s.driverName).
+			Select("DATE_FORMAT(FROM_UNIXTIME(startat / 1000), '%Y-%m') AS Month, COUNT(*) AS Count")
+	}
+	qb = qb.From("calls").Where(sq.And{
+		sq.Expr("EndAt > calls.StartAt"),
+		sq.Eq{"DeleteAt": 0},
+		sq.GtOrEq{"StartAt": time.Now().AddDate(0, -12, 0).UnixMilli()},
```
streamer45, line 204: To make this more performant we could add an index on `calls.StartAt`.

server/api.go
```diff
@@ -534,3 +534,19 @@ func (p *Plugin) ServeHTTP(_ *plugin.Context, w http.ResponseWriter, r *http.Req
...
 	p.apiRouter.ServeHTTP(w, r)
 }
+
+func (p *Plugin) handleGetStats(w http.ResponseWriter) error {
+	stats, err := p.store.GetCallsStats()
+	if err != nil {
+		return fmt.Errorf("failed to get stats from store: %w", err)
+	}
+
+	// TODO: consider implementing some caching for heaviest queries
```
cpoile, line 544: Would you consider making it a ticket? (or did you mean to clean this up before submitting?)

server/db/stats_store.go
```diff
@@ -0,0 +1,329 @@
...
+		Count int64
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), time.Duration(*s.settings.QueryTimeout)*time.Second)
+	defer cancel()
+	if err := s.rDBx.SelectContext(ctx, &rows, q, args...); err != nil {
+		return nil, fmt.Errorf("failed to run query: %w", err)
+	}
+
+	m := make(map[string]int64, 30)
+	for i := 0; i < 30; i++ {
```
cpoile, line 268: I don't really mind, but does it matter (30 vs 28 vs 31)?

server/api_router.go
```diff
@@ -148,6 +148,18 @@ func (p *Plugin) newAPIRouter() *mux.Router {
 		}
 	}).Methods("POST")
 
+	// Stats
+	router.HandleFunc("/stats", func(w http.ResponseWriter, r *http.Request) {
```
cpoile, line 152: Considering the weight of the call, should we at least rate limit this pretty heavily?

webapp/src/index.tsx
```diff
@@ -429,6 +432,107 @@ export default class Plugin {
         registry.registerAdminConsoleCustomSetting('ICEHostPortOverride', ICEHostPortOverride);
         registry.registerAdminConsoleCustomSetting('ServerSideTURN', ServerSideTURN);
 
+        registry.registerSiteStatisticsHandler(async () => {
```
cpoile, line 435: Would it be cleaner to put this fn in its own file? I only ask because it seems a little out of place considering the conciseness of the other register calls above.

cpoile (commented):
Just minor things, looks great!

server/api.go
streamer45: Yes, I can make a ticket. I just don't know yet if this is really needed. I almost feel it may be best to go with https://github.com/mattermost/mattermost-plugin-calls/pull/762#discussion_r1626483383.

server/db/stats_store.go
streamer45: Tracking the last 30 days seemed more straightforward than having to deal with month-by-month cases.

server/api_router.go
streamer45: We could, but the global rate limiter we have on API endpoints would already apply to this admin-only endpoint coming from a pretty specific page, so I am not sure if the extra complication is worth it.

webapp/src/index.tsx
streamer45: Yes, I was planning to do it but then I couldn't think of a great place. I can add it to utils I suppose.

streamer45:
@cpoile Let me know if there's anything else blocking.

cpoile (approved):
Awesome, looks great!

streamer45:
@DHaussermann This is the matching PR you can use to test https://github.com/mattermost/mattermost/pull/27188 if you wish.

DHaussermann:
@streamer45 I have this deployed and took a look ver things. I'm a bit confused by what should be visible and testable right now. 

The data I **can** see appears accurate after creating a few data points but... I can't see the following panels

- Calls by day
- Calls by month
- Calls by channel type (donut graph)
- Daily Calls (line graph)
- Monthly Calls (line graph)

0/5 Perhaps these are being hidden because it's a new server and I don't have data points (though I'd still want the donut graph of calls per channel type)

streamer45:
> @streamer45 I have this deployed and took a look ver things. I'm a bit confused by what should be visible and testable right now.
> 
> The data I **can** see appears accurate after creating a few data points but... I can't see the following panels
> 
> * Calls by day
> * Calls by month
> * Calls by channel type (donut graph)
> * Daily Calls (line graph)
> * Monthly Calls (line graph)
> 
> 0/5 Perhaps these are being hidden because it's a new server and I don't have data points (though I'd still want the donut graph of calls per channel type)

@DHaussermann That's likely because of the version check. Those would only display on v9.10 so you may need to check on a release candidate or something. We can defer to later if it helps.

DHaussermann:
Okay thanks. I was wondering that. I tried to work around this be deploying on `master` but then the minimum version is not met :)  

Yes - it would be reasonable to test the remaining panels post merge in an RC. 👍

streamer45:
@DHaussermann Will go ahead and prepare this to be merged then to unblock the release. We can come back to this again once we have an RC.

DHaussermann (approved):
Tested and passed 
I built some data for this and then observed the stats. The figures shown seem accurate based on the Calls usage eon the server for the following:
- Total Calls
- Total Active Calls
- Total Active Sessions
- Avg Call Duration (minutes)
- Avg Call Participants

LGTM!

As mentioned above, I will test the remaining panels post merge.

ThiefMaster:
Is "Calls by channel type" only available in EE by any chance? If yes, that feels rather strange considering that with #809 you paywall a feature that people may currently be using, but at the same time make it impossible for admins of the free version to actually see if and how much it is being used (which may actually be information that could encourage someone to buy a license!)...

streamer45:
> Is "Calls by channel type" only available in EE by any chance? If yes, that feels rather strange considering that with #809 you paywall a feature that people may currently be using, but at the same time make it impossible for admins of the free version to actually see if and how much it is being used (which may actually be information that could encourage someone to buy a license!)...

@ThiefMaster They require a newer MM server version to be rendered (v9.10+ I believe) and not just plugin v0.28+. You can always check whether `http://localhost:8065/plugins/com.mattermost.calls/stats` returns something.

ThiefMaster:
Ah, good point. I'm still on v9.9.x. Anyway, I spotted the stats endpoint in the meantime and got the information from there :)

PR #764, opened 2024-06-04
[MM-58544] Fix call_end for desktop

cpoile:
#### Summary
- Fixing how we handle call_end, which left the desktop thinking we were still in a call:

 

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-58544


cpoile:
Ideally we would type all redux events. But it would be up to us to keep that up to date.

PR #768, opened 2024-06-06
[MM-58494] Fix duplicate ws event to bot

streamer45:
#### Summary

Fixing a small regression introduced in https://github.com/mattermost/mattermost-plugin-calls/pull/609 that caused sending duplicate events to the bot, causing the double reaction issue.

This was unexpected primarily because of the interesting precedence rules in WebSocket broadcasts:

https://github.com/mattermost/mattermost/blob/04181247f8a42fe8313a28df864224cd2e4850e8/server/channels/app/platform/web_conn.go#L867-L877 

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-58494


cpoile (approved):
Nice, thanks!

PR #769, opened 2024-06-06
[MM-58026] Plugin icon :telephone_receiver: 

streamer45:
#### Summary

Roughly two years after the first release, we have a plugin icon :tada:

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-58026


plugin.json
```diff
@@ -4,6 +4,7 @@
   "description": "Integrates real-time voice communication in Mattermost",
   "homepage_url": "https://github.com/mattermost/mattermost-plugin-calls/",
   "support_url": "https://github.com/mattermost/mattermost-plugin-calls/issues",
+  "icon_path": "server/assets/plugin_icon.svg",
```
streamer45, line 7: @cpoile Do you know whether this is enough to get the respective Markeplace metadata updated automatically once we publish the release or if we should update that side manually? Thanks.

cpoile: Yes that should be enough -- the marketplace `add` cmd should extract it and encode it for the marketplace.

PR #774, opened 2024-06-13
[MM-58285] Fix expanding call thread

streamer45:
#### Summary

PR fixes an issue with the call thread rendering on top of the call view, causing some buttons on the right side (e.g. leave call) to be hidden in some instances.

I had to wrestle with CSS to make sure I didn't break previously fixed things, namely:

- Call settings menus rendering on top of the call thread in case of overlap.
- Reactions rendering on top of calls view.
- Call view rendering on top of the annoying onboarding button.

Please give it a try and let me now if anything unexpected broken due to this.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-58285

### DIFF
--- a/e2e/tests/popout.spec.ts
+++ b/e2e/tests/popout.spec.ts
@@ -76,6 +76,25 @@ test.describe('popout window', () => {
         await expect(page.page.locator('#calls-widget')).toBeHidden();
     });
 
+    test('expanding chat', async () => {
+        const [page, popOut] = await startCallAndPopout(userStorages[0]);
+        await expect(popOut.page.locator('#calls-expanded-view')).toBeVisible();
+
+        // Open chat thread
+        await popOut.page.click('#calls-popout-chat-button');
+        await expect(popOut.page.locator('#sidebar-right [data-testid=call-thread]')).toBeVisible();
+
+        // Expand call thread
+        await popOut.page.locator('[aria-label="Expand"]').click();
+
+        // Verify leave button can be clicked. Checking for visibility would work even
+        // if there's an element showing on top
+        await popOut.page.locator('#calls-popout-leave-button').click();
+        await popOut.page.getByText('Leave call').click();
+
+        await expect(page.page.locator('#calls-widget')).toBeHidden();
+    });
+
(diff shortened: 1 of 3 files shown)


matthewbirtch (approved):
This looks like it fixed it @streamer45. 

I'm not sure if this is related to this PR, but if I start a call from a DM it doesn't seem to respect the active theme. This only seems to be an issue in the desktop app though. I don't see this happening on the community server.

image

streamer45:
To make this work, I had to move the element in which we set the background so it may well be related. I'll take a look.

streamer45:
@matthewbirtch It looks like this may have been broken for a while. I don't see the theme updating on the component, so it's not dynamic any longer, but it should work if you reopen the popout after changing the theme.

matthewbirtch:
> @matthewbirtch It looks like this may have been broken for a while. I don't see the theme updating on the component, so it's not dynamic any longer, but it should work if you reopen the popout after changing the theme.

It doesn't seem to. I've tried changing my theme then starting a new call and opening up in the popout, but it doesn't seem to respect the active theme.

The issue seems to be isolated to the desktop app though and only in DMs, which is weird.

streamer45:
I cannot reproduce it, I tested it in a DM on a Desktop, and if opened after the theme has changed, the color is updated. Maybe it's worth trying a hard refresh on both sides?

matthewbirtch:
> I cannot reproduce it, I tested it in a DM on Desktop, and if opened after the theme has changed, the color is updated. Maybe it's worth trying a hard refresh on both sides?

I tried hard refreshes multiple times. But let me restart the server and try again

cpoile (approved):
Works well!

One suggestion, maybe we should put a right margin of 8px on the participants button, and a left margin of 8px on the leave button, so that when it's collapsed we'll have consistent spacing?

image

streamer45:
CSS margins :grimacing: , let me review.

cpoile:
Or better yet, put a horizontal padding on the center, that way it won't potentially interfere with the complicated leave button.

streamer45:
@cpoile I added a flex gap on the parent. Let me know if that works.

cpoile:
Yep, best way to go, works well, thanks!

PR #778, opened 2024-06-13
Fix asset bundling

cpoile:
#### Summary
- We made a little mistake in bundling. Instead of splitting the files (png staying in the server pkg so we can embed it, and svg moving to the `/assets` dir) to conform to the makefile, just changed the makefile's asset directory var. Also need to change the `plugin.json` location.
- Will need to cut `v0.28.1` after. I did not delete the `v0.28.0`, but I wanted to just to get the automatic comment. :)

#### Ticket Link
- none


streamer45 (approved):
Makes sense, thanks!

PR #783, opened 2024-06-17
[MM-54256] Server side translations support

streamer45:
#### Summary

PR adds server-side translation support. Given that we translate on the client, this shouldn't affect most existing messages, but it will simplify the implementation of https://mattermost.atlassian.net/browse/MM-54847 and a few other things (e.g., certain packaging changes).

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-54256


server/utils.go
```diff
@@ -220,3 +221,19 @@ func getUserIDsFromSessions(sessions map[string]*public.CallSession) []string {
...
+		return i18n.GetUserTranslations(locale)
+	}
+
+	locale = "en"
+	cfg := p.API.GetConfig()
+	if cfg == nil {
+		p.LogError("failed to get configuration")
+	} else if cfg.LocalizationSettings.DefaultClientLocale != nil {
+		locale = *cfg.LocalizationSettings.DefaultClientLocale
+	}
```
streamer45, line 236: We are currently defaulting to this to determine the locale for messages in call posts, given that they are channel-wide.

server/bot_api.go
```diff
@@ -227,10 +227,12 @@ func (p *Plugin) handleBotPostRecordings(w http.ResponseWriter, r *http.Request)
 	}
 	post.AddProp("recording_files", recordingFiles)
 
+	T := p.getTranslationFunc("")
+
 	startAt, _ := post.GetProp("start_at").(int64)
-	postMsg := "Here's the call recording"
+	postMsg := T("app.call.new_recording_message")
```
cpoile, line 233: oh no... we have to name them? boo

server/utils.go
cpoile: Makes sense... I hope they set it. :)

cpoile (approved):
Cool, thanks for adding that!

server/bot_api.go
streamer45: Yeah, although it shouldn't be too challenging to generate hash-based IDs I suppose but it'll require changes in https://github.com/mattermost/mattermost-utilities/tree/master/mmgotool

PR #784, opened 2024-06-17
[MM-58773] Fix broken notifications test

cpoile:
#### Summary
- Actual broken test, bisected the problem to a change in how mattermost mounts react components: https://github.com/mattermost/mattermost/pull/27192
- Looks like the e2e test wasn't detecting a real break, I think it just runs too quickly. So: 1) ensuring that patching is applied by reloading after (kind of unrelated to the above PR but might have been contributing to flakiness in the past), and 2) adding a timeout of 0 to flush the event loop of the page (allowing react to fully load), and that fixes it.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-58773

### DIFF
--- a/e2e/tests/notifications.spec.ts
+++ b/e2e/tests/notifications.spec.ts
@@ -103,7 +103,6 @@ test.describe('notifications', () => {
     test('dm channel, global desktop none', async ({page, request}) => {
         await apiPatchNotifyProps(request, {desktop: 'none'});
         await page.reload();
-        await page.waitForTimeout(0);
         const devPage = new PlaywrightDevPage(page);
         await page.evaluate(() => {
             window.e2eDesktopNotificationsRejected = [];
@@ -130,7 +129,6 @@ test.describe('notifications', () => {
     test('dm channel, global sound false', async ({page, request}) => {
         await apiPatchNotifyProps(request, {desktop: 'mentions', calls_desktop_sound: 'false'});
         await page.reload();
-        await page.waitForTimeout(0);
         const devPage = new PlaywrightDevPage(page);
         await page.reload();
         await page.evaluate(() => {
@@ -196,10 +194,8 @@ test.describe('notifications', () => {
     test('gm channel, global desktop none', async ({page, request}) => {
         await apiPatchNotifyProps(request, {desktop: 'none'});
         await page.reload();
-        await page.waitForTimeout(0);
         const devPage = new PlaywrightDevPage(page);
         await devPage.goto();
-        await page.waitForTimeout(0);
         await page.evaluate(() => {
             window.e2eDesktopNotificationsRejected = [];
             window.e2eNotificationsSoundedAt = [];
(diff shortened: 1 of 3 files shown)


streamer45 (commented):
Thanks for looking into this!

According to the docs, use of `waitForTimeout` is discouraged. I also don't fully understand why we need to reload the page. I wonder if simply waiting for the `domcontentloaded` event could be enough given we are not relying on that from the webapp side?

cpoile:
@streamer45 I did another dive into what was happening, and I think I ruled out that we need the timeout, so removed that. The reload is where the fix is. So I had the reasoning in the summary reversed: the reload is ensuring that the status and notify props patch has been received by the client (which is then used by the webapp and plugin to decide wether or not to send a desktop notification). I'm not sure how the webapp PR affected this, but I tested the before and after multiple times and it seems to be the difference that caused the failing tests.

cpoile:
I seem to be hitting the `close with comment` with the key combo, sorry.

streamer45:
> @streamer45 I did another dive into what was happening, and I think I ruled out that we need the timeout, so removed that. The reload is where the fix is. So I had the reasoning in the summary reversed: the reload is ensuring that the status and notify props patch has been received by the client (which is then used by the webapp and plugin to decide wether or not to send a desktop notification). I'm not sure how the webapp PR affected this, but I tested the before and after multiple times and it seems to be the difference that caused the failing tests.

Interesting. Is there a chance that making `ringForCall` synchronous is sufficient? I still somewhat struggle to understand why reloading would be needed :thinking:

cpoile:
@streamer45 `ringForCall` is synchronous. So my best guess is that because we're patching from the API, we have to wait for the websocket to update the client. Reloading ensures that.

streamer45:
> @streamer45 `ringForCall` is synchronous. So my best guess is that because we're patching from the API, we have to wait for the websocket to update the client. Reloading ensures that.

But you made it synchronous in this PR, am I missing something?

cpoile:
@streamer45 I made it synchronous because I noticed that it doesn't need to be async. It doesn't fix the problem, but I left it there anyway because its the right code. Does that help, or maybe I'm misunderstanding what you mean?

streamer45:
> @streamer45 I made it synchronous because I noticed that it doesn't need to be async. It doesn't fix the problem, but I left it there anyway because its the right code. Does that help, or maybe I'm misunderstanding what you mean?

That makes sense, I was just wondering if making it synchronous was sufficient for some reason. All good :+1:

e2e/tests/notifications.spec.ts
```diff
@@ -128,8 +128,9 @@ test.describe('notifications', () => {
 
     test('dm channel, global sound false', async ({page, request}) => {
         await apiPatchNotifyProps(request, {desktop: 'mentions', calls_desktop_sound: 'false'});
+        await page.reload();
         const devPage = new PlaywrightDevPage(page);
-        await devPage.goto();
+        await page.reload();
```
streamer45, line 133: Is the double reload needed here?

streamer45 (approved):
Thanks for taking the time to dig into this!

e2e/tests/notifications.spec.ts
cpoile: Ah, thanks, missed that.

PR #786, opened 2024-06-18
[MM-57220] Azure AI transcriptions

streamer45:
#### Summary

PR enables configuration of Azure AI powered post call transcriptions.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-57220


webapp/src/components/admin_console_settings/recordings/transcriber_api_azure_speech_key.tsx
```diff
@@ -0,0 +1,70 @@
...
+    const recordingEnabled = useSelector(recordingsEnabled);
+    const transcriptionEnabled = useSelector(transcriptionsEnabled);
+    const api = useSelector(transcribeAPI);
+
+    if (cloud || restricted || !recordingEnabled || !transcriptionEnabled || api !== TranscribeAPI.AzureAI) {
+        return null;
+    }
+
+    // Webapp doesn't pass the placeholder setting.
+    const placeholder = manifest.settings_schema?.settings.find((e) => e.key === 'TranscribeAPIAzureSpeechKey')?.placeholder || '';
```
cpoile, line 27: We can remove this, right?

webapp/src/components/admin_console_settings/recordings/transcriber_api_azure_speech_region.tsx
```diff
@@ -0,0 +1,70 @@
...
+    const recordingEnabled = useSelector(recordingsEnabled);
+    const transcriptionEnabled = useSelector(transcriptionsEnabled);
+    const api = useSelector(transcribeAPI);
+
+    if (cloud || restricted || !recordingEnabled || !transcriptionEnabled || api !== TranscribeAPI.AzureAI) {
+        return null;
+    }
+
+    // Webapp doesn't pass the placeholder setting.
+    const placeholder = manifest.settings_schema?.settings.find((e) => e.key === 'TranscribeAPIAzureSpeechRegion')?.placeholder || '';
```
cpoile, line 27: And this?

server/configuration.go
```diff
@@ -454,6 +470,12 @@ func (c *configuration) liveCaptionsEnabled() bool {
 	return false
 }
 
+func (p *Plugin) getAdminConfig() *configuration {
+	c := p.getConfiguration()
+	c.clientConfig = p.getClientConfig()
```
cpoile, line 475: This is easier for sure, but I wonder if it might cause confusion down the road? I've found that knowing the shape of `clientConfig` has helped keep track of what we send over and what we need to handle on the client (e.g. when double-checking that the shape of the callsConfig is correct). If we send over the entire `configuration` we'll now have to keep track of what part of that we actually use on the client -- we'll have to search through and try to reason about it (like I did right now to make sure we only need `TranscribeAPI` in the `callsConfig`). A getAdminConfig that specifies what we send over would be easier, imo.

cpoile (commented):
Awesome! Really only one comment, otherwise looks great.

server/configuration.go
streamer45: @cpoile I think admins should be allowed to read the whole thing since they can configure it in the first place so the shape would the full object, aka `configuration`. Creating another subset feels just more indirection to me, but I'm happy to reconsider.

webapp/src/components/admin_console_settings/recordings/transcriber_api_azure_speech_key.tsx
streamer45: I guess so, done!

server/configuration.go
cpoile: Certainly they should be able to see the configuration, but I'm thinking more about us trying to make sure we know what it is we're using from that. Maybe I'm just worried that at some point we'll use another field somewhere and forget to add it to the callsConfig because we don't need it stored in the state, and now we've got a hidden dependency. Just a thought, I'll let you decide what's best. :)

cpoile (approved):
Nice!

server/configuration.go
streamer45: @cpoile I reworked this a bit to keep it scoped to [what we added on the client side](https://github.com/mattermost/calls-common/commit/e463e78cc76302d1bdbd085d9e1846c456f48d45) as well. Let me know if that's closer to what you had in mind. See https://github.com/mattermost/mattermost-plugin-calls/pull/786/commits/d8fcf87f81f5acccdcb85d94ed63e9a939b53588

cpoile (approved):
Yep, this is what I meant, I think it clarifies what we intend the client to know. Thanks!

PR #790, opened 2024-06-20
[MM-57346] Improve accessibility of Calls UIs

streamer45:
#### Summary

This PR attempts to improve the almost non-existent accessibility of our UIs. The main focus is on icon-only buttons, which were pure guesses from a screen reader standpoint as they lacked any text or description. 

I also improved some element roles and interactions using `aria-controls` and `aria-expanded` for some of the menus and dialogs we show conditionally.

I don't consider this a comprehensive solution to the accessibility problem, which would likely require a much deeper pass, but I'd like to assess where we stand after these changes are implemented to ensure at least a basic level of usability. 

I've asked @DHaussermann to QA this PR using a screen reader since he has some experience with it. 

#### Example

*Before*

image

*After*

image

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-57346


cpoile (approved):
Code looks good, I'll defer to Dylan about the functionality of it all, thanks!

DHaussermann:
@streamer45 I have done some testing with Mac Voice Over and quick look with JAWS. 
I'm no expert with screen readers and I'm new to Voice Over. That being said, I see a couple issue that seem to affect both. 

1. When I expand the `...` menu in the widget, I cannot navigate up into this menu. It's possible the tab index is weird so there is a way to get in that I can't see. But 0/5 it seems like these options are not assigned any aria labels.   
image

2. For the expanded view, I can open and navigate into the setting option for mic and audio devices but I cannot expand them. There is area labeling for the screen reader to get details on the menu options but I cannot action them. To my understanding `VO`+`spacebar` (the predefined Voice Over key)  should action the button but I can't make it happen. (I used the mouse to open it to take the screenshot. 
FWIW I check the emoji reactions and `VO`+`spacebar`  does work to action them.
image

streamer45:
Thanks @DHaussermann . The idea was to assign labels only to buttons missing any form of text-based data (i.e. icon-only buttons). Our menu items should all have an accompanying text, which I assumed could be used for accessibility purposes. That said, it's possible that the way we render them through multiple wrapping elements is making it difficult to navigate. I'll review and add some extra labels to see if it helps.

streamer45:
@DHaussermann I sent an update. Could you give it a quick pass and let me know if anything has improved?

DHaussermann:
Hi @streamer45, 

1. The widget has certainly improved. I can navigate into the new expanded menu now. And for the mic and audio options I can expand those menus, move into them to cycle through options and make a selection. 

The navigation for me feels a bit strange still.     
- It's a bit confusing because to move into the `...` menu I need to move backwards up the modal to move into it but that is consistent with the layout of visual elements
- It can be a bit counter intuitive  how to open a menu and move into the selection because your using `<` to nav down into a structure and the menu visualy on the right side of the current selection.
For these bullets I don't know that theres an established best practice or other product area to compare against. It may be common with screen readers to use either `<` or `>` to move down into a structure. Worth mentioning for context that I think it wold be common to use horizontal chevrons to cycle through a vertical set of options 0/5 So, maybe it's not worth getting too hung up on needing to use `<` to move deeper into a menu structure. 




2. The expanded view for me shows no change.  I can move to the option in the settings menu but cannot expand them image
This is curious given that for 1. these seem like the same menus I'm trying use.

DHaussermann (approved):
Tested and passed based on testing above. 
 @streamer45 has now resolved last issue where the menues could not be expanded with the keyboard. 
 
Tested the last fix on Mac Voice Over and on Windows using JAWS. Devices menus can now be expanded and options can be cycled and selected. 

LGTM!

PR #792, opened 2024-06-20
[MM-59027] Fix: Ringing not stopping after leave or /call end

cpoile:
#### Summary
- When refactoring `CALL_END` I tried to make it do too much (it removed the incoming notification, but that hid that we weren't also removing the ringing). Also, `CALL_END` was being sent in a couple places. 
- Consolidated the call end logic to one function, and it is its job to do all the call cleanup.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-59027


streamer45 (approved):
Looks good! Thanks for the quick fix.

PR #795, opened 2024-06-24
[MM-52657] AV1 support

streamer45:
#### Summary

PR adds experimental support for encoding screen-sharing tracks using the AV1 codec.

Changes should be backward compatible with regard to older clients since we must always send a fallback VP8 track.

We are also bumping the minimum `rtcd` version since some breaking changes to the join event payload were required.

#### Related PRs

https://github.com/mattermost/rtcd/pull/150

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-52657


webapp/src/client.ts
```diff
@@ -555,8 +566,21 @@ export default class CallsClient extends EventEmitter {
         };
 
         logDebug('adding stream to peer', screenStream.id);
+
+        // Always send a fallback track (VP8 encoded) for receivers that don't yet support AV1.
         this.peer.addStream(screenStream);
 
+        if (this.config.enableAV1 && this.av1Codec) {
+            if (this.config.simulcast) {
+                logWarn('both simulcast and av1 support are enabled');
```
cpoile, line 575: What about prioritizing AV1 over simulcast?
And either way, maybe we should throw a configuration error if the system admin selects both? Otherwise they will select them and then wonder why they're not seeing any improvement...

cpoile (approved):
Nice! Non-blocking suggestions. :)

webapp/src/client.ts
streamer45: @cpoile I honestly didn't want to overcomplicate it for some experimental features that we don't yet know if we want to retain. Our config can still be changed by different means (e.g., JSON file), so then we'd have to decide what to do in case the mutually exclusive settings are turned on (e.g. fail on start?).

The idea of logging the warning was that it would be trivial to determine if they fall into such a case. This is also documented in the config itself. I feel that may be enough for the time being.

streamer45:
@DHaussermann This is ready for review. What I am looking for QA wise is making sure we are not introducing regressions with screen sharing on any of our supported clients, with a special focus on mobile.

So I'd like to test the following cases:

Presenter side:

- Chrome
- Desktop App
- Firefox

Receiving side:

- Chrome
- Desktop App
- Firefox
- Android App
- iOS App

DHaussermann (approved):
Tested and passed
- Deployed the plugin locally and enabled AV1 codec  
- Tested that browsers can send and receive video for screen share
- Tested Firefox, Chrome, Safari and MM Desktop app for send and recieve
- Tested production mobile build can receive video (android and iOS)
- Tested Mobile builds from linked PR can receive video (android and iOS)
- With all tests ensure latency was about on par by scrolling vertically on a static page from the client sending and ensuring lag was nothing unexpected on the client receiving. 

LGTM!

DHaussermann:
@streamer45 a couple unrelated questions here:

1. This build won't run on a cloud server for Mattermost v9.10.0 I don't think the config make use of standalone RTCD but I see this when I try enabling the plugin. Has something changed recently on the plugin side? Or maybe it was just misconfigured in a way I did not notice.  
image

2. Probably a silly question but... Why is the share screen option sometimes in the modal and sometimes not? Is there some simple condition I have missed here? Sometimes the share option is only in the expanded view and not the modal.
image

streamer45:
@DHaussermann 

1. All Cloud servers use RTCD as it's enforced through environment overrides so it doesn't appear in the config. We haven't released v0.17.0 yet, which is required by these changes (https://github.com/mattermost/rtcd/pull/150), so this is best tested locally.
2. It looks like you may have screen sharing disabled in config (`AllowScreenSharing`). There's also a condition based on having multiple teams. With the team sidebar in place, we can expand the widget and fit more buttons, including screen sharing. Otherwise, we should have the button in the widget menu. I don't see it in either location there, though.

DHaussermann:
Thanks. Re: 2. I don't actually see any instances where the share is not in  the modal once it is enabled.

PR #794, opened 2024-06-24
Translations update from Mattermost Weblate

weblate:
Translations update from [Mattermost Weblate](https://translate.mattermost.com) for [Calls/webapp](https://translate.mattermost.com/projects/calls/webapp/).

Current translation status:

Weblate translation status

PR #796, opened 2024-06-25
fix: Remove status check posting from e2e tests

toninis:
#### Summary

We do not need to trigger the e2e report summary status check . 
We can post it after the report is generated

Fixes: https://github.com/mattermost/mattermost-plugin-calls/pull/794#issuecomment-2186422441


streamer45 (approved):
Thanks!

PR #801, opened 2024-06-26
:wrench: Fix test CI

streamer45:
#### Summary

It turns out that we have not been running our unit/integration tests on CI since May 15th. The step was running but completing in a few seconds with a `make: Nothing to be done for 'test'.` message, so we didn't spot anything. :man_facepalming: Tests were also failing, not due to a bug thankfully.

It looks like this was another side effect of some upgrades that happened in https://github.com/mattermost/actions-workflows/pull/43 and https://github.com/mattermost/actions-workflows/pull/44. 

From now on, we are pinning the Github workflow actions so that this should hopefully not happen again without explicitly upgrading the CI.


cpoile (approved):
Yikes, thanks for catching that!

PR #802, opened 2024-06-26
[MM-57782]

streamer45:
#### Summary

PR implements an admin warning in case the plugin is not using the `RTCD` service and 50 active call sessions are connected (globally).

The warning happens in real-time and there's a backoff mechanism implemented such as it shouldn't be posted again for at least a week after it triggers.

Note: I slightly reworded the messaging when compared to the Figma specs.

/cc @esethna


server/i18n/en.json
```diff
@@ -1,4 +1,20 @@
...
+    "translation": "Your installation is experiencing heavy usage from Mattermost Calls that can negatively impact the performance and responsiveness of the system."
+  },
+  {
+    "id": "app.admin.concurrent_sessions_warning.pro_or_e0",
+    "translation": "We highly recommend upgrading to [Enterprise](https://mattermost.com/pricing/) and [deploying the RTCD service](https://docs.mattermost.com/configure/calls-deployment.html#the-rtcd-service) to offload calls processing to a separate instance in order to maintain the performance, scalability and reliability of your main Mattermost server."
+  },
+  {
+    "id": "app.admin.concurrent_sessions_warning.team",
+    "translation": "We highly recommend switching to [Mattermost Enterprise Edition](https://docs.mattermost.com/install/enterprise-install-upgrade.html) and [deploying the RTCD service](https://docs.mattermost.com/configure/calls-deployment.html#the-rtcd-service) to offload calls processing to a separate instance in order to maintain the performance, scalability and reliability of your main Mattermost server."
+  },
```
streamer45, line 17: @cwarnermm This is what I am asking to review from your side.

server/limits.go
```diff
@@ -0,0 +1,243 @@
...
+	return int64(val)
+}
+
+func getConcurrentSessionsWarningBackoffTime() time.Duration {
+	val, err := time.ParseDuration(os.Getenv("MM_CALLS_CONCURRENT_SESSIONS_WARNING_BACKOFF_TIME"))
+	if err != nil {
+		return concurrentSessionsWarningBackoffTimeDefault
+	}
+	return val
+}
```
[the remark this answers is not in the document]
streamer45: @cpoile Are you there?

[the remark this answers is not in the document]
cpoile: No, MM hates me.

server/i18n/en.json
```diff
@@ -1,4 +1,20 @@
 [
+  {
+    "id": "app.admin.concurrent_sessions_warning.enterprise",
+    "translation": "We highly recommend [deploying the RTCD service](https://docs.mattermost.com/configure/calls-deployment.html#the-rtcd-service) to offload calls processing to a separate instance in order to maintain the performance, scalability and reliability of your main Mattermost server."
```
cwarnermm, line 4:
```suggestion
    "translation": "We highly recommend [deploying the RTCD service](https://docs.mattermost.com/configure/calls-deployment.html#the-rtcd-service) to offload calls processing to a separate instance in order to maintain the performance, scalability, and reliability of your main Mattermost server."
```

server/i18n/en.json
```diff
@@ -1,4 +1,20 @@
...
+  },
+  {
+    "id": "app.admin.concurrent_sessions_warning.intro",
+    "translation": "Your installation is experiencing heavy usage from Mattermost Calls that can negatively impact the performance and responsiveness of the system."
```
cwarnermm, line 8:
```suggestion
    "translation": "Your installation is experiencing heavy usage from Mattermost Calls that can negatively impact system performance and responsiveness."
```

server/i18n/en.json
```diff
@@ -1,4 +1,20 @@
...
+  },
+  {
+    "id": "app.admin.concurrent_sessions_warning.pro_or_e0",
+    "translation": "We highly recommend upgrading to [Enterprise](https://mattermost.com/pricing/) and [deploying the RTCD service](https://docs.mattermost.com/configure/calls-deployment.html#the-rtcd-service) to offload calls processing to a separate instance in order to maintain the performance, scalability and reliability of your main Mattermost server."
```
cwarnermm, line 12:
```suggestion
    "translation": "We highly recommend upgrading to [Enterprise](https://mattermost.com/pricing/) and [deploying the RTCD service](https://docs.mattermost.com/configure/calls-deployment.html#the-rtcd-service) to offload calls processing to a separate instance in order to maintain the performance, scalability, and reliability of your main Mattermost server."
```

cwarnermm (commented):
Provided editorial feedback inline and details on how to request PL links for in-product docs references.

server/limits_test.go
```diff
@@ -0,0 +1,313 @@
...
+				Id:     "adminID",
+				Locale: "it",
+			},
+		}, nil).Once()
+
+		mockAPI.On("GetDirectChannel", "adminID", "botID").Return(&model.Channel{
+			Id: "channelID",
+		}, nil).Once()
+
+		mockAPI.On("CreatePost", &model.Post{
```
cpoile, line 264: I wonder if it would be good to use `mockAPI.AssertNumberOfCalls...` for these, since that's what you're testing, right?

server/telemetry.go
```diff
@@ -8,15 +8,35 @@ import (
...
+	// professionalSKUs = []string{model.LicenseShortSkuProfessional, model.LicenseShortSkuEnterprise}
+
+	// We only need to map events that require a SKU (i.e., licensed features). Anything available on unlicensed
+	// servers will map to null as expected.
+	eventToSkusMap = map[string][]string{
+		"user_start_recording": enterpriseSKUs,
+		"user_stop_recording":  enterpriseSKUs,
+		"live_captions_on":     enterpriseSKUs,
+		"live_captions_off":    enterpriseSKUs,
+	}
```
cpoile, line 39: telemetry PR?

server/websocket_test.go
```diff
@@ -960,4 +961,121 @@ func TestHandleJoin(t *testing.T) {
...
+		// Call unlock
+		mockAPI.On("KVDelete", "mutex_call_"+channelID).Return(nil).Once()
+
+		err := p.handleJoin(userID, connID, authSessionID, callsJoinData{
+			CallsClientJoinData: CallsClientJoinData{
+				ChannelID: channelID,
+			},
+		})
+		require.NoError(t, err)
+
```
cpoile, line 1075: wow... that's a lot of mocks. how long did that take?

server/websocket_test.go
```diff
@@ -960,4 +961,121 @@ func TestHandleJoin(t *testing.T) {
...
+		// We'd be starting a new call
+		mockMetrics.On("IncWebSocketEvent", "out", wsEventCallHostChanged).Once()
+		mockAPI.On("PublishWebSocketEvent", wsEventCallHostChanged, mock.Anything,
+			&model.WebsocketBroadcast{UserId: userID, ChannelId: channelID, ReliableClusterSend: true}).Once()
+
+		// Call started post creation
+		mockAPI.On("GetUser", userID).Return(&model.User{Id: userID}, nil).Once()
+		mockAPI.On("GetConfig").Return(&model.Config{}, nil).Times(3)
+		mockAPI.On("CreatePost", mock.AnythingOfType("*model.Post")).Return(&model.Post{Id: postID}, nil).Once()
```
cpoile, line 997: Whichever one is the admin warning, that's what you'd want to run an assert on.

cpoile (commented):
Some thoughts to improve the testing?

server/telemetry.go
streamer45: Likely got it as the parent PR got merged. Will rebase.

server/limits_test.go
streamer45: @cpoile I'm not sure I fully understand what you mean by *these*.

In general, I try to be as specific as possible so we can verify the inputs and outputs of our mocked methods rather than just whether they were called.

server/websocket_test.go
streamer45: A bit. The problem is that after implementing this, I realized the single subcases were dependent on the order, so they would fail if executed independently (e.g., `TestHandleJoin/admin_warning`); this is because there isn't an easy way I could find to reset mock registrations between new test cases without re-initializing them completely.

We could, in theory, try to extract all the common ones and put them above before any case starts, but then it would probably be even less clear what's happening.

server/websocket_test.go
streamer45: @cpoile Is the assertion just a way to avoid the panic, or is there more to it?

server/limits_test.go
cpoile: What I mean is: keep what you have, and then for the API calls that we are assuming are being called (esp. the `CreatePost`), do a `mockAPI.AssertNumberOfCalls(t, "CreatePost", 1)` to make sure. That's the real test, right?

server/websocket_test.go
cpoile: It's what I clarified above -- asserting that we're actually doing the CreatePost.

server/limits_test.go
streamer45: @cpoile Ah. Are you saying that if `CreatePost` doesn't execute for some reason, it won't complain?

cpoile: > @cpoile Ah. Are you saying that if `CreatePost` doesn't execute for some reason, it won't complain?

Yah, iirc once just says mock this once, but doesn't complain if it doesn't get called at all.

streamer45: I'll fix the ones touched in here, but then I'll need to send a separate PR to review the rest.

streamer45:
Note: I'll review permanent links in a separate PR since they need more than just replacing (a custom wrapping component to add some query parameters). We should also review pre-existing links. Created https://mattermost.atlassian.net/browse/MM-59257

DHaussermann:
@streamer45 I think I'm understating how to test this but, I seem to have missed something because I don't see the warning DM from the bot when I hit the warn limit.

Steps:
- Running `master` branch of Mattermost
- Change 2 lines in `limits.go` as follows 
```
concurrentSessionsThresholdDefault          = 2
concurrentSessionsWarningBackoffTimeDefault = 1 * 1 * time.Hour // 1 week
```
- Build and deploy the plugin with these lines changed
- Remove license file from the server
- Start a call with 3 Participants 
**Expected**: The Calls bot should DM the admin user when I hit or crossed the concurrent sessions threshold  
**Observed** No DM from the bot
image

Should I be seeing the DM?

Couple other questions:
- Was it your intention to have me change the values on lines `30` and `31`? Maybe this was parameterized somewhere and I missed it.   
- Is there a server side PR needed here. I'm actually confused because on Mattermost `master` branch on my local I'm not seeing the new metrics for Calls under _Site Statistics_ which should have been merged a while ago. Could this be related?   
image

streamer45:
@DHaussermann 

> Was it your intention to have me change the values on lines 30 and 31? Maybe this was parameterized somewhere and I missed it.

Not exactly. I implemented environment overrides to make this simpler. See https://github.com/mattermost/mattermost-plugin-calls/pull/802#discussion_r1654469588. But it shouldn't make a difference.

> Is there a server side PR needed here. I'm actually confused because on Mattermost master branch on my local I'm not seeing the new metrics for Calls under Site Statistics which should have been merged a while ago. Could this be related?

No MM side changes are needed for this one.



Make sure you are not using RTCD and restart the plugin when you make the config change. This warning is meant for installations not using RTCD which is the first requirement to trigger any of this logic.

Also, you can always run the following DB query to clear the backoff in case you want to repeat the test quickly:

```sql
DELETE FROM pluginkeyvaluestore WHERE pluginid = 'com.mattermost.calls' AND pkey = 'concurrent_sessions_warning';
```

DHaussermann:
@streamer45 thanks for adding details. 

I was not sure how to use the override. I tried using it as an environment variable `make  run-server MM_CALLS_CONCURRENT_SESSIONS_THRESHOLD=2` but when I did not get the warning DM I assumed I misunderstood so I changed the hard coded values instead. However, since it still not working, something else must be going on. 

- I ensured I'm not using RTC. All config values for this are empty
- I ran the SQL statement you provided but id did not update any rows
- I'm still curious why I'm not seeing any calls metrics in the system console and wonder if this could be the cause. For fun, I switched my branch from `master` to `release-9.9` but I still don'r see this data.

Let me know if you have any other thoughts.

DHaussermann:
Tested and passed
- Confirmed the warning works as expected 
- Confirmed the warning message is accurate based on license. Tested Enterprise, Professional, E0(no license) and Team(no Enterprise repo included)
- Also confirmed when RTCD is already in use with the plugin, exceeding warn limit has no affect.
- Confirmed Bot DMs to happen during the cool off period
- Confirmed all sysadmin role users get the DM

LGTM!

Thank @streamer45 for helping me debug my webapp issue locally ☝️  so I could test this!

DHaussermann (approved):

PR #805, opened 2024-07-02
[MM-53118] Dynamically sized participants grid

streamer45:
#### Summary

PR takes a first pass at implementing a dynamically sized call participants grid in the pop-out window view.

#### Specs

https://mattermost.atlassian.net/wiki/spaces/Calls/pages/2707193901/Calls+Participant+Grid
https://www.figma.com/design/8kcq0HRN2TUaVAdPeeSvsb/Calls-Participants-Grid?node-id=1-7&m=dev

#### Video

https://github.com/mattermost/mattermost-plugin-calls/assets/1832946/e6a5dec4-8068-4cea-8720-ac5059de579c

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-53118


webapp/src/websocket_handlers.ts
```diff
@@ -185,7 +185,7 @@ export function handleUserJoined(store: Store, ev: WebSocketMessage<UserJoinedDa
     }
 
     if (ringingEnabled(store.getState()) && userID === currentUserID) {
-        const callID = calls(store.getState())[channelID].ID || '';
+        const callID = calls(store.getState())[channelID]?.ID || '';
```
streamer45, line 188: It's unrelated, but I spotted a crash, so I'm fixing it here.

cpoile: Yikes, thanks.

abhijit-singh:
@streamer45 The screen cap looks great. Could you please put this on a test server?

streamer45:
> @streamer45 The screen cap looks great. Could you please put this on a test server?

@abhijit-singh Done.

abhijit-singh (changes requested):
Thanks @streamer45! Looks like a huge improvement to me!

Just a couple of requests:
1. If the username is displayed for a participant and it is long, it does not seem to be getting truncated properly. The full name does get truncated as expected even if it is long. 
image

2. Probably just a nice-to-have, but can we center align the new row of participants that's not fully populated. Here's screenshots to explain better - 

Current: Left aligned bottom row (blank space on the right of participants)
image

Ideal: Center aligned bottom row (blank space on the left and right)
image

streamer45:
@abhijit-singh I fear that neither is going to be straightforward:

1. We are already using a [non-stardard CSS feature](https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-line-clamp) to make it work properly on two lines. The easiest thing to do without adding more dynamic logic would be to simply break a very long string (e.g., the username you tested there) to the next line. Then, on two lines, the ellipsis would take care of it if it still exceeds the available space. If, instead, we want a per-line ellipsis, then it will be more tricky.
2. That's how CSS grid is designed to work. To achieve centering of the last row, we'd have to go back to flex and see how it looks. I can check, but no promises.

abhijit-singh:
> The easiest thing to do without adding more dynamic logic would be to simply break a very long string

I'd be good with that. It is an edge case but how it renders currently clearly looks broken. 

>  I can check, but no promises.

Not super confident if it adds value so I won't want you to spend too much time here. If it can be done without too much rework or breaking other things then that's awesome. Else we can choose to let it stay as is and review it again once we've used it for a bit.

streamer45:
@abhijit-singh I haven't committed here yet but I updated the test server to see how it looks with flex. One side effect of that I notice is that now the grid items are not forced to be of equal height. It may also take a bit to ensure there's no funny breakage since we've been using grid for a long time. Let me know what you think, otherwise we can revert this bit.

abhijit-singh (approved):
> now the grid items are not forced to be of equal height.

Looks to me like they're still of equal heights when I tested so I'm not sure of the impact. 

---
And seems perfect to me now. We can go ahead with this from a UX perspective. 

Maybe we can take help from someone in QA to do comprehensive testing on this if we fear things might break?

streamer45:
> Looks to me like they're still of equal heights when I tested so I'm not sure of the impact.

@abhijit-singh It's more noticeable when you hover as host.

https://github.com/mattermost/mattermost-plugin-calls/assets/1832946/688a0cdf-7bbc-40c6-a383-ab9f4516beba

> Maybe we can take help from someone in QA to do comprehensive testing on this if we fear things might break?

Yeah, we'll do some extra checking.

abhijit-singh:
Aah got it. I think that's okay and expected. 

I couldn't see it earlier since I had a host in both rows :P

image

streamer45:
> Aah got it. I think that's okay and expected.
> 
> I couldn't see it earlier since I had a host in both rows :P
> 
> image

Cool :+1:

streamer45:
@DHaussermann For this one, we'd like to check for any funny UX behaviour in the Calls pop-out based on the number of participants, their state, and different window sizes.

DHaussermann (approved):
Tested and passed.
- Calls with several dozen participants uses horizontal space more efficiently
- Participant profile are show in wider rows that span further towards the right and left edges as expected
- Spot checked several odd and event participant counts raging form only a few up to over 60 
- As participant list grows the width of the rows grow gradually and the size of the the profile images gradually shrinks as expected  
- My test rig could not handle pushing the count up 75 - 100 but seems like the grid will scroll in the available horizontal resolution now rather than page.
- Tested in browser and desktop as a precauton

LGTM!

PR #807, opened 2024-07-04
[MM-58292] Suppress push notifications to call thread if receiver is in call

streamer45:
#### Summary

PR adds some logic to suppress mobile push notifications for replies to a call thread if the receiver is in that call. This is to mimic (see note) the behavior of the web/desktop [client-side hook](https://github.com/mattermost/mattermost-plugin-calls/blob/29644adc8548c9fabccd37706209f54565547a81/webapp/src/desktop_notifications.ts#L14).

Note: I haven't found an easy way to check for mentions, so we'd be disabling any type of notification, which I find reasonable for the call thread, given the user should have the app open somewhere.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-58292

### DIFF
--- a/server/db/calls_sessions_store.go
+++ b/server/db/calls_sessions_store.go
@@ -215,3 +215,34 @@ func (s *Store) GetCallSessionsCount(callID string, opts GetCallSessionOpts) (in
 
 	return count, nil
 }
+
+func (s *Store) IsUserInCall(userID, callID string, opts GetCallSessionOpts) (bool, error) {
+	s.metrics.IncStoreOp("IsUserInCall")
+	defer func(start time.Time) {
+		s.metrics.ObserveStoreMethodsTime("IsUserInCall", time.Since(start).Seconds())
+	}(time.Now())
+
+	qb := getQueryBuilder(s.driverName).Select("1").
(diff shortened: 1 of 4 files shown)


server/db/calls_sessions_store.go
```diff
@@ -215,3 +215,34 @@ func (s *Store) GetCallSessionsCount(callID string, opts GetCallSessionOpts) (in
 
 	return count, nil
 }
+
+func (s *Store) IsUserInCall(userID, callID string, opts GetCallSessionOpts) (bool, error) {
```
cpoile, line 219: What do you think of the performance impact of this approach? I guess we don't really have any other option, and it's not an event that will happen too often, but just curious.

streamer45: @cpoile I'd expect this to be extremely efficient, given that `callID` is a primary key and the sessions table is very small. Plus the query doesn't need to return any structured data. The only further optimization would be an index on `calls_sessions.userid`, but given the premise (small table + filter on call), it's not worth it since a scan will be the best option anyway. To put it in comparison, I think it could be more costly the RPC hook call itself than the DB query.

PR #815, opened 2024-07-17
[MM-57418] System console config page review

streamer45:
#### Summary

PR does a complete review of Calls config UI in the system console:

- Grouped related settings into sections.
- Full i18n support. All the settings can now be translated.
- E2E tests

#### Design

https://www.figma.com/design/feH2AjJIzIkgRATJSxYyH6/Calls-System-Console-Config?node-id=2032-7981&m=dev

#### Screenshot

image

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-57418
https://mattermost.atlassian.net/browse/MM-58160
https://mattermost.atlassian.net/browse/MM-59364


streamer45:
@cwarnermm Everything for you to review should be in the `en.json` file -> https://github.com/mattermost/mattermost-plugin-calls/pull/815/files#diff-bd9c8eded3f2f6d40191d91defa64d238b56387d4f61e3bdb75140d02cb1bb80

webapp/i18n/en.json
```diff
@@ -134,38 +183,53 @@
...
+  "ePVyxv": "Test mode",
   "eisS9Z": "Call ended",
   "elmIJt": "Avg Call Participants",
   "esxtQH": "Call from {callerName} with {others}",
   "euPd26": "Do you want to leave and join a call with {user}?",
+  "evecl7": "Azure API Key",
   "f7Iz8V": "Allow screen recording access to Mattermost.",
+  "fBafHF": "ICE and TURN",
   "fLgmQZ": "Unable to stop recording",
+  "fl+8IT": "(Optional) When set to on it will pass and use configured TURN candidates to server initiated connections.",
```
cwarnermm, line 208:
```suggestion
  "fl+8IT": "(Optional) When enabled, it will pass and use the configured TURN candidates to server initiated connections.",
```

webapp/i18n/en.json
```diff
@@ -178,35 +242,51 @@
...
+  "omP/e4": "Use your own WebRTC service",
+  "ovJ26C": "Medium",
   "p/C72L": "Total Active Calls",
   "p0+4B8": "<b>You're unmuted.</b> Select {muteIcon} to mute.",
   "p7D2e3": "In this channel",
   "paBpxN": "Ignore",
+  "pkW7OA": "The number of threads per live-captions transcriber. The product of LiveCaptionsNumTranscribers * LiveCaptionsNumThreadsPerTranscriber must be in the range [1, numCPUs].",
+  "q/D7UA": "Configure a dedicated service used to offload calls and efficiently support scalable and secure deployments",
   "qR+5t1": "is talking…",
+  "r39PWQ": "When set to true it enables simulcast for screen sharing. This can help to improve screen sharing quality.",
```
cwarnermm, line 264:
```suggestion
  "r39PWQ": "When set to true, simulcast for screen sharing is enabled. This can help to improve screen sharing quality.",
```

webapp/i18n/en.json
```diff
@@ -217,7 +297,10 @@
   "xYhChH": "Are you sure you want to remove <b>{userName}</b> from the call?",
   "yL9vD/": "Welcome to your Mattermost Enterprise trial! It expires on {trialExpirationDate}. You now have access to <rtcdDocsLink>RTCD services</rtcdDocsLink>, <recordingsDocsLink>call recordings</recordingsDocsLink>, <guestAccountsLink>guest accounts</guestAccountsLink>, <autoComplianceReportsLink>automated compliance reports</autoComplianceReportsLink>, and <mobileSecureNotificationsLink>mobile secure-ID push notifications</mobileSecureNotificationsLink>, among many other features. View all features in our <documentationLink>documentation</documentationLink>.",
   "yO68rk": "Calls are currently running in test mode and only system admins can start them. Reach out directly to your system admin for assistance",
+  "yfYHc5": "ICE Host Override",
   "yjQNKt": "Calls are disabled in this channel.",
   "zUH89x": "The call recording will be processed and posted in the call thread. Are you sure you want to stop the recording?",
-  "zx0myy": "Participants"
+  "zmjvyC": "Settings for participants, screen sharing, ringing and more",
```
cwarnermm, line 303:
```suggestion
  "zmjvyC": "Settings for participants, screen sharing, ringing, and more",
```

cwarnermm (commented):
Provided editorial feedback on in-product strings inline.

cpoile (approved):
This is nice, I like how it handles backwards compatibility. Great work!

e2e/utils.ts
```diff
@@ -105,3 +105,72 @@ export async function joinCallAndPopout(userState: string) {
 export function getChannelURL(channelName: string) {
     return `${baseURL}/${defaultTeam}/channels/${channelName}`;
 }
+
+export async function resizeAndScreenshot(page: Page, testID: string) {
```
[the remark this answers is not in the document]
streamer45: @cpoile This is needed to cover potential UX regressions on any of our fields. The reason for this function is that most of them are hidden as they need scrolling to be visible and Playwright doesn't support this very well. It could happen that a single field type is somehow bugged without showing issues in the rest of the section. This could be seen as "lazy" solution to avoid adding a single test for each and every input we support.

streamer45:
@abhijit-singh Thanks. Are the above blockers or nice to have? Unfortunately, the first two items won't be possible to address on this side but will need reworking how the webapp renders custom plugin sections (an addendum to https://github.com/mattermost/mattermost/pull/27654 essentially).

Could we get the Figma files updated as first step to cover for these cases? Then I'll be in a better position to estimate how much work is still required.

streamer45:
@abhijit-singh Thanks for updating the design. There's one remaining complexity I hope we can find a compromise for.

It's particularly hard to intermix settings and sections and "Enable/Disable Plugin" is a bit of a special setting. What we can do is have an additional section always on top with just that setting (see below examples). It's not exactly like if the setting was on its own but I think it's the best I can come up with without substantially reworking the general system console rendering logic which would be fairly outside the scope for this work. Let me know.


**Enabled case**

image

**Disabled case**

image

abhijit-singh (approved):
@streamer45 Thanks for making these updates! It is definitely better and I can live with it.

Just wanted to mention that we were able to achieve the same thing with the [User Survey Plugin](https://github.com/mattermost/mattermost-plugin-user-survey) that @harshilsharma63 recently developed. It also has a Enable/Disable config at the top and everything else below it was in sections. Maybe we can take pointers from there? 

A last resort could also be just styling the top section differently? Removing the white background and tweaking some spacing on the left and top should do, but again if it is a non-trivial amount of effort, I'm okay to skip this.

streamer45:
@abhijit-singh That's not using the plugin config schema as it's meant to be. The AI plugin is another example of that pattern. They are wrapping the whole config object into a single "custom" setting so they can have full control over everything. It's hacky, but it works for them. In the case of Calls it would be a huge breaking change to our config as our settings are well defined on the top level object. See the difference:

https://github.com/mattermost/mattermost-plugin-calls/blob/dcdbab2f20af437dce26c8f6cb29cf2a07cee0a4/plugin.json#L23

https://github.com/mattermost/mattermost-plugin-user-survey/blob/2bd1ae2f4a64eb1397cc1f182071138fa8c871db/plugin.json#L23-L28

https://github.com/mattermost/mattermost-plugin-ai/blob/9f4404642937f39a89df223b732d36f46cbcfd65/plugin.json#L25-L29

Here I have been trying to fix this the hard way, going back to the webapp side and implementing the missing functionality. Plugins shouldn't be able to override their own Enable/Disable setting in my mind. 

> A last resort could also be just styling the top section differently? Removing the white background and tweaking some spacing on the left and top should do, but again if it is a non-trivial amount of effort, I'm okay to skip this.

We could do that, just hack the CSS a little since it's a bit of a special section. Not sure it'll look necessarily nicer but I guess consistency wise it would be a superior solution.

abhijit-singh:
Thanks @streamer45, with the changes in the webapp PR, the page looks great!

Just one thing that I had mentioned earlier but maybe got missed - 
- Minor UI tweak - can we change the font of the section titles to Metropolis to match the design? Currently it is Open Sans.
Current: 
image

Ideal:
image

streamer45:
> Thanks @streamer45, with the changes in the webapp PR, the page looks great!
> 
> Just one thing that I had mentioned earlier but maybe got missed -
> 
> * Minor UI tweak - can we change the font of the section titles to Metropolis to match the design? Currently it is Open Sans.

Thanks, will send an update.

DHaussermann:
@streamer45 this is looking good. I have couple follow-up questions.
1. Just do double check, you intentionally removed the `(Beta)` label from the calls ringing feature correct? 
2. Regarding i8n, it seem I can't confirm the `true` / `false` labels for the bools can be translated. 
The top one is translated but that's likely the webapp translating and not the plugin.
image
I do see other labels AND values being translated so, maybe this is not a concern.
image
Is there any risk here at all or just no translations for the bools?

streamer45:
@DHaussermann 

1. Yep, that commit got merged through a separate PR (https://github.com/mattermost/mattermost-plugin-calls/pull/834).
2. Thanks for checking. That looks expected since we just introduced i18n support so user translated strings are missing. The reason some are translated is because they were already custom components, prior to this PR. For example, the `Server Side TURN` setting is not using the True/False labels but rather On/Off. As for the very first setting (on top) that's not rendered by the plugin so it's core webapp which has a different set of translations available.

For consistency, I have sent an update to convert the labels for `Test Mode` to On/Off since it's more accurate.

DHaussermann:
Thanks @streamer45 

I see the change you implemented to ensure the boolean can be translated.

image

DHaussermann (approved):
Tested and Passed
- Confirmed settings page had been reorganized as per proposed layout.  
- Beta and Experimental labels have been removed  
- I8n is functional for the plugin

LGTM!

PR #822, opened 2024-07-29
Remove CGO dependency for load-test main binary

streamer45:
#### Summary

PR removes the CGO dependency from our main load-test command (i.e. `./lt/cmd/lt`) since it has caused a bit of pain for our customers as it requires some extra system dependencies which are not always straightforward to install. Given the opus encoder is only used by the `speech` command, we can extract it over there through the use of a common interface.


cpoile (approved):
That's clever, nice solution. :)

PR #825, opened 2024-07-29
Add AV1 support to load-test client

streamer45:
#### Summary

PR adds support to our load-test client for sending AV1-encoded screen tracks when the flag is enabled. Similarly to what happens on a browser, in such case we also send a fallback VP8 track.


cpoile (approved):
Looks good, thanks!

PR #827, opened 2024-07-30
Fix dates flakiness in calls stats

streamer45:
#### Summary

Working with time is hard enough. Working with dates is just brutal.

PR tries to fix some flakiness caught in CI tests on this day (Jul 30th). The way `time.AddDate` works made it possible to get duplicate months due to normalization. Now we are ensuring to always use the first day of the month rather than the current one which should guarantee no duplicates are added (added a test).


isacikgoz (approved):
Wow who could've think of this beforehand 🤯

PR #828, opened 2024-08-01
Fix flaky test

streamer45:
#### Summary

Working with dates is a bit tricky. Test could fail based on which day of the month we were on.


streamer45:
It's one of those weeks...pipeline is broken after some update it seems.

cpoile (approved):
😬 Thanks!

PR #831, opened 2024-08-05
[MM-59070] Add telemetry events for host controls

streamer45:
#### Summary

Adding some telemetry tracking for the recently added host control feature.

@ifoukarakis We are not sending any additional application-specific data other than the event because I am not sure it's needed (e.g., CallID). Of course, we are including the correct SKUs as agreed before.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-59070

### DIFF
--- a/server/host_controls.go
+++ b/server/host_controls.go
@@ -69,6 +69,8 @@ func (p *Plugin) changeHost(requesterID, channelID, newHostID string) error {
 		UserIDs:             getUserIDsFromSessions(state.sessions),
 	})
 
+	p.track(evHostChangeHost, nil)
+
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


cpoile (approved):
🙇 Thank you 🙇

ifoukarakis (approved):

ifoukarakis:
> We are not sending any additional application-specific data other than the event because I am not sure it's needed (e.g., CallID). Of course, we are including the correct SKUs as agreed before.

These can be added later if we have requirements for tracking specific flows. The benefit of keeping it minimal is that we don't overload property names unless needed. 

CC @catalintomai

PR #832, opened 2024-08-05
[MM-59056] Improve label for channel types in admin stats

streamer45:
#### Summary

Since labels are expected to be plain strings, we need to localize them outside of a react component's context. 
Slightly hacky, but it should work as webapp is doing [more or less the same](https://github.com/mattermost/mattermost/blob/f2907454960e9f05f5b3c3f3cb2864e71932cff2/webapp/channels/src/components/analytics/format.tsx#L9-L10). 

#### Screenshot

*Before*

image

*After*

image

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-59056


cpoile (approved):
Sounds great, thanks!

PR #833, opened 2024-08-06
[MM-60010] Fix recording state cleared when session for same participant leaves

streamer45:
#### Summary

Found an interesting edge case when testing the new recorder image. If a user is connected more than once to the same call (multi sessions), then any of their sessions leaving would cause the recording state to be reset for all of them.

I tried to fix this in a couple of different ways (i.e., persisting the previous call sessionID or just avoiding clearing the state), but they all come with some tradeoffs, so I eventually decided that having a more explicit redux action to track the local client closing may be the cleanest option.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-60010


cpoile (approved):
Heh, that's a fun little bug, nice work finding it. :)

PR #840, opened 2024-08-07
[MM-59996] Make stats and client logs more easily accessible through slash commands

streamer45:
#### Summary

PR fixes an issue with the `/call stats` command not working on the Desktop app. This bug has existed since we introduced the global widget, which is a different window and requires special handling in order to communicate any data back to the center channel (main app).

Here, we are using a *trick* that doesn't require touching the Desktop. We use the `localStorage`, which is shared between all the windows. This is a temporary solution while we work on a better path forward ([MM-60028](https://mattermost.atlassian.net/browse/MM-60028)).

While we are here, I also wanted to address another significant limitation concerning Call logs. On Desktop, these would disappear along with the widget window as soon as the call dropped, making it almost impossible to debug network issues.

To address this, PR implements a convenient storage-backed `/call logs` command that returns the client logs for the last call session without requiring users to go through multiple hops (i.e., the console) in order to access them (see screenshot).

#### Screenshot 

calls_logs

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-59996
https://mattermost.atlassian.net/browse/MM-59995


webapp/src/client.ts
```diff
@@ -484,7 +490,8 @@ export default class CallsClient extends EventEmitter {
         this.closed = true;
         if (this.peer) {
             this.getStats().then((stats) => {
-                sessionStorage.setItem('calls_client_stats', JSON.stringify(stats));
+                const storage = window.desktop ? localStorage : sessionStorage;
+                storage.setItem(STORAGE_CALLS_CLIENT_STATS_KEY, JSON.stringify(stats));
```
streamer45, line 494: I didn't want to overcomplicate it by adding listeners, so the stats are saved when the client disconnects (user leaves call) and can be retrieved after that. 

This slightly differs from Web browser behavior, where stats can be fetched at any time during or after a call, but the simple fix is enough to cover the customer's use case (troubleshooting).

server/slash_command.go
```diff
@@ -245,6 +264,17 @@ func (p *Plugin) ExecuteCommand(_ *plugin.Context, args *model.CommandArgs) (*mo
 		return resp, nil
 	}
 
+	if subCmd == logsCommandTrigger {
```
enzowritescode, line 267: This if subCmd == subCommandCXyz { ... } format in this function is getting kind of long.
Either in this PR or another can we shorten it?

Maybe something like the following (or a switch):

if subCmd == subCommandXyz {
	return subCommandXyzHelper()
}

webapp/src/client.ts
```diff
@@ -484,7 +490,8 @@ export default class CallsClient extends EventEmitter {
         this.closed = true;
         if (this.peer) {
             this.getStats().then((stats) => {
-                sessionStorage.setItem('calls_client_stats', JSON.stringify(stats));
+                const storage = window.desktop ? localStorage : sessionStorage;
```
enzowritescode, line 493: This line is repeated in many places. Perhaps a helper function.

enzowritescode (changes requested):

server/slash_command.go
cpoile: On the other hand, having a long function that reads simply top-to-bottom can be easier and quicker to read and understand, compared to having to jump back and forth to/from functions. So there's an argument for keeping it simple, too.  :)

webapp/src/client.ts
```diff
@@ -99,8 +104,8 @@ export default class CallsClient extends EventEmitter {
             };
         }
 
-        const defaultInputID = window.localStorage.getItem('calls_default_audio_input');
-        const defaultOutputID = window.localStorage.getItem('calls_default_audio_output');
+        const defaultInputID = window.localStorage.getItem(STORAGE_CALLS_DEFAULT_AUDIO_INPUT_KEY);
```
cpoile, line 107: 👍

webapp/src/log.ts
```diff
@@ -1,20 +1,54 @@
 /* eslint-disable no-console */
 
+import {STORAGE_CALLS_CLIENT_LOGS_KEY} from 'src/constants';
+
 import {pluginId} from './manifest';
 
+let clientLogs = '';
```
cpoile, line 7: ❤️ js

cpoile (approved):
Thanks! Nice simple solution to a tough problem.

server/slash_command.go
enzowritescode: @cpoile single responsibility principle + easier to write tests for / understand the code coverage, especially if you use an IDE tool that shows highlights for code coverage.

I'm fine with leaving it as is if nobody else thinks it's an issue, but it's definitely a code smell.

enzowritescode (approved):

server/slash_command.go
enzowritescode: Another option is I can make this change if you all want in another ticket.

PR #845, opened 2024-08-14
Calls v1 prep

streamer45:
#### Summary

We are making MM v10 a requirement for running Calls v1 since it makes it more explicit we are introducing breaking changes (namely Calls being available only in DMs for unlicensed servers).

Also bumping some Calls dependencies in the process as we are wrapping up the release.


isacikgoz (approved):
LGTM

Tampa:
And here we go, the downfall has begun. No calls in public channels anymore, wtf is this shit. Excuse my french, but that's killing the one thing that made the call function useful!

PR #846, opened 2024-08-15
Utilize reporting script from default branch

mvitale1989:
#### Summary

This PR utilizes the `main` branch version of the repository when producing the E2E test report.

The ref to check out doesn't need to be specified, because for `workflow_run` it's the default branch out-of-the-box:
- https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#workflow_run
- https://github.com/actions/checkout?tab=readme-ov-file#checkout-v4


streamer45 (approved):
Thanks :+1:

saturninoabril (approved):

PR #847, opened 2024-08-15
[MM-60199] Fix race condition in case of concurrent leave/join in the same channel

streamer45:
#### Summary

When working on unrelated `rtcd` changes, I noticed a couple of flaky tests. At times, some sessions would fail to connect due to timeout without a clear explanation.

After a good amount of investigation, I was able to track this down to the batching behaviour not working as expected in case of concurrent join/leave operations on the same channel. In fact, it would be possible to stop batchers that still had some queued items (e.g., sessions trying to join). This resulted in those sessions to timeout as they wouldn't ever be handled.

We fix this by simply checking (under lock) for emptiness before stopping the batches.

Writing a test for this wasn't easy but I managed to reproduce it several times, although it's not 100% consistent due to the specific timings involved.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-60199

### DIFF
--- a/server/batching/batcher.go
+++ b/server/batching/batcher.go
@@ -111,3 +111,8 @@ func (b *Batcher) Stop() {
 	close(b.stopCh)
 	<-b.doneCh
 }
+
+// Empty returns whether or not the are more items left in the batcher's queue.
+func (b *Batcher) Empty() bool {
+	return len(b.itemsCh) == 0
+}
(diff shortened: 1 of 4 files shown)


cpoile (approved):
Ok, you win the "most complicated test for a one line change" award. :trophy: 
Nice work!

PR #851, opened 2024-08-21
[MM-60252] Fix failure to negotiate tracks after ws reconnect

streamer45:
#### Summary

In https://github.com/mattermost/mattermost-plugin-calls/pull/821/files#diff-d42de9238c805b64e9ba33d375fc6a246ff6e08feeecb4f1a0f54285cc7e4261 I inadvertently introduced a regression in case of websocket reconnect to a different HA node. 

Since this case causes a new connection to be created, we cannot use the original connection ID to forward subsequent messages. This worked fine before because we'd be broadcasting them to all connections belonging to the user but that's not ideal, so the solution here is to find the right connection ID that maps to the original one.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-60252

### DIFF
--- a/e2e/tests/media.spec.ts
+++ b/e2e/tests/media.spec.ts
@@ -306,4 +306,53 @@ test.describe('sending voice', () => {
         await devPage.leaveCall();
         await userPage.leaveCall();
     });
+
+    test('unmuting after ws reconnect', async ({page}) => {
+        const userPage = await startCall(userStorages[1]);
+
+        const devPage = new PlaywrightDevPage(page);
+        await devPage.joinCall();
+
+        const reconnectHandler = () => {
+            return new Promise((resolve) => {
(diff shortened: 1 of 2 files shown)


agnivade (approved):
This only works in non-HA right? It's not possible to get the original connection ID from a different node.

streamer45:
> This only works in non-HA right? It's not possible to get the original connection ID from a different node.

@agnivade The fix is specifically for HA. The original connection ID is provided by the Calls side as we track it throughout the lifetime of the WebRTC connection.

DHaussermann (approved):
Tested and passed
- Using the steps provided I observed that the issue reproduced 3/5 times
- Updated the plugin to this PR branch and restarted resting
- Tested over 10+ attempts and could not observe the track issue any longer
- As a precaution - Also explored and regression tested cases where users needed the WS reconnect during the course of the call
No issues found.

LGTM!

Thanks for the steps on this @streamer45. It was really cool to be able to reproduce this somewhat consistently!

PR #867, opened 2024-09-25
Fix e2e pipeline flakyness

streamer45:
#### Summary

Printing calls-recorder/transcriber logs have been often failing as of late with `Error response from daemon: can not get logs from container which is dead or marked for removal`. 

We only really care about exited containers, which are retained past their execution. Ideally we'd filter by exit code but the `docker ps` filter functionality doesn't seem to support advanced logic such as `exited!=0`.

### DIFF
--- a/e2e/scripts/run.sh
+++ b/e2e/scripts/run.sh
@@ -58,13 +58,13 @@ docker ps -a
 docker logs "${COMPOSE_PROJECT_NAME}_callsoffloader"
 
 # Print transcriber job logs in case of failure.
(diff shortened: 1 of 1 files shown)


cpoile (approved):
That's pretty advanced logic.
I'm sure you've checked if there's another way that it does support?

streamer45:
> That's pretty advanced logic. I'm sure you've checked if there's another way that it does support?

@cpoile You can filter by exit code but what we need is anything but 0 (success). With some advanced scripting we could compare the two lists (all containers and those exited with 0) and find the difference, but that's overkill. If we get more failures I'll simply get rid of the logs since they are only used to debug.

PR #870, opened 2024-10-02
DC messages

streamer45:
#### Summary

Not much of a change on this side. I removed the `pako` dependency since we now use `fflate`. Also reworked logging and updated dependencies.

#### Related PRs

https://github.com/mattermost/rtcd/pull/158
https://github.com/mattermost/calls-common/pull/40


lt/client/client.go
```diff
@@ -468,6 +468,10 @@ func (u *User) Speak(text string) chan struct{} {
 }
 
 func (u *User) onConnect() {
+	// Waiting a second before transmitting to avoid
+	// the known race between incoming and outgoing track.
+	// This is needed as pion doesn't support perfect negotiation yet.
+	time.Sleep(time.Second)
```
streamer45, line 474: Using DC further exacerbated the impact of signaling conflicts, so I had to add this to make it usable. Browsers are not affected as they support perfect negotiation (SDP rollback).

cpoile: Gotcha. :+1:

webapp/src/client.ts
```diff
@@ -5,8 +5,7 @@ import type {EmojiData, CallsClientJoinData} from '@mattermost/calls-common/lib/
 
 import {EventEmitter} from 'events';
 
-// @ts-ignore
-import {deflate} from 'pako/lib/deflate';
+import {zlibSync, strToU8} from 'fflate';
```
cpoile, line 8: My own interest, why this change?

cpoile (approved):

webapp/src/client.ts
streamer45: @cpoile It's ~40% faster, proper typescript and 5 times smaller bundle size. It got me convinced.

PR #876, opened 2024-10-11
Prepare v1.2.0 release

streamer45:
#### Summary

Updating the various dependencies. The only significant change here is I am bumping `rtcMonitorInterval` from 4s to 10s as it results in much higher quality stats (less variance) and possibly avoid more false positives with call quality detection.


cpoile (approved):
:+1:

PR #881, opened 2024-10-14
Fix docker login step

streamer45:
#### Summary

I think the existing check was overly strict since it's only running on PRs. In fact,  `main` branch is [now failing](https://github.com/mattermost/mattermost-plugin-calls/actions/runs/11331363665/job/31511439495) due to rate limiting.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -158,7 +158,7 @@ jobs:
       - name: e2e/docker-login
         uses: docker/login-action@0d4c9c5ea7693da7b068278f7b52bda2a190a446 # v3.2.0
         # Do not authenticate on Forks
-        if: github.event.pull_request.head.repo.full_name == github.repository
+        if: "! github.event.pull_request.head.repo.fork "
         with:
           username: ${{ secrets.DOCKERHUB_DEV_USERNAME }}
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e.yml
```diff
@@ -158,7 +158,7 @@ jobs:
       - name: e2e/docker-login
         uses: docker/login-action@0d4c9c5ea7693da7b068278f7b52bda2a190a446 # v3.2.0
         # Do not authenticate on Forks
-        if: github.event.pull_request.head.repo.full_name == github.repository
+        if: "! github.event.pull_request.head.repo.fork "
```
mvitale1989, line 161: Good find Claudio, thanks! This requires a bit more tweaking, because on `main` branch pushes the `github.event.pull_request` field will be empty.
```suggestion
        if: "github.event.repository.full_name == github.repository"
```

streamer45: Thanks!

mvitale1989 (approved):
Thank you Claudio!
For the E2E test, I can submit a fix PR tomorrow morning

streamer45:
> Thank you Claudio! For the E2E test, I can submit a fix PR tomorrow morning

No worries, there's nothing too urgent that needs to go in. I appreciate it!

PR #885, opened 2024-10-21
[MM-61166] Ensure full validation of Calls post props

streamer45:
#### Summary

Since props cannot be trusted (i.e., users can alter them at any time on their posts), we must take extra care to ensure that the type of them is what we expect it to be.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-61166


cpoile (approved):
Nice!

esarafianou (approved):

PR #887, opened 2024-10-23
[MM-61233] Verify no error is shown when running slash commands

streamer45:
#### Summary

Adding some extra checks so that we could have caught this way sooner.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-61233

### DIFF
--- a/e2e/tests/start_call.spec.ts
+++ b/e2e/tests/start_call.spec.ts
@@ -77,6 +77,7 @@ test.describe('start new call', () => {
         await page.locator('#post_textbox').fill('/call leave');
         await page.getByTestId('SendMessageButton').click();
         await expect(page.locator('#calls-widget')).toBeHidden();
+        await expect(page.locator('#postCreateFooter .has-error')).toBeHidden();
     });
 
     test('dm channel', async ({page}) => {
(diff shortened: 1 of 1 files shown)


e2e/tests/start_call.spec.ts
```diff
@@ -77,6 +77,7 @@ test.describe('start new call', () => {
         await page.locator('#post_textbox').fill('/call leave');
         await page.getByTestId('SendMessageButton').click();
         await expect(page.locator('#calls-widget')).toBeHidden();
+        await expect(page.locator('#postCreateFooter .has-error')).toBeHidden();
```
cpoile, line 80: nit: or not be visible? (because it's not really hidden...) 🤷

cpoile (approved):

e2e/tests/start_call.spec.ts
streamer45: @cpoile Playwright semantic is the same though. It's just a bit more compact to use `toBeHidden()` vs `not.toBeVisible()`.

Otherwise we'd have to use something like `toBeAttached` which specifically checks for presence in the DOM rather than visibility.

PR #888, opened 2024-10-24
[MM-61298] Fix e2e pipeline

streamer45:
#### Summary

Fixing the e2e pipeline. The only dependency we need is Postgres, so we can spin up that service by ourselves without relying on files from a different repo.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-61298


cpoile (approved):
Awesome

PR #892, opened 2024-10-28
Fix E2E tests for forks

mvitale1989:
#### Summary

Fix conditional for not logging into dockerhub when running on forks:
- Explicitly state which repository the E2E test should run for
- Remove quotes around `if` condition

I'll double check that this works and then mark this as ready-for-review

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -158,7 +158,7 @@ jobs:
       - name: e2e/docker-login
         uses: docker/login-action@0d4c9c5ea7693da7b068278f7b52bda2a190a446 # v3.2.0
         # Do not authenticate on Forks
-        if: "github.event.repository.full_name == github.repository"
+        if: github.event.repository.full_name == 'mattermost/mattermost-plugin-calls'
         with:
           username: ${{ secrets.DOCKERHUB_DEV_USERNAME }}
           password: ${{ secrets.DOCKERHUB_DEV_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e.yml
```diff
@@ -158,7 +158,7 @@ jobs:
       - name: e2e/docker-login
         uses: docker/login-action@0d4c9c5ea7693da7b068278f7b52bda2a190a446 # v3.2.0
         # Do not authenticate on Forks
-        if: "github.event.repository.full_name == github.repository"
+        if: github.event.repository.full_name == 'mattermost/mattermost-plugin-calls'
```
streamer45, line 161: Was the real problem the quotes then? I'd still expect `github.repository` to match `mattermost/mattermost-plugin-calls` in non forks right?

mvitale1989: The Github Actions linter in VScode marked the quotes as a problem, so yes they may indeed be.
I am running a some tests in another repository to be sure about the root cause, before marking this PR as ready4review

mvitale1989: After some testing, it turns out that the event origin is the target repo instead of the fork, even in the `pull_request` event. Here are the test results:
- PR, not from fork: https://github.com/mattermost/test-public-project/actions/runs/11556816926/job/32165381940?pr=23
- PR, from fork: https://github.com/mattermost/test-public-project/actions/runs/11556888244/job/32165619193?pr=25#step:2:12
- From branch push: https://github.com/mattermost/test-public-project/actions/runs/11557170713/job/32166562102#step:2:11
- 
The condition I used is ` "! github.event.pull_request.head.repo.fork"`, which is the one [you originally used](https://github.com/mattermost/mattermost-plugin-calls/pull/881#discussion_r1806478961) and works perfectly fine :sweat_smile: 

I have adapted this PR accordingly, sorry for the disruption

streamer45: Ah...cool, thanks for checking all that :raised_hands:

streamer45 (approved):

PR #893, opened 2024-10-29
[MM-61224] Add tooltip to screen source names

streamer45:
#### Summary

We implement a tooltip so that longer source names that it's easy to display the full name in case it gets truncated.

/cc @matthewbirtch 

#### Screenshot

image

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-61224

### DIFF
--- a/webapp/src/components/screen_source_modal/component.tsx
+++ b/webapp/src/components/screen_source_modal/component.tsx
@@ -2,6 +2,7 @@ import './component.scss';
 
 import {DesktopCaptureSource} from '@mattermost/desktop-api';
 import React, {CSSProperties} from 'react';
+import {OverlayTrigger, Tooltip} from 'react-bootstrap';
 import {IntlShape} from 'react-intl';
 import CompassIcon from 'src/components/icons/compassIcon';
 import {logDebug, logErr} from 'src/log';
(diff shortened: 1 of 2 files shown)


webapp/src/components/screen_source_modal/component.tsx
```diff
@@ -133,7 +134,17 @@ export default class ScreenSourceModal extends React.PureComponent<Props, State>
                             src={source.thumbnailURL}
                         />
                     </div>
-                    <span style={this.style.sourceLabel as CSSProperties}>{source.name}</span>
+
+                    <OverlayTrigger
```
cpoile, line 138: If I'm not mistaken, I think new code should use the new floating-ui framework? (There was a PR by Daniel adding a new tooltip we should use, I think...)
I'll leave it up to you though, considering this is just a small addition.

cpoile (approved):
Thanks!

webapp/src/components/screen_source_modal/component.tsx
streamer45: I think these should be migrated to WithTooltip (https://mattermost.atlassian.net/browse/MM-57219). Something for next Fixathon I suppose.

PR #908, opened 2024-11-20
[MM-61786] Reduce number of open file descriptors

streamer45:
#### Summary

Two changes on this side:

- Using the new `UDPSocketsCount` setting to limit the number of listening UDP sockets created by the embedded RTC server. In this mode, scaling isn't a goal, so we reduce them to NumCPU.
- I can't exactly recall what led to the existing activation sequence, but we'd be starting up the embedded RTC server regardless of whether RTCD is used. This seems wrong as it causes us to use unnecessary resources on the Mattermost side (e.g. opening sockets that will never be used).

#### Related PR

https://github.com/mattermost/rtcd/pull/162

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-61786

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -24,7 +24,7 @@ require (
 	github.com/mattermost/mattermost-plugin-calls/server/public v0.0.3
 	github.com/mattermost/mattermost/server/public v0.1.5-0.20240613070149-4b0ae20ef7b4
 	github.com/mattermost/morph v1.1.0
-	github.com/mattermost/rtcd v0.18.0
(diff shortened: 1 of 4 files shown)


cpoile (approved):
Huh, good catch.

Re: numCPUs, is that enough sockets, even if you're not worried about scaling?

streamer45:
> Huh, good catch.
> 
> Re: numCPUs, is that enough sockets, even if you're not worried about scaling?

We used numCPUs on rtcd before the latest ceiling tests and could easily achieve thousands of sessions per node. Certainly far more than what I'd be recommending when running in plugin mode.

PR #913, opened 2024-12-04
[MM-62065] e2e HA support

streamer45:
#### Summary

This required a decent amount of sweat, but it looks like it's finally working :sweat_smile: :tada: 

PR adds HA support to our e2e pipeline. This means the tests are now running against a 2-nodes MM cluster. 

As a follow up I'd like to start using `rtcd`, but I'll do it [later](https://mattermost.atlassian.net/browse/MM-62090) to let this soak a bit.

Also, there's still some flakyness with the `unmuting after ws reconnect` case which will need some further digging but this can still be reviewed.

#### Ticket

https://mattermost.atlassian.net/browse/MM-62065


e2e/scripts/run.sh
```diff
@@ -2,37 +2,59 @@
...
+}
+
+trap print_logs EXIT
+
+mkdir -p "${WORKSPACE}/results"
+
+# Generate sysadmin
+echo "Generating sysadmin ..."
+docker exec \
+	${CONTAINER_SERVER}1 \
```
cpoile, line 18: QQ: did I miss the setup for `${CONTAINER_SERVER}2`?

cpoile (approved):
Wow, this is... dockerrific. I gotta get on a call and see how you tested all this and maybe walk through some of it. 
Nice work!

e2e/scripts/run.sh
streamer45: We only need to create it once per cluster as the DB is shared.

streamer45:
Well, the test is failing due to an actual bug. This was worth it already :)

PR #920, opened 2024-12-10
[MM-62178] Fix signaling after WS reconnect in non-RTCD HA

streamer45:
#### Summary

Our new HA-capable e2e tests have surfaced a bug causing signaling issues for sessions reconnecting in a non-RTCD HA deployment.

The optimal path would still be to pull the trigger on https://mattermost.atlassian.net/browse/MM-53376, but we'll have to defer it to the next major version bump (MM v11 / Calls v2).

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-62178

### DIFF
--- a/server/cluster_message.go
+++ b/server/cluster_message.go
@@ -13,7 +13,10 @@ import (
 )
 
 type clusterMessage struct {
-	ConnID        string           `json:"conn_id,omitempty"`
+	ConnID string `json:"conn_id,omitempty"`
+	// NewConnID is used by clusterMessageTypeReconnect to inform other nodes
+	// about the new WS connection ID in case it changed.
+	NewConnID     string           `json:"new_conn_id,omitempty"`
 	UserID        string           `json:"user_id,omitempty"`
 	ChannelID     string           `json:"channel_id,omitempty"`
 	CallID        string           `json:"call_id,omitempty"`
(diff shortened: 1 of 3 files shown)


streamer45:
image
 :tada:

cpoile (approved):
I guess these are covered under the existing e2e tests? Will this bring down our test/code ratio? :)

I would argue that we don't really need to wait for v2 -- it's a breaking change in that it's removing functionality, but it's not breaking an API or a version compatibility agreement, right? Maybe I'm a little fuzzy on what a "breaking change" really means...

streamer45:
> I guess these are covered under the existing e2e tests? Will this bring down our test/code ratio? :)

Yes, for the time being, they are. If we move to using `rtcd` in our pipeline (https://mattermost.atlassian.net/browse/MM-62090) then we won't be stressing these paths any longer unless we test both scenarios at the expense of likely doubling execution time (or cost).

> I would argue that we don't really need to wait for v2 -- it's a breaking change in that it's removing functionality, but it's not breaking an API or a version compatibility agreement, right? Maybe I'm a little fuzzy on what a "breaking change" really means...

Well, it's totally breaking because if even a single customer is using this today, as soon as they upgrade it will break Calls. And replacing with rtcd is far from being a config setting away.

PR #922, opened 2024-12-11
[MM-62090] Include rtcd service in e2e pipeline

streamer45:
#### Summary

PR adds `rtcd` support to our e2e pipeline. The rtcd image is dynamically built from a matching branch name (if it exists) or `master` as fallback. This will help with running the entire pipeline against specific changes on the other side.

I am also tagging a subset of tests to execute in plugin mode (i.e. without rtcd service). These should be fairly quick and will run in addition to the above and help to avoid a potential blind spot.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-62090

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -52,6 +52,53 @@ jobs:
           compression-level: 0
           retention-days: 1
 
+  build-rtcd-image:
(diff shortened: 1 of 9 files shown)


cpoile (commented):
Nice!

cpoile (approved):

PR #932, opened 2024-12-20
Fix flaky AV1 e2e

streamer45:
#### Summary

The AV1 e2e tests have been failing too often for my liking, so we are trying something slightly different and hopefully more consistent.

### DIFF
--- a/e2e/tests/media.spec.ts
+++ b/e2e/tests/media.spec.ts
@@ -160,46 +160,35 @@ test.describe('screen sharing', () => {
 
         type rtcCodecStats = {
             type: string,
-            mimeType: string,
+            codecId: string,
         };
 
-        type rtcCodecs = {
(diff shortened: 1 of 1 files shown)


cpoile (approved):
I was going to make some smarty-pants comment that instead of `stats.forEach(...`) you could use a `stats.find(..`, but it really won't make much of a difference. :)

streamer45:
> I was going to make some smarty-pants comment that instead of `stats.forEach(...`) you could use a `stats.find(..`, but it really won't make much of a difference. :)

@cpoile Sadly it won't work because `stats` is not an array. It's this weird map like object that requires looping with `forEach`.

See https://developer.mozilla.org/en-US/docs/Web/API/RTCStatsReport if you are curious.

cpoile:
Oh, well good thing I didn't make that comment then! :)

PR #940, opened 2025-01-08
[MM-62410] Automatically cleanup stale calls in RTCD enabled deployments

streamer45:
#### Summary

PR adds logic to automatically clean up stale calls/sessions on plugin activation under RTCD-backed deployments. This is achieved by asking the RTCD side if the calls are still ongoing.

#### Related PR

https://github.com/mattermost/rtcd/pull/166

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-62410


server/rtcd.go
```diff
@@ -706,3 +708,51 @@ func (h *rtcdHost) isFlagged() bool {
...
+	m.ctx.LogDebug("RTCD host is set in call, checking...", "callID", call.ID, "rtcdHost", call.Props.RTCDHost)
+	host := m.getHost(call.Props.RTCDHost)
+	if host != nil {
+		m.ctx.LogDebug("RTCD host found", "callID", call.ID, "rtcdHost", call.Props.RTCDHost)
+
+		// Version compatibility check. We need to be talking to RTCD v1.0.0 or higher to be able to call GetSessions.
+		info, err := host.client.GetVersionInfo()
+		if err != nil {
+			m.ctx.LogDebug("failed to get version info", "err", err.Error(), "callID", call.ID, "rtcdHost", call.Props.RTCDHost)
+			return false
```
cpoile, line 722: This defaults to call is still running if there's a problem getting the version info -- will that cause trouble? Should we return the err and have logic for it?

server/rtcd.go
```diff
@@ -706,3 +708,51 @@ func (h *rtcdHost) isFlagged() bool {
...
+		}
+
+		cfgs, code, err := host.client.GetSessions(call.ID)
+		if err != nil {
+			m.ctx.LogDebug("failed to get sessions for call", "err", err.Error(), "callID", call.ID, "rtcdHost", call.Props.RTCDHost)
+		}
+
+		if code != http.StatusOK && code != http.StatusNotFound {
+			m.ctx.LogDebug("unexpected status code from RTCD", "code", code, "callID", call.ID, "rtcdHost", call.Props.RTCDHost)
+			// The request above could fail for various reasons, in which case we can't assume the call has ended.
```
cpoile, line 740: ah, this confirms the default behavior. Maybe add this comment to the above as well? (would just save some cognitive overhead)

server/state_test.go
```diff
@@ -545,3 +555,555 @@ func BenchmarkCallStateClone(b *testing.B) {
...
+
+			err = p.store.CreateCallSession(&public.CallSession{
+				ID:     "connA",
+				CallID: callID,
+				UserID: "userA",
+				JoinAt: time.Now().UnixMilli(),
+			})
+			require.NoError(t, err)
+
+			mockAPI.On("LogDebug", "cleaning up calls state",
```
cpoile, line 617: this must have been a pain to get exactly right, but I kind of like that the tests have the logs laid out like this -- you can see exactly what's supposed to happen. It's easy to understand.

streamer45: Yeah, it takes some time, but it ensures the tests are solid, even in the presence of heavy mocking.

PR #948, opened 2025-01-20
[MM-62241] Fix screen sharing from popout on new Desktop versions

streamer45:
#### Summary

We need some plugin changes to make https://github.com/mattermost/desktop/pull/3258 work. We'll need to backport this fix in earlier plugin versions to prevent the bug when users upgrade desktop version.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-62241


cpoile (approved):
Does this mean we're removing the backwards compatibility?

streamer45:
> Does this mean we're removing the backwards compatibility?

@cpoile It's getting removed from Desktop since we had a couple of releases already with the new API, which is what caused the issue to surface. But this is a breaking change because we failed to update the plugin side in time. In popout we were still relying on the legacy (deprecated) API.

DHaussermann (approved):
Tested and passed 
- Share option works on Desktop from expanded view
- Leave Call option embedded in the post is now working
- Did a click-through test to ensure all options in the modal and the expanded view are functional on desktop
https://github.com/mattermost/desktop/pull/3258#pullrequestreview-2562898135

LGTM!

DHaussermann:
@streamer45 we could merge this now and open a new issue on the off chance Desktop `v5.11.0-rc1` presents any new problems.

streamer45:
> @streamer45 we could merge this now and open a new issue on the off chance Desktop `v5.11.0-rc1` presents any new problems.

I am waiting to fix e2e tests (https://github.com/mattermost/mattermost-plugin-calls/pull/950). I am also planning a couple of Calls dot releases, namely v1.5.1 and v0.29.7 to include this fix since these functionalities will break when using the new desktop version against older plugins.

PR #949, opened 2025-01-20
[MM-62637] Fix invalid parsing of `TURNStaticAuthSecret` setting

### LINKED ISSUE #945: TURN Static Auth Secret is parsed as int, should be string
When trying to set the turn static auth secret in the admin console, I cannot set the secret because the input only allows numeric values... but the secret is a string value!

I think the relevant code is here:
https://github.com/mattermost/mattermost-plugin-calls/blob/main/webapp/src/components/admin_console_settings/turn_static_auth_secret.tsx#L16

I'm not familiar enough with the framework to say what should be there, but the call to `parseInt` seems wrong.

streamer45:
#### Summary

Fixes #945 

I had to update some e2e snapshots because the checks were passing with clearly different values. We'll probably need to review the global [`maxDiffPixelRatio`](https://github.com/mattermost/mattermost-plugin-calls/blob/c799a1f98fa9c7318acd9fdbfeeb041d298faf52/e2e/playwright.config.ts#L16) value as well but will defer to a separate PR.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-62637


cpoile (approved):
Hm, interesting. I guess this diff test (and the tolerances) aren't meant for UI elements like this? (we should be testing with the dom?) Maybe it's more for things that can only be visually tested.

streamer45:
There's a UI regression in MM `master` (https://community.mattermost.com/core/pl/4o8sr1xtjinyud119ip8xbrxhr), but nothing from this side that should block the release. Will merge.

streamer45:
> Hm, interesting. I guess this diff test (and the tolerances) aren't meant for UI elements like this? (we should be testing with the dom?) Maybe it's more for things that can only be visually tested.

I like visual testing because it matches what end users will experience. I am slightly worried about solely relying on DOM level checks.

PR #973, opened 2025-02-17
Fix config defaults in admin console

streamer45:
#### Summary

While updating the snapshots to fix some e2e failures I noticed some config defaults were not handled correctly in the presence of a set of options. If the setting isn't defined, the first value in the options array is picked, which is not always correct (e.g., the default recording quality is medium and not low).


cpoile:
> image

nice 🎉

cpoile (approved):

PR #978, opened 2025-02-24
Use stricter maxDiffPixelRatio in e2e tests

streamer45:
#### Summary

Using a stricter `maxDiffPixelRatio` value for snapshot matching. To make it happy, we have to remove some of the randomness around usernames and regenerate a few snapshots.


cpoile (approved):
Hah, clever solution. 
If you do that too much though... :)

PR #979, opened 2025-02-25
[MM-63274] Use device label instead of id to save defaults

streamer45:
#### Summary

I noticed that every time I replug my camera or restart the laptop, my default devices are reset. This is because the device IDs are not consistent across these events. 

What remains stable, though, is the device label, so I am opting to use this instead.

I think there's a small chance we could get more than one device with the same label, so we still support strict ID matching as a fallback in such cases.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-63274


cpoile (approved):
Cool, looks clean, thanks!

PR #986, opened 2025-03-03
[MM-63283] Fix UI regression in call card component

streamer45:
#### Summary

Using the external (webapp provided) `styled-components` broke our `@container` CSS queries since they are supported starting from v6 (webapp is still using v5). This caused a UI regression in the call card. I added an e2e test to spot future breakage. 

~~Also updating `styled-components` to latest since we are here.~~

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-63283


cpoile:
The incompatibility looks to be deep in the IStyledComponentBase (did they change from div to button somewhere?). Anyway, I vote for not doing the change and waiting.

PR #990, opened 2025-03-12
[MM-63373] Replace usage of `SELECT *`

streamer45:
#### Summary

When reverting from the video branch, everything broke down. This is because `SELECT *` is not backward compatible in case a new column is added.

Tests were Aider-generated, which I reviewed.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-63373


cpoile (approved):
AI is good at these kinds of tests, nice.

PR #995, opened 2025-03-19
E2E improvements

streamer45:
#### Summary

Some improvements to e2e tests:

- Update Playwright to latest
- Using resources more wisely by avoiding spawning unnecessary browser processes.
- Using `Promise.all()` where applicable to increase parallelism of expensive operations.

### DIFF
--- a/e2e/tests/slash_commands.spec.ts
+++ b/e2e/tests/slash_commands.spec.ts
@@ -4,7 +4,7 @@
 import {expect, test} from '@playwright/test';
 
 import PlaywrightDevPage from '../page';
-import {getUserIdxForTest, getUsernamesForTest, getUserStoragesForTest, joinCall, startCall} from '../utils';
+import {getUserIdxForTest, getUsernamesForTest, getUserStoragesForTest, joinCall} from '../utils';
 
 const userStorages = getUserStoragesForTest();
 const usernames = getUsernamesForTest();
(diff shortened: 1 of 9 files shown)


e2e/tests/slash_commands.spec.ts
```diff
@@ -33,9 +33,13 @@ test.describe('slash commands', () => {
...
-    test('end call as second host', async () => {
-        const user0Page = await startCall(userStorages[0]);
-        const user1Page = await joinCall(userStorages[1]);
+    test('end call as second host', async ({page}) => {
+        const user0Page = new PlaywrightDevPage(page);
+
+        const [, user1Page] = await Promise.all([
+            user0Page.startCall(),
+            joinCall(userStorages[1]),
+        ]);
```
streamer45, line 42: Here and in a few places that expect certain permissions on a given user, we are potentially introducing flakiness since the host is the first user to join and now both users join in parallel. That said, in one case (the second user) we have to spawn a new browser process, so I am assuming the first user will consistently beat that, guaranteeing the order. If that doesn't happen to be the case, I'll revert.

cpoile: Is there a reason you need to do this part in parallel then?

streamer45: @cpoile The reason is merely to speed up. Are you concerned?

cpoile: Maybe a comment then that this is a potential issue, just to remind someone else if it's not you. 👍

cpoile (approved):

e2e/tests/slash_commands.spec.ts
streamer45: Done

PR #997, opened 2025-03-24
[MM-63430] Calls 1-1 video support (MVP) :video_camera: 

streamer45:
#### Summary

PR implements initial video support in DMs. The functionality is expected to be behind a new experimental `EnableVideo` config setting, which defaults to false.

The only slightly more involved logic is the image segmentation that powers the background blur functionality. This required adding an external dependency ([`@mediapipe/tasks-vision`](https://www.npmjs.com/package/@mediapipe/tasks-vision)) plus some WASM assets and a model file (see under `public/`). We can consider removing this from the initial MVP if we think it's too much.

UX is a real work in progress, so don't focus too much on it. It will most certainly change before we merge as we are waiting for some designs.

#### Screenshots

*Widget view*
image

*Popout view*
image

*Screen sharing*

image

*User settings*

image

#### Related PRs

https://github.com/mattermost/rtcd/pull/173
https://github.com/mattermost/calls-common/pull/47
https://github.com/mattermost/mattermost-mobile/pull/8716

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-63430


standalone/src/widget/index.tsx
```diff
@@ -59,6 +66,50 @@ async function initWidget({store, startingCall}: InitCbProps) {
         }
     }
 
+    window.callsClient?.on('mute', () => {
```
streamer45, line 69: This is an inconsistency I found when testing video on desktop. On the web side we do this to keep the redux state in sync with the local client. This avoids the issue of a perceivable lag between pressing mute/unmute and getting the UX to react, since otherwise we'd be waiting for a server side websocket event.

webapp/src/client.ts
```diff
@@ -18,16 +18,34 @@ import {
...
+    // TODO: consider exposing in user preferences
+    frameRate: {
+        ideal: 30,
+    },
+    aspectRatio: {
+        ideal: 4 / 3,
+    },
+    width: {
+        ideal: 640,
+    },
```
streamer45, line 47: This is temporary, for testing things. In the absence of user preferences, we should respect the device's default aspect ratio.

webapp/src/components/call_widget/component.tsx
```diff
@@ -2252,6 +2456,28 @@ export default class CallWidget extends React.PureComponent<Props, State> {
...
                         />
 
+                        {this.props.enableVideo &&
+                            <WidgetButton
+                                id='video-start-stop'
+                                // eslint-disable-next-line no-undefined
+                                onToggle={noVideoInputDevices ? undefined : this.onVideoToggle}
+                                ariaLabel={videoTooltipText}
+
+                                //shortcut={} TODO: add shortcut
```
streamer45, line 2466: TBD, waiting for UX input.

webapp/src/components/call_widget/loading_overlay.tsx
```diff
@@ -1,77 +0,0 @@
...
-        return joining;
-    }, [/* intentionally empty */]);
-
-    if (!visible && animationEnded) {
-        return null;
-    }
-
-    const onAnimationEnd = () => {
-        setAnimationEnded(true);
-    };
```
streamer45, line 27: This would never fire and leave the component mounted for the whole duration of the call at zero opacity. I am fixing it as part of this PR.

webapp/src/components/expanded_view/component.tsx
```diff
@@ -1142,6 +1273,33 @@ export default class ExpandedView extends React.PureComponent<Props, State> {
...
+                            <ControlsButton
+                                id='calls-popout-video-button'
+                                dataTestId={isVideoOn ? 'calls-popout-stop-video' : 'calls-popout-start-video'}
+                                ariaLabel={videoTooltipText}
+                                // eslint-disable-next-line no-undefined
+                                onToggle={noVideoInputDevices ? undefined : this.onVideoToggle}
+                                tooltipText={videoTooltipText}
+                                tooltipSubtext={videoTooltipSubtext}
+                                // eslint-disable-next-line no-undefined
+                                // shortcut={noVideoInputDevices || noVideoPermissions ? undefined : reverseKeyMappings.popout[MUTE_UNMUTE][0]}
```
streamer45, line 1286: Same as for widget, TBD.

webapp/src/components/user_settings/video_devices_settings_section.tsx
```diff
@@ -0,0 +1,446 @@
...
+                                <input
+                                    type='checkbox'
+                                    id='mirror'
+                                    name='mirror'
+                                    checked={mirrorVideo}
+                                    onChange={e => setMirrorVideo(e.target.checked)}
+                                />
+                                <label htmlFor='mirror'>{formatMessage({defaultMessage: 'Mirror video'})}</label>
+                            </CheckBoxContainer>
+                            {!isFirefox() &&
```
streamer45, line 396: Unfortunately, GPU acceleration isn't working yet on Firefox (see https://github.com/google-ai-edge/mediapipe/issues/5879). I tested CPU only but it's just not going to work. Anyhow, this is experimental on top of an already experimental feature.

webapp/src/index.tsx
```diff
@@ -317,6 +334,9 @@ export default class Plugin {
             Client4.setUrl(window.basename);
         }
 
+        const theme = getTheme(store.getState());
+        setCallsGlobalCSSVars(theme.sidebarBg);
```
streamer45, line 338: Another small bug I found when testing on Desktop.

matthewbirtch:
> /cc @matthewbirtch for visibility.

@streamer45 do you want me to review this from a UX standpoint, or are you just keeping me in the loop?

streamer45:
> > /cc @matthewbirtch for visibility.
> 
> @streamer45 do you want me to review this from a UX standpoint, or are you just keeping me in the loop?

@matthewbirtch The latter, for the time being. Of course early feedback is always welcome, but ideally I'd be waiting to have some Figma designs ready before asking for UX review because I've mostly improvised so far :)

webapp/src/components/user_settings/video_devices_settings_section.tsx
```diff
@@ -0,0 +1,446 @@
...
+                                    const stream = await navigator.mediaDevices.getUserMedia({
+                                        video: videoOpts,
+                                        audio: false,
+                                    });
+
+                                    setPreviewStream(stream);
+                                }}
+                                ref={videoInputsRef}
+                            />
+                            <VideoPreview
```
cpoile, line 380: is there a way to reuse components, or is it not worth the trouble?

webapp/src/index.tsx
cpoile: ah, that's why it was up above. What was the bug?

webapp/src/local_storage.ts
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) 2020-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {STORAGE_CALLS_BLUR_BACKGROUND_KEY} from 'src/constants';
+
+export type BgBlurData = {
```
cpoile, line 6: i keep reading this as "big blur data" and I like it

webapp/src/segmenter/segmenter.worker.ts
```diff
@@ -0,0 +1,107 @@
...
+        }
+
+        segmenter.segmentForVideo(data.frame, performance.now(), async (result) => {
+            if (!result.categoryMask) {
+                console.warn('segmeneter.worker: no category mask in result');
+                return;
+            }
+
+            // TODO: consider applying blur using WebGL instead to avoid using the
+            // expensive CPU canvas context and CSS filter.
```
cpoile, line 62: that would be cool

cpoile (approved):
🎉  Amazing work.
Just lots of questions. 🧑‍🎓

webapp/src/client.ts
streamer45: Most likely, although this is just about the source. 

I still need to review aspect ratios in general in the UI, was waiting for some designs to understand what we want to do with them. I see a few options:

- Clipping to cover all the available space without stretching (what we do today through [`object-fit: cover`](https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit)).
- Show the full video content but add bands (lateral or vertical) to maintain the original ratio.
- Dynamically modify widget dimensions (either height or width) to match the original ratio.
- Stretch to cover the entire space (I wouldn't do this).

standalone/src/init.ts
```diff
@@ -319,6 +331,7 @@ export default async function init(cfg: InitConfig) {
 
     const theme = getTheme(store.getState());
     applyTheme(theme);
+    setCallsGlobalCSSVars(theme.sidebarBg);
```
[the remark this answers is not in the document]
streamer45: Yep, as explained in https://github.com/mattermost/mattermost-plugin-calls/pull/997#discussion_r2011057511.

standalone/tsconfig.json
```diff
@@ -13,7 +13,7 @@
     "strict": true,
     "strictNullChecks": true,
     "forceConsistentCasingInFileNames": true,
-    "module": "commonjs",
+    "module": "es2020",
```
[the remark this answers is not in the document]
streamer45: This matches what we do on [webapp](https://github.com/mattermost/mattermost/blob/c3f02b8ebc1f6d9df80554c51cc06f61e48fbf08/webapp/channels/tsconfig.json#L17). Without it, we'd be getting an error about `import.meta.url` not existing, which is needed for the web worker to function.

webapp/src/segmenter/segmenter.worker.ts
streamer45: It's actually not too tricky, but we'd likely still have to support canvas as a fallback, so it becomes a little more troublesome to maintain.

webapp/src/components/user_settings/video_devices_settings_section.tsx
streamer45: From the widget you mean? They serve very different purposes. Maybe a shared subcomponent but I am not sure.

webapp/src/client.ts
```diff
@@ -724,6 +929,121 @@ export default class CallsClient extends EventEmitter {
...
+        canvas.height = localVideoTrack.getSettings().height!;
+        video.autoplay = true;
+        video.srcObject = new MediaStream([localVideoTrack]);
+
+        const outStream = canvas.captureStream(30);
+        this.streams.push(outStream);
+        const outTrack = outStream.getVideoTracks()[0];
+
+        stream.removeTrack(localVideoTrack);
+        stream.addTrack(outTrack);
```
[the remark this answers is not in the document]
streamer45: Yeah, we are replacing on the original stream so that its id doesn't change.

cpoile (approved):

PR #1011, opened 2025-04-24
[MM-63889] Add backend version check to enable DC locking

streamer45:
#### Summary

We expose RTCD versioning info on top of the existing plugin version/build pair. We then use this information to detect whether we can safely use the new DC locking functionality.

A test plan for both backward and forward compatibility is attached to the JIRA ticket.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-63889

### DIFF
--- a/webapp/src/types/types.ts
+++ b/webapp/src/types/types.ts
@@ -37,6 +37,7 @@ export type CallsClientConfig = {
     simulcast?: boolean;
     enableAV1: boolean;
     dcSignaling: boolean;
+    dcLocking: boolean;
 }
 
 export type AudioDevices = {
--- a/server/api.go
+++ b/server/api.go
@@ -26,10 +26,16 @@ import (
 const requestBodyMaxSizeBytes = 1024 * 1024 // 1MB
 
 func (p *Plugin) handleGetVersion(w http.ResponseWriter, _ *http.Request) {
-	info := map[string]interface{}{
-		"version": manifest.Version,
-		"build":   buildHash,
+	p.mut.RLock()
+	defer p.mut.RUnlock()
+
+	info := public.VersionInfo{
+		Version:     manifest.Version,
+		Build:       buildHash,
+		RTCDVersion: p.rtcdVersionInfo.BuildVersion,
+		RTCDBuild:   p.rtcdVersionInfo.BuildHash,
 	}
+
 	w.Header().Set("Content-Type", "application/json")
 	if err := json.NewEncoder(w).Encode(info); err != nil {
 		p.LogError(err.Error())
--- a/server/plugin.go
+++ b/server/plugin.go
@@ -19,6 +19,7 @@ import (
 	"github.com/mattermost/mattermost-plugin-calls/server/interfaces"
 	"github.com/mattermost/mattermost-plugin-calls/server/telemetry"
 
+	rtcd "github.com/mattermost/rtcd/service"
 	"github.com/mattermost/rtcd/service/rtc"
 
 	"github.com/mattermost/mattermost/server/public/model"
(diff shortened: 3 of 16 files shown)


webapp/src/types/types.ts
```diff
@@ -37,6 +37,7 @@ export type CallsClientConfig = {
     simulcast?: boolean;
     enableAV1: boolean;
     dcSignaling: boolean;
+    dcLocking: boolean;
 }
```
cpoile, line 41: I think we talked about this, but it's been awhile I guess -- shouldn't this be moved over to calls-common?  Maybe that was a TODO...?

cpoile (approved):
Nice!

webapp/src/types/types.ts
streamer45: @cpoile This is not RTCPeer, it's the wrapper, which at this point is only used on this side.

DHaussermann:
@streamer45 I don't have mobile builds yet but I did start looking and the webapp cases
in the ticket.
- _Older plugin version (pre 1.7.0) against newest RTCD_  - This seemed to work fine  when exercising mute/unmute and screen share
- _Newest plugin version (i.e. with changes fixing this ticket) against older RTCD (pre 1.1.0)_ - This is not working for me.  
When I run the Calls build from this PR and RTCD v1.0.2 I can't start a call. The option in the UI becomes non-responsive. I see nothing of interest in the server logs. So I looked at the dev console and there is a relevant error.
image

In case this was just some issue where the front end was out of date, I made sure my local was on `master` and rebuild the front end but the behavior did not change. Can you reproduce this? ☝️ or advise on if I may be hitting some type of setup issue.

streamer45:
@DHaussermann It looks like the client still thinks it's running against a newer version. A couple of things may be happening here:

- You haven't reloaded the page after switching RTCD version. We don't broadcast new versions dynamically. I didn't think the complexity was worth it for an edge case (e.g. upgrading/downgrading RTCD without restarting the plugin).
- You are running a dev RTCD version.
- There's a bug :p

The source of truth for deciding whether or not to use the new logic is the output of `/plugins/com.mattermost.calls/version` HTTP handler. Could you share that?

DHaussermann:
Hi @streamer45, 
- The Calls version info shows `{"version":"1.7.0+999b4a04","build":"999b4a04e1122d9b2f5315021b945503002d3c50"}`
- > "You are running a dev RTCD version."

Yes. I suppose I must be running a dev version. All I did to get this going was to checkout the `v1.0.2` from source and type `make go-run` to get it up and going. 
This is the only way I've ever setup RTCD locally. Since I'm on Mac OS, I'm not sure the pipeline accommodates me so, I use my own Go installation.
Can you provide high level guidance for how I would get a a release version running on Mac? Or another option may be to run the desired version of RTCD elsewhere and I'd be able to point my local to it. (Or I guess spin-up a `/cloud` server).

DHaussermann (approved):
Tested and passed

I have tested what I believe is every applicable scenario. 
A couple tests may be redundant as RTCD being enabled may superseded what  happens at the plugin level. But this seems like a safe list of scenarios based on details provided by Claudio.   
In all cases no issues joining a call and muting / unmuting from the mobile clients. Also no issue with viewing screen share. 

_Old and current RTCD with new Calls plugin_   
- Tested latest call 1.7.0+pr with RTCD v1.0.2 with new mobile app 
- Tested latest call 1.7.0+pr with RTCD v1.1.0 with new mobile app

_Older version of calls with old and latest RTCD_
- Tested Calls 1.5.2 with RTCD v1.0.2 and new mobile app
- Tested Calls 1.5.2 with RTCD v1.1.0 and new mobile app

_Older version of calls and new with no RTCD_
- Tested Call 1.5.2 with no RTCD and new Mobile PR apps
- Tested latest call 1.7.0+pr with no RTCD and new Mobile PR apps

_Old mobile clients work with newest_   
- Tested Older mobile version (2.25) against Calls 1.7.0+pr
- Tested Older mobile version (2.25) against Calls 1.7.0+pr with newest RTCD V1.1.0

**Notes**:
1. In all cases "_new mobile app_" refers to the pr builds https://github.com/mattermost/mattermost-mobile/pull/8799 and **not** `2.27..0`
2. When Mobile apps were tested - I tested both Android and iOS 

@streamer45 if you see anything missing please ping and I'm happy to revisit. Thaks for all the setup help on this 🎉

streamer45:
Amazing work @DHaussermann , thank you for covering everything and more :raised_hands:

PR #1014, opened 2025-05-09
Enteprise Advanced sku support

marianunez:
#### Summary

Added support for new high level sku Enterprise Advanced. Removed legacy E10/E20 sku.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-63646


webapp/src/selectors.ts
```diff
@@ -567,16 +569,17 @@ export const shouldPlayJoinUserSound = (state: GlobalState): boolean =>
 
 export const isOnPremNotEnterprise = (state: GlobalState): boolean => {
     const license = getLicense(state);
-    const enterprise = license.SkuShortName === LicenseSkus.E20 || license.SkuShortName === LicenseSkus.Enterprise;
+    const enterprise = license.SkuShortName === LicenseSkus.Enterprise || license.SkuShortName === LicenseSkus.EntepriseAdvanced;
```
cpoile, line 572: When did we completely remove the `E10` and `E20` sku names? (I just want to make sure we're not removing support for our minimum supported MM version.)

server/license/license.go
```diff
@@ -62,11 +61,11 @@ func IsE10LicensedOrDevelopment(config *model.Config, license *model.License) bo
 	return IsConfiguredForDevelopment(config)
 }
 
-// IsE20LicensedOrDevelopment returns true when the server is licensed with a legacy Mattermost
-// Enterprise E20 License or a Mattermost Enterprise License, or has `EnableDeveloper` and
+// IsEnterpriseLicensedOrDevelopment returns true when the server is at least licensed with
```
cpoile, line 64:
```suggestion
// IsMinimumEnterpriseLicensedOrDevelopment returns true when the server is at least licensed with
```

server/license/license.go
```diff
@@ -26,25 +25,25 @@ func IsEnterpriseLicensedOrDevelopment(config *model.Config, license *model.Lice
...
 }
 
-// IsE10LicensedOrDevelopment returns true when the server is at least licensed with a legacy Mattermost
-// Enterprise E10 License or a Mattermost Professional License, or has `EnableDeveloper` and
+// IsProfessionalLicensedOrDevelopment returns true when the server is at least licensed with
+// a Mattermost Professional License, or has `EnableDeveloper` and
 // `EnableTesting` configuration settings enabled, signaling a non-production, developer mode.
-func IsE10LicensedOrDevelopment(config *model.Config, license *model.License) bool {
-	if IsProfessional(license) || IsEnterprise(license) {
+func IsProfessionalLicensedOrDevelopment(config *model.Config, license *model.License) bool {
```
cpoile, line 45: While we're at it, we should fix this:
```suggestion
func IsMinimumProfessionalLicensedOrDevelopment(config *model.Config, license *model.License) bool {
```

server/license/license.go
```diff
@@ -26,25 +25,25 @@ func IsEnterpriseLicensedOrDevelopment(config *model.Config, license *model.Lice
...
+	case professional, enterprise, enterpriseAdvanced:
 		return true
 	default:
 		return false
 	}
 }
 
-// IsE10LicensedOrDevelopment returns true when the server is at least licensed with a legacy Mattermost
-// Enterprise E10 License or a Mattermost Professional License, or has `EnableDeveloper` and
+// IsProfessionalLicensedOrDevelopment returns true when the server is at least licensed with
```
cpoile, line 42:
```suggestion
// IsMinimumProfessionalLicensedOrDevelopment returns true when the server is at least licensed with
```

cpoile (commented):
Thanks Maria!

webapp/src/selectors.ts
marianunez: E10 and E20 are legacy SKUs that were phased out several years ago, and all of our customers have since been updated to the latest SKUs. Their deprecation is being officially announced at the code level in version 10.8.here: https://github.com/mattermost/docs/pull/7888/files#diff-1c3ff347e1594f6e28919888f66ac98d74b7c69a6f45906de559bd4619e3a226

Removing it from the codebase for simplicity as is no longer needed.

server/license/license.go
marianunez: Updated!

marianunez:
@streamer45 @cpoile Is there a reason we are not using the plugin api for these license checks? The original idea of this change was to just update the mm public dependency to the latest that would have this changes included but was surprised that the plugin api is not used here. Just curious

streamer45:
> @streamer45 @cpoile Is there a reason we are not using the plugin api for these license checks? The original idea of this change was to just update the mm public dependency to the latest that would have this changes included but was surprised that the plugin api is not used here. Just curious

Maintaining that dependency back then was a pain as it caused all sorts of conflicts, so since we didn't need it for 99% of functionality, we decided it was simpler to implement the remaining 1%.

webapp/src/components/channel_header_dropdown_button/index.ts
```diff
@@ -26,7 +26,7 @@ const mapStateToProps = (state: GlobalState) => {
         hasCall: currentChannelHasCall(state),
         isAdmin: isCurrentUserSystemAdmin(state),
         isCloudStarter: isCloudStarter(state),
-        isCloudPaid: isCloudProfessionalOrEnterpriseOrTrial(state),
+        isCloudPaid: isCloudProfessionalOrEnterpriseorEnterpriseAdvanceOrTrial(state),
```
marianunez, line 29: Wondering if these checks are still needed for cloud? Could be out of scope but wondering if we can simplify these and just check for trials @cpoile @streamer45

cpoile:
@marianunez Sorry I can't be of much help, I'm not sure what that means.

streamer45:
> @marianunez Sorry I can't be of much help, I'm not sure what that means.

But you approved the fix for it :p https://github.com/mattermost/mattermost-plugin-calls/pull/1016

@marianunez Just merge with main and it should be good.

webapp/src/components/channel_header_dropdown_button/index.ts
streamer45: I'd leave these be for now because they would need changes on Mobile as well I think. We'll review them eventually.

streamer45 (approved):

cpoile (approved):
Thanks Maria!

PR #1016, opened 2025-05-12
Fix CI

streamer45:
#### Summary

Trying to fix tests in CI. I have no idea why this works, but I had to do the same locally for a long time so I assume it's a Docker version quirk.


cpoile (approved):
🤷

PR #1018, opened 2025-05-15
[MM-63741] Implement automatic fallback logic for audio devices

streamer45:
#### Summary

PR implements automatic fallback logic if the current audio (input/output) device suddenly disconnects (e.g., if it is unplugged). In such a case, we fall back to the system default.

In addition, if the user-selected device (i.e., one stored in localStorage) comes back, we'll automatically switch to it.

#### Example

https://github.com/user-attachments/assets/6c8dcb27-40a4-4ce5-9a89-70785af52f48

#### Test Plan

- Test the fallback logic in calls for both input and output audio devices:
  - Trying to unplug a microphone during a call should automatically fall back to the system default.
  - The same should happen for an output device (e.g. bluetooth headset). If unavailable, we should fall back to the system default (e.g., wired headphones or speakers).
  - Plugging back / reconnecting a user-selected device (i.e., explicitly chosen in the input settings menu) should cause it to be used as soon as available.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-63741


webapp/src/client.ts
```diff
@@ -84,100 +84,133 @@ export default class CallsClient extends EventEmitter {
...
-                };
-                this.currentAudioInputDevice = devices[0];
-            } else {
-                logDebug('audio input device not found');
-                window.localStorage.removeItem(STORAGE_CALLS_DEFAULT_AUDIO_INPUT_KEY);
-            }
+        if (devices.length > 1) {
+            // If there are multiple devices with the same label, we select the selected device by ID.
+            logInfo(`getSelectedAudioDevice: multiple audio ${deviceType} devices found with the same label, checking by id`, devices);
+            devices = devices.filter((dev) => dev.deviceId === selectedDevice.deviceId);
```
cpoile: Could we use a find instead of a filter? It fits the intention better than a filter + check.

cpoile (approved):
Nice and clean, thanks!

streamer45:
@DHaussermann Would be great if you could give this a good try. I left a test plan in the PR description.

DHaussermann (approved):
Tested and passed
- If you pair a Bluetooth mic and headphones to the OS as default device, Calls will also use them by default as the audio in/out devices.
- When the bluetooth device is disconnected, Call will fail over to the previous defaults automatically without the user needing to select them
- If you reconnect the bluetooth device and have the OS set to use it by default when connected, Calls will then automatically switch to it again
- Similar to case above - If I pair or connect a device while the call is already ongoing the new audio in/out options become available and will automatically be used as default if the OS is using them as default for sound   
- Tested with conventional audio in/out using the device stereo jack. The fail over works here as well. If a device is unplugged Calls automatically switches to the new OS default
- Regression tested that if a user manually sets audio in/out devices that are different from the OS default, these are respected and connecting new audio in/out devices while the Call in running does not affect this

Tested on Mac OS
No issues found
LGTM!

PR #1023, opened 2025-05-21
Generate code coverage during testing

streamer45:
#### Summary

PR implements code coverage generation and reporting through Codecov. Eventually, we should generalize and port the changes to the shared GitHub action workflows, but that would require additional work and review, which feels a bit overkill at this stage, as we are still validating this solution.


streamer45:
image

cpoile (approved):
Thanks!

PR #1027, opened 2025-05-23
[MM-64390] Safely access `window.opener` in cross-origin contexes

streamer45:
#### Summary

It's possible that the main Mattermost view already has `window.opener` set. This can happen when clicking a link to the MM instance that doesn't set [`noopener`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/rel/noopener) or [`noreferrer`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/rel/noreferrer), from a calendar app, for example (how we reproduced this).

When this happens, there are two problems:

1. `window.opener` won't be a good way to determine whether we are in the pop-up window anymore.
2. The opener could be on a different origin which would cause a security error when accessing it (i.e., `window.opener.callsClient` will throw). 

PR adds some guards to handle this case while also limiting the places that rely on `window.opener` directly.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-64390

### DIFF
--- a/webapp/src/utils.ts
+++ b/webapp/src/utils.ts
@@ -85,7 +85,7 @@ export function getChannelURL(state: GlobalState, channel?: Channel, teamId?: st
 export function getCallsClient(): CallsClient | undefined {
     let callsClient;
     try {
-        callsClient = window.opener ? window.opener.callsClient : window.callsClient;
+        callsClient = getCallsWindow().callsClient;
     } catch (err) {
         logErr(err);
     }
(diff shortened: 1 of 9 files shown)


webapp/src/utils.ts
```diff
@@ -108,19 +108,36 @@ export function isCallsPopOut(): boolean {
     try {
         return window.opener && window.opener.callsClient;
     } catch (err) {
-        logErr(err);
+        // This can happen if the MM window already has an opener on a different origin (e.g. user clicked a link on a calendar app to open MM).
+        // In this case, we know we are not in the expanded view so we can directly return window.callsClient.
+        logWarn(err);
```
streamer45, line 113: We are lowering the level to warning, but we'd still get quite the spam in such a case since this is used by a few components and selectors. I'll see if there's a way to detect the specific case and avoid logging, but I'm not sure.

cpoile: Something like: ?
```suggestion
        if (e.name !== 'SecurityError') {
            // This is likely not a cross-origin error.
            logErr(err);
  }
```

cpoile (approved):

DHaussermann (approved):
Tested and passed
- Confirmed JS console opens the `calls` URL. (I get directed to a page that says my team was not found but, that seems beside the point. The redirect worked.)
- Embedded the Calls URL in a G-Suite calendar event and confirmed the call launches
- Regression tested slash commands
- Regression tested switching calls by joining a call while on a call
LGTM!

PR #1037, opened 2025-06-04
[MM-64474] Fix e2e tests pipeline

streamer45:
#### Ticket Link

https://mattermost.atlassian.net/browse/MM-64474

### DIFF
--- a/e2e/scripts/run.sh
+++ b/e2e/scripts/run.sh
@@ -35,25 +35,28 @@ mkdir -p "${WORKSPACE}/results"
 echo "Generating sysadmin ..."
 docker exec \
 	${CONTAINER_SERVER}1 \
-	sh -c "/mattermost/bin/mmctl --local user create --email sysadmin@example.com --username sysadmin --password 'Sys@dmin-sample1'"
+	/mattermost/bin/mmctl user create --email-verified --email sysadmin@sample.mattermost.com --username sysadmin --password Sys@dmin-sample1 --system-admin --local
+
+# Copy admin password file
+docker cp e2e/scripts/pwd.txt ${CONTAINER_SERVER}1:/mattermost
 
 # Auth mmctl
 echo "Authenticating mmctl ..."
-docker exec \
+docker exec --env="XDG_CONFIG_HOME=/mattermost/config" \
 	${CONTAINER_SERVER}1 \
-	sh -c "echo 'Sys@dmin-sample1' > pwd.txt && /mattermost/bin/mmctl auth login http://localhost:8065 --username sysadmin --name local --password-file pwd.txt"
+	/mattermost/bin/mmctl auth login http://localhost:8065 --username sysadmin --name local --password-file /mattermost/pwd.txt
 
 # Install Playbooks
 echo "Installing playbooks ..."
-docker exec \
+docker exec --env="XDG_CONFIG_HOME=/mattermost/config" \
 	${CONTAINER_SERVER}1 \
-	sh -c "/mattermost/bin/mmctl plugin add /mattermost/prepackaged_plugins/mattermost-plugin-playbooks-v2*.tar.gz"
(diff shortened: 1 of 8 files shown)


webapp/src/components/incoming_calls/hooks.ts
```diff
@@ -236,13 +237,20 @@ export const useNotification = (call: IncomingCallNotification) => {
...
     const teammateNameDisplay = useSelector(getTeammateNameDisplaySetting);
     const caller = useSelector((state: GlobalState) => getUser(state, call.callerID));
     const currentUser = useSelector(getCurrentUser);
     const doGetProfilesInChannel = makeGetProfilesInChannel();
     const gmMembers = useSelector((state: GlobalState) => doGetProfilesInChannel(state, call.channelID));
     const callerName = displayUsername(caller, teammateNameDisplay, false);
 
+    useEffect(() => {
+        if (call.channelID) {
+            dispatch(getProfilesInChannel(call.channelID, 0));
```
streamer45, line 250: Not sure why this was needed all of the sudden, likely as a result of some changes on webapp but without it, `gmMembers` would be consistently empty.

webapp/src/components/incoming_calls/call_incoming_condensed.tsx
```diff
@@ -43,19 +43,28 @@ export const CallIncomingCondensed = ({call, onWidget = false, joinButtonBorder
         setShowTooltip(show);
     }, [messageRef]);
 
-    const message = (
+    let message = (
         <FormattedMessage
-            defaultMessage={'Call from <b>{callerName}</b> with <b>{others}</b>'}
```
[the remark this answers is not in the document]
cpoile: huh, yes definitely a mistake. We even do the check just below.

webapp/src/components/incoming_calls/call_incoming_condensed.tsx
```diff
@@ -80,13 +89,15 @@ export const CallIncomingCondensed = ({call, onWidget = false, joinButtonBorder
                     }
                     trigger={showTooltip ? ['hover', 'focus'] : []}
                 >
-                    <Message ref={messageRef}>
+                    <Message
+                        ref={messageRef}
+                    >
```
cpoile, line 94: Automatic formatting?

webapp/src/components/incoming_calls/hooks.ts
cpoile: hmm, interesting.
Nice that the tests caught it though!

cpoile (approved):

e2e/scripts/run.sh
```diff
@@ -65,15 +68,24 @@ docker cp e2e/config-patch.json ${CONTAINER_SERVER}1:/mattermost
...
 echo "Patching calls config ..."
-docker exec \
+docker exec --env="XDG_CONFIG_HOME=/mattermost/config" \
+	${CONTAINER_SERVER}1 \
+	/mattermost/bin/mmctl plugin disable com.mattermost.calls
+sleep 2
+docker exec --env="XDG_CONFIG_HOME=/mattermost/config" \
+	${CONTAINER_SERVER}1 \
+	/mattermost/bin/mmctl config patch /mattermost/config-patch.json
+sleep 2
```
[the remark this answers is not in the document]
streamer45: @cpoile The server Docker image recently changed to distroless (https://github.com/mattermost/mattermost/pull/31136) so the container has no shell which means we have to split everything and work with binaries directly since we can't chain commands anymore.

webapp/src/components/incoming_calls/call_incoming_condensed.tsx
streamer45: During the investigation (which took a while), I temporarily set `data-testid` for this element, which I later removed since the Playwright locator wasn't the issue. Formatting is likely a result of that.

PR #1039, opened 2025-06-06
Fetch config env overrides only if admin

streamer45:
#### Summary

Just noticed we'd make the call to fetch env config overrides from non-admin clients, which would fail with 403.


cpoile (approved):
yikes. 👍

PR #1045, opened 2025-06-10
[MM-64487] Improve error checking in load-test client signup logic

streamer45:
#### Summary

If SSO is enabled, the returned error ID is slightly different than expected (`api.user.login.invalid_credentials_sso` vs `api.user.login.invalid_credentials_email_username`), so we wouldn't try to register the users. Tweaking the check to cover all the potential "invalid credentials" errors.

@DHaussermann (not urgent) Could you give this a pass when you get a chance? 

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-64487

### DIFF
--- a/lt/client/client.go
+++ b/lt/client/client.go
@@ -14,6 +14,7 @@ import (
 	"log/slog"
 	"net/http"
 	"os"
+	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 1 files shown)


cpoile (approved):

DHaussermann (approved):
Tested and passed
- Confirmed that when oAuth is enabled the user creation and workflow and login works
- Regression tested no change when oAuth is disabled

LGTM!

PR #1052, opened 2025-06-17
[MM-61593] Attach icon and dropdownText for the calls button for mobile view

devinbinnie:
#### Summary
This attached an icon and dropdown text so that the Calls button shows up properly on mobile view.

Goes with https://github.com/mattermost/mattermost/pull/31654

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-61593

### DIFF
--- a/webapp/src/index.tsx
+++ b/webapp/src/index.tsx
@@ -16,7 +16,7 @@ import {getCurrentUserId, isCurrentUserSystemAdmin} from 'mattermost-redux/selec
 import {ActionFuncAsync} from 'mattermost-redux/types/actions';
 import React, {useEffect} from 'react';
 import ReactDOM from 'react-dom';
-import {injectIntl, IntlProvider} from 'react-intl';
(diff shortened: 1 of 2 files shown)


streamer45:
> > @devinbinnie Is the call icon supposed to render on the left of the label?
> 
> I'm not sure what's _supposed_ to happen, I think you can add it there though.

Okay, since I see the changes are passing the compass icon I thought we expected it to render somehow. Feels like a webapp side change though.

devinbinnie:
> > > @devinbinnie Is the call icon supposed to render on the left of the label?
> > 
> > 
> > I'm not sure what's _supposed_ to happen, I think you can add it there though.
> 
> Okay, since I see the changes are passing the compass icon I thought we expected it to render somehow. Feels like a webapp side change though.

It would actually be on the plugin side I think, you would just edit `dropdownText`. But I'm not sure if we want that there or not. And how it would be styled if we did pass it through.

streamer45:
Cool. Looks like this now:

image

streamer45 (approved):

PR #1055, opened 2025-06-24
[MM-64237] e2e: use proper CSRF header in requests

streamer45:
#### Summary

We'll soon deprecate cookie-authenticated API requests without a valid CSRF token, which would break the e2e pipeline completely.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-64237

### DIFF
--- a/e2e/utils.ts
+++ b/e2e/utils.ts
@@ -175,7 +175,14 @@ export function releaseLock(lockID: string) {
 }
 
 export async function getHTTPHeaders(context: APIRequestContext) {
+    const storageState = await context.storageState();
+    const csrfCookie = storageState.cookies.find(cookie => cookie.name === 'MMCSRF');
+
+    if (!csrfCookie) {
+        throw new Error('CSRF token cookie not found');
+    }
+
     return {
-        'X-CSRF-Token': (await context.storageState()).cookies[2].value,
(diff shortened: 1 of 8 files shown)


e2e/utils.ts
```diff
@@ -174,3 +173,9 @@ export function releaseLock(lockID: string) {
     // eslint-disable-next-line no-console
     console.log(test.info().title, 'lock released');
 }
+
+export async function getHTTPHeaders(context: APIRequestContext) {
+    return {
+        'X-CSRF-Token': (await context.storageState()).cookies[2].value,
```
cpoile, line 179: is there any chance this (using an index) will break?

streamer45: @cpoile That's for you to figure out. I did my part :p

It shouldn't be hard to convert this into a more robust "search" function.

cpoile (approved):

e2e/utils.ts
cpoile: equivalent to starting fires before turning the lights out. 😝

streamer45: @cpoile I could have left this ticket in the backlog, and in a couple of months, everything would have broken down due to this (v11)

cpoile: I was kidding, but I know, I appreciate it

PR #1058, opened 2025-06-26
Fix flaky e2e test

streamer45:
#### Summary

This has been flaky for a while, getting stuck on pressing `Enter` key. We take a different approach and use the `Save` button instead.


cpoile (approved):
Thanks!

PR #1064, opened 2025-07-03
[MM-64737] Nil session from plugin API's GetSession

cpoile:
#### Summary
- A customer provided logs showing a nil deference after calling `p.API.GetSession`. There was no `AppErr`, but apparently the session was nil. That shouldn't happen. But regardless of how or why, let's protect against that in the future.
- We don't have any other calls to `p.API.GetSession` to protect.
- Added a test which fails if the check is removed.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-64737

### DIFF
--- a/server/websocket.go
+++ b/server/websocket.go
@@ -504,7 +504,8 @@ func (p *Plugin) wsReader(us *session, authSessionID, handlerID string) {
 				continue
 			}
 
-			if s, appErr := p.API.GetSession(authSessionID); appErr != nil || (s.ExpiresAt != 0 && time.Now().UnixMilli() >= s.ExpiresAt) {
+			s, appErr := p.API.GetSession(authSessionID)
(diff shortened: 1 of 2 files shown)


server/websocket.go
```diff
@@ -514,7 +515,9 @@ func (p *Plugin) wsReader(us *session, authSessionID, handlerID string) {
 					us.connID,
 				}
 
-				if appErr != nil {
+				if appErr == nil && s == nil {
+					p.LogWarn("no apErr and no session found")
```
streamer45, line 519: Small typo. Also, we should add the `fields`.

streamer45: Yeah, logs can happen concurrently, I'd rather them being "self-contained".

streamer45 (approved):
Thanks!

PR #1075, opened 2025-07-30
Remove confusion on using FQDN as ICE Host Override

streamer45:
#### Summary

Most admins are unlikely to review our documentation notes. Additionally, it's improbable that it will function as they expect, leading to confusion and ultimately resulting in a failure to connect.

image


cpoile (approved):
Thanks, that's a good change.

PR #1123, opened 2026-01-29
Handle join_call links in webapp

bgardner8008:
#### Summary
Problem: Call links (/call link) don't work when clicked from within webapp - join_call parameter is read once at initialization and only checked on channel changes. These links only work when pasted into browser or clicked externally, causing new page load. See Jira ticket for more explanation.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-67348

### DIFF
--- a/webapp/src/components/join_call_watcher.tsx
+++ b/webapp/src/components/join_call_watcher.tsx
@@ -1,8 +1,9 @@
 // Copyright (c) 2020-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
+import {Channel} from '@mattermost/types/channels';
 import {GlobalState} from '@mattermost/types/store';
-import {getChannelsNameMapInCurrentTeam} from 'mattermost-redux/selectors/entities/channels';
+import {getChannel, getChannelsNameMapInCurrentTeam} from 'mattermost-redux/selectors/entities/channels';
 import {getCurrentTeam} from 'mattermost-redux/selectors/entities/teams';
 import {useEffect} from 'react';
 import {useSelector} from 'react-redux';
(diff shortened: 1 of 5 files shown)


cpoile (commented):
Thanks Bill! Just a couple questions: 
- could we add tests (sorry, but important for this kind of thing I think)
- it's been awhile, so I'm not sure if it applies, but does this play nicely with Desktop?

M-ZubairAhmed (approved):

webapp/src/index.tsx
hmhealey: I'm not the biggest fan of this approach since it 

Broadly speaking, I think there might be a better approach here where we handle this by using `registry.registerRootComponent` to mount an invisible component which uses React Router's `useLocation` hook to watch for the query parameter to be added. That approach is a lot more self-contained, takes advantage of the React lifecycle, and has the nice benefit of handling both the initial load case and whenever the URL changes.

```jsx
function JoinCallWatcher() {
    const location = useLocation();

    useEffect(() => {
        const params = new URLSearchParams(location.search);

        console.log('This is logged whenever anything in the URL changes', location, params);
    }, [location]);

    const joinCall = new URLSearchParams(location.search).get('join_call');
    useEffect(() => {
        console.log('This is logged on initial mount and whenever that parameter changes, but not on other URL changes', joinCall);
    }, [joinCall]);

    return null;
}
```

Ideally, you'd dispatch a Redux action at that point which could get the store state and see if it should join or change calls. The dispatched action could have access to the current channel ID and any of the other Calls plugin state like the channel of a current call. Even if we did that though, you'd still need to add handling for the channel ID not changing until after the `join_call` parameter is removed (like `pendingJoinChannelId` does currently in your code).

That may however be easier said than done with the way functions like `connectToCall` and a lot of `./utils` is currently written in way that isn't Redux-friendly because they pass around the Redux store singleton we get from the web app instead of being written as Thunk actions which can access the store from React much more easily. I'd be glad to help look into doing that more, but I'm fully aware doing that might be breaking some of Calls' code conventions even if those aren't general best practices

hmhealey (changes requested):
I've got some concerns about this way of doing this that I laid out below. Most of them are more about doing things in a more React-friendly way that I think would be easier to maintain in the long run (more encapsulated, easier to test, etc), but they seem to clash with how Calls generally interacts with the Redux store in somewhat non-standard ways. Ideally, I think we'd change that, but that definitely balloons the scope of these changes quite a bit.

My comments on `handleLinkClick` are focused on more things that could affect the user or other parts of the app, but I'm not entirely certain if they're actually things that can occur or not.

Let me know if you'd like to chat about any of this since it might be easier to hash this out synchronously. I'd like to help figure out what the right approach here is to balance maybe improving some of this without bogging you down with it

webapp/src/index.tsx
```diff
@@ -1115,9 +1125,84 @@ export default class Plugin {
...
+                return;
+            }
+
+            // Check if link contains join_call parameter
+            try {
+                const url = new URL(href, window.location.origin);
+                if (url.searchParams.get('join_call') === 'true') {
+                    // Extract channel ID from URL
+                    // URL format: /team-name/channels/channel-id?join_call=true
+                    const channelMatch = url.pathname.match(/\/channels\/([a-z0-9]+)/i);
```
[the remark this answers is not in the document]
bgardner8008: Good points about the other possible call links, I was unaware.

webapp/src/components/join_call_watcher.tsx
```diff
@@ -0,0 +1,66 @@
...
+            return;
+        }
+
+        const channel = channelsById[idOrName] || channelsByName[idOrName];
+        if (!channel) {
+            return;
+        }
+
+        onJoinCall(channel.id);
+    }, [joinCall, location.pathname, channelsById, channelsByName, currentTeamName, onJoinCall]);
```
hmhealey: This looks pretty good, but there's one thing to note: `onJoinCall` coule potentially be called multiple times if one of these other values changes before the query parameter is cleared.

I'm not sure if that's an issue with how `onJoinCall` works, but if that's at all an issue, we could probably rewrite this a bit to move the location access into a selector so that the effect only triggers when either the query parameter or channel change. I think it'd look like this:
```js
const JoinCallWatcher = ({onJoinCall}: Props) => {
    const location = useLocation();
    const joinCall = new URLSearchParams(location.search).get('join_call');

    const channelFromUrl = useSelector((state: GlobalState) => {
        const match = location.pathname.match(/^\/([^/]+)\/channels\/([^/]+)/);
        if (!match) {
            return undefined;
        }
        const [, teamName, idOrName] = match;
        if (!currentTeamName || teamName !== currentTeamName) {
            return undefined;
        }
    
        return getChannel(state, idOrName) || getChannelsNameMapInCurrentTeam(state)[idOrName];
    });

    useEffect(() => {
        if (joinCall !== 'true') {
            return;
        }
    
        if (!channelFromUrl) {
            return;
        }

        onJoinCall(channelFromUrl.id);
    }, [joinCall, channelFromUrl]);

    return null;
};
```

hmhealey (approved):

PR #1142, opened 2026-03-27
MM-67050: Pass through recorder and transcriber environment variables to job containers

bgardner8008:
## Summary
- Pass through `MM_CALLS_RECORDER_*` environment variables from the Mattermost server to recorder job containers (prefix stripped, keys lowercased)
- Pass through `MM_CALLS_TRANSCRIBER_*` environment variables to transcriber job containers (same convention)
- Extract `applyEnvOverrides()` helper to eliminate duplication between recorder and transcriber pass-through logic
- Unit tests covering prefix matching, key lowercasing, value preservation (including values with `=`), key collision behavior, and multiple variables

This enables administrators to configure self-signed/private CA TLS certificates and custom Chromium flags on recorder and transcriber containers via server-side environment variables.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-67050


go.mod
```diff
@@ -22,17 +22,19 @@ require (
...
 	github.com/testcontainers/testcontainers-go/modules/postgres v0.34.0
-	golang.org/x/time v0.7.0
+	golang.org/x/time v0.10.0
 	gopkg.in/yaml.v3 v3.0.1
 )
 
-replace github.com/pion/interceptor v0.1.40 => github.com/streamer45/interceptor v0.0.0-20250701195358-9e4ca8111c7a
+replace github.com/pion/interceptor v0.1.44 => github.com/bgardner8008/interceptor v0.1.44-mm-mods
+
+replace github.com/pion/ice/v4 => github.com/bgardner8008/ice/v4 v4.2.0-role-conflict-fix-v4
```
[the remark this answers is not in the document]
bgardner8008: Fixed

.github/workflows/ci.yml
M-ZubairAhmed, line 1: Lets remove CI dependency change from this PR.
1. lt/go.mod shouldn't drive CI Go version — it's a load-testing tool, not the main plugin. CI should derive its Go version from the root go.mod.
2. Version divergence gets worse — after this PR: root is 1.24.13, lt is 1.25.0, build is 1.23, server/public is 1.23.6. Four different versions across the repo.
3. If Go 1.25 is intended for the project, the root go.mod should be bumped too, and CI should still point at root go.mod as the source of truth.

I have a WIP PR which deals with the issue to bumping and uniforming all goversions

M-ZubairAhmed (changes requested):

.github/workflows/ci.yml
bgardner8008: Right, that was desperation on my part. Snyk complained about golang.org/x/net v0.50.0, blocking PR, requiring update to 0.51.0, but that forces go to 1.25 in lt/go.mod, and caused a ton of errors building with 1.24. The ci.yml golang-version: "1.24" doesn't seem to be used, hence I can't change it to 1.25 there, so I used the lt/go.mod.

OK, I guess I will rewind the x/net back to v0.50.0 and ask Eva to override snyk. It's still not clear to me why I didn't have same issue in the main plugin project.

M-ZubairAhmed (approved):

PR #1152, opened 2026-04-09
Add LiveKit config and token endpoint

bgardner8008:
#### Summary
Add LiveKit URL, API key, API secret, to configuration, and provide API endpoint to fetch an access token and URL.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-68263


go.mod
```diff
@@ -1,10 +1,10 @@
 module github.com/mattermost/mattermost-plugin-calls
 
-go 1.24.13
+go 1.25.0
```
M-ZubairAhmed, line 3: Does thisPR warrants a Go version change — we'd prefer to keep Go version upgrades in a separate, dedicated PR so they can be properly tracked, tested, and coordinated across the codebase.

M-ZubairAhmed (approved):

PR #1162, opened 2026-04-22
[MM-68428] Consolidate webapp clients under webapp/src/clients/ folder

M-ZubairAhmed:
#### Summary
This pull request reorganizes the client codebase by moving the callsclient, restclient, and websocketclient into a new src/clients directory. We're about to introduce a   livekit client as a fourth transport/client in the webapp. Today, the existing clients live scattered at the top level of webapp/src/. Dropping a fourth client into  that same flat layout would make src/ harder to navigate and would obscure the fact that these files share a distinct role, separate from app logic.

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-68428

#### Release Note

```release-note
NONE
```


bgardner8008 (approved):
LGTM, thanks. I'm assuming you expect LiveKit client to need multiple files, or perhaps you have a preference to isolate each client code in a separate directory.

M-ZubairAhmed:
> LGTM, thanks. I'm assuming you expect LiveKit client to need multiple files, or perhaps you have a preference to isolate each client code in a separate directory.

Yes both, i am expecting livekit client to grow along with preference to have clients being isolated.

PR #1166, opened 2026-04-27
[M-68304] Introduce LiveKit-backed CallClient foundation

M-ZubairAhmed:
#### Summary
**What this does**

Replaces the old custom WebRTC/signaling code with a new `CallClient` built on LiveKit. This is the foundation for the broader LiveKit/SIP migration.

**What works**
- Joining a call, mic permission prompt, mute/unmute, users hearing each other in a call.

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-68304

#### Release Note

```release-note

```


webapp/src/clients/call/call_client.ts
```diff
@@ -0,0 +1,261 @@
...
+
+        // Request microphone permission in the background so connection
+        // handling is not blocked by the user's interaction.
+        void this.requestMicrophonePermission();
+
+        this.emit(CALL_EVENT.CONNECTED);
+        logInfo('call client: connected to room');
+    }
+
+    private async requestMicrophonePermission() {
```
bgardner8008, line 50: Using setMicrophoneEnabled(true) will publish the track briefly. Better to call getUserMedia in order to get the permissions popup:

  private async requestMicrophonePermission() {                                                                                        
      try {                                                                                                                            
          const stream = await navigator.mediaDevices.getUserMedia({audio: true});
          stream.getTracks().forEach((t) => t.stop());                                                                                 
      } catch (err) {                                               
          logErr('call client: failed to request microphone permission', err);
          this.emit(CALL_EVENT.ERROR, err);                                                                                            
      }
  }

bgardner8008 (changes requested):
Please see my comment on requestMicrophonePermission.

The other issue is that all the client tests have been deleted. You could test the parts that are currently wired up, but if you prefer to wait until more is completed that's fine too.

webapp/src/clients/call/call_client.ts
M-ZubairAhmed: Thanks for the suggestion, added. And yes i will add tests later when i have more things in the client file

bgardner8008 (approved):
Claude is fussing about the dropped no-void rule. I find linters generally annoying so if this is your preference, fine with me, just wanted to note it.

The other thing to have on radar is that the initial mute state can depend on context. In general we'll start muted, but 1:1 DMs start unmuted. And we may have a pre-join dialog to set all this up. All good for now.

PR #1168, opened 2026-04-28
[MM-68545] Centralize Go version management with .go-version file

M-ZubairAhmed:
#### Summary
This pull request standardizes and updates the Go version used across the project to 1.25.0, ensuring consistency between development, CI, and documentation. The workflow files are updated to use a single source of truth for the Go version via a new `.go-version` file. 

#### Ticket Link

Fixes https://mattermost.atlassian.net/browse/MM-68545

#### Release Note

```release-note

```


bgardner8008 (approved):
This all looks good. An alternative would be to use a newer (master) version of the plugin-ci workflow that directly supports using the go-version-file. But I don't know if there are other changes in the workflow that will impact things. We can also pick it up later.

bgardner8008:
I assume we'll also have to be careful to update both go.mod(s) and .go-version to use the same version.

M-ZubairAhmed:
> This all looks good. An alternative would be to use a newer (master) version of the plugin-ci workflow that directly supports using the go-version-file. But I don't know if there are other changes in the workflow that will impact things. We can also pick it up later.

Picked this up in this pr bumped to latest on main and switched to go-version-file


> I assume we'll also have to be careful to update both go.mod(s) and .go-version to use the same version.

Good point. Right now they have to be kept in sync manually across. May be can add a version drift check in CI which checks for versions across.

PR #1172, opened 2026-05-05
[MM-68305] [MM-68556] Implement mute, unmute, subscribe & handle participants join, left, count, info

M-ZubairAhmed:
#### Summary
What this does

  Implements MM-68305 (mic publish + mute/unmute) and MM-68556 (subscribe to remote audio) on the new LiveKit CallClient. While we were here, also bundled the same
  architectural cutover for mute state and participant tracking — both now driven entirely by LiveKit events instead of plugin WebSocket.

  Mic toggles publish/mute the LiveKit track; remote audio is subscribed and shipped to the existing widget for <audio> playback. Mute state and participant join/leave both
  flow through RoomEvent.* handlers in CallClient, which emit to index.tsx via custom events. Same code path handles self and remote — (session_id, userID) come from
  participant.sid / participant.identity, so one Redux dispatcher works for everyone.

  What works

  - Two browsers in same channel hear each other after both unmute
  - Self + remote mute toggles update mute icons on every client
  - Widget participant count + list correctly reflect everyone in the call (snapshot at connect + live join/leave events)
  - join_self plays once on your own join; join_user plays only for live remote joins (silent for participants who were already in when you joined)
  - Late join / pop-out window attaches remote audio
  - Mic permission denial and mid-call device loss both show the error modal

#### Ticket Link

Fixes https://mattermost.atlassian.net/browse/MM-68305
Fixes https://mattermost.atlassian.net/browse/MM-68556

#### Release Note

```release-note

```

### DIFF
--- a/webapp/src/clients/call/call_client.ts
+++ b/webapp/src/clients/call/call_client.ts
@@ -3,7 +3,20 @@
 
 import type {EmojiData} from '@mattermost/calls-common/lib/types';
 import {EventEmitter} from 'events';
-import {ConnectionState, DisconnectReason, LocalTrackPublication, RemoteParticipant, RemoteTrack, RemoteTrackPublication, Room, RoomEvent, Track} from 'livekit-client';
+import {
+    ConnectionState,
+    DisconnectReason,
+    LocalParticipant,
+    LocalTrackPublication,
+    Participant,
+    RemoteParticipant,
+    RemoteTrack,
+    RemoteTrackPublication,
+    Room,
+    RoomEvent,
+    Track,
+    TrackPublication,
+} from 'livekit-client';
 import RestClient from 'src/clients/rest';
 import {CALL_EVENT, CALL_TOKEN_API_PATH} from 'src/constants';
 import {logDebug, logErr, logInfo} from 'src/log';
@@ -15,8 +28,12 @@ export type RtcTokenResponse = {
     url: string;
 };
 
-export async function fetchRtcToken(channelID: string): Promise<RtcTokenResponse> {
-    const url = `${getPluginPath()}/${CALL_TOKEN_API_PATH}?channel_id=${encodeURIComponent(channelID)}`;
+export async function fetchRtcToken(
+    channelID: string,
+): Promise<RtcTokenResponse> {
+    const url = `${getPluginPath()}/${CALL_TOKEN_API_PATH}?channel_id=${encodeURIComponent(
+        channelID,
+    )}`;
     return RestClient.fetch<RtcTokenResponse>(url, {method: 'GET'});
 }
 
@@ -25,7 +42,6 @@ export default class CallClient extends EventEmitter {
     public initTime = 0;
     public room: Room | null = null;
 
-    // Stub props — real values land in follow-up PRs (audio/video/screen).
     public audioTrack: MediaStreamTrack | null = null;
     public localVideoStream: MediaStream | null = null;
     public currentAudioInputDevice: MediaDeviceInfo | null = null;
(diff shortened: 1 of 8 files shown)


webapp/src/clients/call/call_client.ts
```diff
@@ -35,14 +47,28 @@ export default class CallClient extends EventEmitter {
...
         // Request microphone permission in the background so connection
         // handling is not blocked by the user's interaction.
         void this.requestMicrophonePermission();
 
-        // User starts muted.
-        this.emit(CALL_EVENT.MUTE);
+        // Seed the initial state for everyone already in the room (local + remote)
+        const localParticipant = this.room.localParticipant;
+        const isLocalMicMuted = localParticipant.getTrackPublication(Track.Source.Microphone)?.isMuted ?? true;
+        this.emit(isLocalMicMuted ? CALL_EVENT.MUTE : CALL_EVENT.UNMUTE, localParticipant.sid, localParticipant.identity);
```
bgardner8008, line 65: The MUTE event needs to be emitted after the USER_JOINED event, because the USER_JOINED event handler sets unmuted: false.

webapp/src/index.tsx
```diff
@@ -745,24 +728,57 @@ export default class Plugin {
...
+                    });
+                });
+
+                window.callsClient.on(CALL_EVENT.USER_LEFT, (session_id: string, userID: string) => {
+                    store.dispatch({
+                        type: USER_LEFT,
+                        data: {
+                            channelID: window.callsClient?.channelID,
+                            userID,
+                            session_id,
```
bgardner8008: The previous websocket logic userLeft also checked for last person in room leaving and would trigger callEnd. The new logic doesn't do this but should.

webapp/src/index.tsx
```diff
@@ -745,24 +728,57 @@ export default class Plugin {
...
+                    store.dispatch(loadProfilesByIdsIfMissing([userID]));
+                    store.dispatch({
+                        type: USER_JOINED,
+                        data: {
+                            channelID: window.callsClient?.channelID,
+                            userID,
+                            currentUserID,
+                            session_id,
+                        },
+                    });
```
bgardner8008: The websocket logic for handleUserJoined also 1) cleared the incoming call notification and 2) stopped ringing when the current user joined. New logic is missing these.

Also missing is the USER_JOINED_TIMEOUT. This means any user joined notifications in the widget will be persistent.

bgardner8008 (changes requested):
Overall looks great. Claude pointed out a few bugs, probably best to address these on each PR so things stay solid. Also note that we need tests at some point. Perhaps we should make a separate ticket for client tests? Or you can add them for the functions that are implemented in each PR.

webapp/src/clients/call/call_client.ts
M-ZubairAhmed: Good catch, fixed.

webapp/src/index.tsx
M-ZubairAhmed: Fixed.

webapp/src/index.tsx
M-ZubairAhmed: All three pieces fixed

bgardner8008 (approved):
Thanks this looks good. One final point that claude is reminding me about: The standalone pop-out window (standalone/src/widget/index.tsx) registers mute/unmute handlers on callsClient but has no handlers for CALL_EVENT.USER_JOINED or CALL_EVENT.USER_LEFT. I'm assuming you'll revisit the popout window in the later ticket.

PR #1173, opened 2026-05-07
[MM-68698] LiveKit webhook handler, SIP participant tracking, and host logic

bgardner8008:
## Summary

- Add `POST /livekit-webhook` endpoint that verifies LiveKit JWT signatures (HMAC-SHA256 via `LiveKitAPISecret`) and handles `participant_joined`/`participant_left` events for SIP participants only; non-SIP events are silently ignored
- Add `IsSIPParticipant bool` to `CallSession` struct and DB migration 006 to persist it
- Rewrite `getHostID()` to deprioritize SIP participants: non-SIP users always preferred as host; SIP-only calls fall back to earliest SIP joiner; `HostLockedUserID` retains top priority
- Webhook route bypasses Mattermost-User-Id auth middleware and rate limiter (LiveKit server has no user session)
- Regular (non-SIP) clients continue to use the v1 Mattermost WS join/leave path unchanged — this is a purely additive change for SIP

## Test plan

- [ ] `go test ./server/... ` — all pass including 14 new tests in `api_livekit_webhook_test.go`
- [ ] `TestHandleLiveKitWebhook` — signature verification (valid/invalid/unconfigured), auth middleware bypass, non-SIP events ignored
- [ ] `TestHandleLiveKitSIPParticipant` — end-to-end with real PostgreSQL: join creates session row with `IsSIPParticipant: true` and publishes `user_joined`; no active call is a no-op; leave removes session and publishes `user_left`; leave is idempotent
- [ ] `TestGetHostIDSIPDeprioritization` — 6 unit tests covering all host priority tiers
- [ ] `make golangci-lint` — no new issues (pre-existing `interfaces` package-name warning unchanged from `main`)
- [ ] `make gomod-check` — passes after `go mod tidy`

Jira: https://mattermost.atlassian.net/browse/MM-68698

### DIFF
--- a/server/api.go
+++ b/server/api.go
@@ -23,6 +23,8 @@ import (
 
 	"github.com/gorilla/mux"
 	"github.com/livekit/protocol/auth"
+	"github.com/livekit/protocol/livekit"
+	"github.com/livekit/protocol/webhook"
 )
 
 const requestBodyMaxSizeBytes = 1024 * 1024 // 1MB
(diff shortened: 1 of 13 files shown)


bgardner8008:
> This pull request contains a high-severity issue where the SIP webhook handler trusts an untrusted participant.Identity from signaling metadata, allowing an attacker to spoof a legitimate Mattermost user's ID.

The concern about impersonating a Mattermost user ID doesn't apply here. For SIP participants, `participant.Identity` is the PSTN caller ID (phone number, e.g. `+14155551234`), which is set by LiveKit from the SIP `From` header — not a Mattermost user ID. Phone numbers are not valid Mattermost user IDs and are never looked up in the Mattermost user store. The `IsSIPParticipant: true` flag on the session tells the UI to render these participants differently (displaying the caller ID directly rather than resolving a Mattermost profile), so there is no path to impersonate a real Mattermost user through this mechanism.

Caller ID spoofing at the PSTN level is possible in principle, but is mitigated by the PIN code requirement in the LiveKit SIP dispatch rule: a caller must enter a lengthy PIN to be admitted to the room at all. Combined with the webhook signature verification (HMAC-SHA256 via `LiveKitAPISecret`), which ensures the webhook payload genuinely originates from our LiveKit server, the trust model is appropriate for SIP participants.

server/api.go
```diff
@@ -591,3 +593,188 @@ func (p *Plugin) handleGetStats(w http.ResponseWriter) error {
...
+			"channelID", channelID, "err", err.Error())
+	}
+
+	p.publishWebSocketEvent(wsEventUserJoined, map[string]interface{}{
+		"user_id":    identity,
+		"session_id": sid,
+	}, &WebSocketBroadcast{ChannelID: channelID, ReliableClusterSend: true})
+}
+
+func (p *Plugin) handleLiveKitParticipantLeft(event *livekit.WebhookEvent) {
```
M-ZubairAhmed: So this is purely for SIP participant leave the room right? If yes then may be we should explicitly say SIP participant

bgardner8008: Yes, web hooks are the only way the plugin can know about SIP participants joining or leaving. If the web hooks are non-functional the SIP participants will be heard but will not show up as a participant, a rather poor user experience. Similarly, we would also need to rely on web hooks for external guests joining via webapp, those users don't have a websocket connection to mattermost, but external guests is not currently in scope.

bgardner8008: And to support external webapp guests this logic will have to change; they are neither SIP participant nor are they mattermost participant, it would be another designation.

bgardner8008: Just reread your comment and updated the code to use new "SIP" function names. In the future if we ever support external guest users these could be updated to "External" encompassing both SIP and external browser users.

M-ZubairAhmed (approved):

PR #1174, opened 2026-05-07
[MM-68307] [MM-68558] [MM-68699] Add voice activity detection & audio devices along with PluginWS client connection

M-ZubairAhmed:
#### Summary

- LiveKit-driven voice activity detection.
- Audio device management
- `Room` and `WebSocketClient` instances + all event handlers are wired up at construction.
- Plugin WS plumbing reintroduced to `CallClient` to send `'join'` / `'leave'` `'reconnect'` — required so the server creates the call record and broadcasts `call_start` (which carries `thread_id`).
- sessions reducer extracted into `state/session/`

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-68307
https://mattermost.atlassian.net/browse/MM-68558
https://mattermost.atlassian.net/browse/MM-68699

Merged Ticket https://mattermost.atlassian.net/browse/MM-68719 in this PR

#### Release Note

```release-note

```


webapp/src/clients/websocket/index.ts
```diff
@@ -48,12 +48,15 @@ export class WebSocketClient extends EventEmitter {
 
     constructor(wsURL: string, authToken?: string) {
         super();
+
         this.wsURL = wsURL;
         this.authToken = authToken || '';
-        this.init(false);
     }
```
bgardner8008, line 54: I like the change of splitting construction and connection into separate phases, it's much better.

webapp/src/reducers.ts
```diff
@@ -308,6 +312,39 @@ const sessions = (state: sessionsState = {}, action: sessionsAction) => {
...
+            // If the voice flag already matches the new speaking state, reuse the existing session object reference.
+            if (channel[session_id].voice === isSessionInActiveSpeakersList) {
+                nextState[session_id] = channel[session_id];
+            } else {
+                // Else, voice flag flipped — create a new session object with the updated value.
+                nextState[session_id] = {...channel[session_id], voice: isSessionInActiveSpeakersList};
+                stateChanged = true;
+            }
+        }
+
```
bgardner8008: claude suggested an optimization in the reducer loop. Instead of using .includes() inside the loop which is O(n * m), move the active speakers test outside of the loop:

      // Build the Set ONCE, outside the loop.
      const activeSpeakers = new Set(action.data.session_ids);

And then in the loop:

          const isSessionInActiveSpeakersList = activeSpeakers.has(session_id);  // O(1) lookup

Not sure, even in a large call with 100s of users I suspect the linear searches still go fast.

webapp/src/reducers.ts
```diff
@@ -308,6 +312,39 @@ const sessions = (state: sessionsState = {}, action: sessionsAction) => {
...
+        if (!stateChanged) {
+            return state;
+        }
+
+        return {
+            ...state,
+            [action.data.channelID]: nextState,
+        };
+    }
```
bgardner8008: USER_VOICE_ON / USER_VOICE_OFF cases are now dead code.

webapp/src/reducers.ts
```diff
@@ -308,6 +312,39 @@ const sessions = (state: sessionsState = {}, action: sessionsAction) => {
                 },
             },
         };
```
bgardner8008: Regarding USERS_VOICE_ACTIVITY_CHANGED

currently user_ids is not used in the reducer, but after identity change we will be using this instead of session_ids. At that time there is no reason to pass around the session_ids which are LK Participant.Sid.

Currently broken until we make identity change, VAD will only work for local user and not remotes.

webapp/src/clients/call/call_client.ts
```diff
@@ -34,17 +46,394 @@ export async function fetchRtcToken(channelID: string): Promise<RtcTokenResponse
...
+            // and (if we were the last participant) call_end.
+            this.websocketClient.sendLeave();
+            this.websocketClient.close();
+            this.websocketClient = null;
+        }
+    }
+
+    public destroy() {
+        this.disconnect();
+    }
```
bgardner8008: This will partially run disconnect up to first await, any subsequent failures will lead to unhandled promise rejection and there is no catch to log errors. You could add catch

this.disconnect().catch((err) => logErr('CallClient: destroy disconnect failed', err));

You could also make destroy async and then await it from caller.

Somehow this needs to be improved so teardown is orderly.

webapp/src/clients/call/call_client.ts
```diff
@@ -34,17 +46,394 @@ export async function fetchRtcToken(channelID: string): Promise<RtcTokenResponse
...
+     * participant we use the WS originalConnID (matches what the server broadcasts);
+     * for REMOTE participants we use LiveKit's sid (best we have until MM-68557).
+     */
+    private sessionIDForParticipant(p: Participant): string {
+        if (this.room && p === this.room.localParticipant) {
+            return this.websocketClient?.getOriginalConnID() ?? p.sid;
+        }
+        return p.sid;
+    }
+
```
bgardner8008, line 376: Right, we need to fix the identity issue by setting LK Participant.identity to our sessionID (same as websocket connection ID).

Should we make the change in this PR or separate?

bgardner8008 (changes requested):
Thanks, looks great. Two important things:
1) we need to change/fix the identity mechanism, either in this PR, or a subsequent one, and note it here.
2) please improve the destroy logic.

webapp/src/reducers.ts
M-ZubairAhmed: Worst case for us (~200-person call, 10 concurrent speakers) is ~2K string compares per event, and LiveKit only fires this on change. Sub-ms either way. So this could be premature optimization. Happy to revisit later if we find the problem.

webapp/src/reducers.ts
M-ZubairAhmed: Yes this is removed now.

webapp/src/clients/call/call_client.ts
M-ZubairAhmed: ~Made destroy() async with an internal try/await this.disconnect()/catch that logs — so it can never produce an unhandled rejection regardless of caller~

webapp/src/reducers.ts
M-ZubairAhmed: Yes this is fixed now.

webapp/src/clients/call/call_client.ts
M-ZubairAhmed: Fixed

webapp/src/clients/call/call_client.ts
M-ZubairAhmed: consolidated destroy() into disconnect()

M-ZubairAhmed:
@coderabbitai review

webapp/src/clients/call/call_client.ts
```diff
@@ -17,34 +20,415 @@ import {
...
+            this.connectPayload = null;
+            this.room = null;
+
+            this.emit(CALL_EVENT.ERROR, err);
+            throw err;
+        }
+
+        // Hydrate the device cache + restore the user's last selection.
+        await this.restoreAudioDevicesFromStorage();
+    }
```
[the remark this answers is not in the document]
M-ZubairAhmed: the called function already handles its own errors at each async boundary — enumerateDevices,
  setAudioInputDevice both wrap their externally-called code. The await here is safe

standalone/src/widget/index.tsx
```diff
@@ -59,30 +52,16 @@ async function initWidget({store, startingCall}: InitCbProps) {
...
-        store.dispatch({
-            type: USER_UNMUTED,
-            data: {
-                channelID: window.callsClient?.channelID,
-                userID: getCurrentUserId(store.getState()),
-                session_id: window.callsClient?.getSessionID(),
-            },
-        });
+        store.dispatch(userUnmuted(window.callsClient?.channelID ?? '', getCurrentUserId(store.getState()), window.callsClient?.getSessionID() ?? ''));
     });
```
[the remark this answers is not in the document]
M-ZubairAhmed: done

M-ZubairAhmed:
@coderabbitai review

webapp/src/reducers.ts
bgardner8008: Agreed, premature optimization.

bgardner8008 (approved):
LGTM, thanks!

PR #1175, opened 2026-05-11
[MM-68719] Use user_id & ession_id for LiveKit identity

bgardner8008:
## Summary

Stacks on #1174. Server now issues LiveKit tokens with `Identity = plugin-WS session_id` (originalConnID) and `metadata.user_id = userID`, instead of `Identity = userID`. This:

- Removes the local-vs-remote session_id asymmetry #1174 introduced (`sessionIDForParticipant`, `getSessionID` fallback) — session_id is now a single canonical value across plugin-WS broadcasts and LiveKit events.
- Lets one Mattermost user hold multiple sessions in a call (LiveKit identity uniqueness no longer kicks duplicates).
- Fixes remote-participant VAD: active-speaker session_ids (LK identity) now match reducer-state keys.

## Server

`GET /livekit-token` now requires `session_id` and rejects it (403) unless the session exists in `p.sessions` and belongs to the requesting user — without this, a client could mint a token claiming any session_id. Validation uses the existing `getSessionByOriginalID` helper so a WS reconnect mid-call still works.

## Client

`fetchRtcToken` passes `session_id` from the connID returned by `websocketClient.ready()`. Removed the temporary workarounds in `call_client.ts`; `getSessionID()` returns `localParticipant.identity` and a new `userIDFromMetadata()` helper parses `participant.metadata`.

## Notes

- Targets #1174's branch; squash-merging this PR will fold both changes into that PR's diff.
- SIP webhook path (`handleLiveKitParticipantJoined`) is unaffected — SIP participants have no plugin-WS session and continue using `participant.Sid` as session_id.

## Test plan

- [x] Server: `TestHandleGetLiveKitToken` covers missing session_id, not-found, owned-by-different-user, reconnect (originalConnID match), success (identity == sessionID, metadata.user_id == userID).
- [x] Webapp: full jest suite passes (315/315); call_client tests updated to flow new (identity, user_id-from-metadata) pairs through every emit.
- [ ] Manual: start a call from two tabs as the same user — both sessions should stay connected (no LiveKit identity-collision kick).
- [ ] Manual: confirm remote-participant voice activity indicator lights up.

### DIFF
--- a/server/api.go
+++ b/server/api.go
@@ -497,23 +497,16 @@ func (p *Plugin) handleGetLiveKitToken(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	// userID rides in metadata so other clients can map identity (session_id)
-	// back to a Mattermost user. LiveKit relays metadata to all participants.
-	metadata, err := json.Marshal(map[string]string{"user_id": userID})
-	if err != nil {
-		res.Err = fmt.Sprintf("failed to encode metadata: %s", err.Error())
-		res.Code = http.StatusInternalServerError
-		return
-	}
-
+	// Identity is the only LiveKit field sealed post-connect, so userID and
+	// sessionID both ride in it. The "___" separator both delimits the two
+	// halves and tags the format as an MM user (vs. "guest:" / "sip:" kinds).
 	at := auth.NewAccessToken(cfg.LiveKitAPIKey, cfg.LiveKitAPISecret)
 	grant := &auth.VideoGrant{
 		RoomJoin: true,
 		Room:     channelID,
 	}
 	at.SetVideoGrant(grant).
-		SetIdentity(sessionID).
-		SetMetadata(string(metadata)).
+		SetIdentity(userID + "___" + sessionID).
 		SetName(user.GetDisplayName(model.ShowFullName)).
 		SetValidFor(time.Hour)
 
--- a/server/api.go
+++ b/server/api.go
@@ -459,7 +459,24 @@ func (p *Plugin) handleGetLiveKitToken(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	sessionID := r.URL.Query().Get("session_id")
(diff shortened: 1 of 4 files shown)


server/api.go
```diff
@@ -480,13 +497,23 @@ func (p *Plugin) handleGetLiveKitToken(w http.ResponseWriter, r *http.Request) {
...
+
 	at := auth.NewAccessToken(cfg.LiveKitAPIKey, cfg.LiveKitAPISecret)
 	grant := &auth.VideoGrant{
 		RoomJoin: true,
 		Room:     channelID,
 	}
 	at.SetVideoGrant(grant).
-		SetIdentity(userID).
+		SetIdentity(sessionID).
+		SetMetadata(string(metadata)).
```
M-ZubairAhmed: The user_id-in-metadata approach works, but participant.setMetadata is callable from the client (we expose CallClient on window, plus extensions/devtools), so user_id can be vandalized at runtime. Identity is the only LiveKit field that's actually sealed post-connect.
The server-side cost is one extra map lookup when constructing identity for host-control LiveKit calls, since we already have sessions[id].userID on hand. 
I was thinking identity to be something as `userID___sessionID` ? Thoughts

M-ZubairAhmed (commented):
Open for discussion

server/api.go
bgardner8008: Nice find, I didn't think of that. I like your approach bundling userID and sessionID, plus it brings identity more inline with userID. I will update.

bgardner8008:
I've updated to use "<user_id>___<session_id>", the triple underscores will indicate this style of identity. Note that when SIP users are implemented, their identity will be "sip:<phone_number>", and the PoC for "guest" (external) users is using "guest:<session_id>". The SIP identity is baked into SIP bridge, we control the guest identity if we ever implement that.

M-ZubairAhmed (approved):

PR #1176, opened 2026-05-12
[MM-68330] Adapt host controls to LiveKit, stub recording/transcription

bgardner8008:
## Summary
- Host mute/mute-others call LiveKit `MutePublishedTrack` on the microphone track so the mic is force-muted server-side; user can still self-unmute, matching v1 semantics. `hostRemoveSession` calls `RemoveParticipant` in place of the no-op `closeRTCSession`. WS event broadcasting, DB updates, and permission checks unchanged.
- Recording and transcription start/stop are gated at their entry points with a "not yet supported" 501 return so `jobService.RunJob` is never invoked. Orchestration logic, `job_service.go`, and `bot_api.go` are left intact for restoration when the recorder/transcriber move to LiveKit.
- Adds `github.com/livekit/server-sdk-go/v2 v2.12.8` and a small `livekit_admin.go` helper (`getLiveKitRoomClient`, `livekitMuteParticipant`, `livekitRemoveParticipant`). Pinned to v2.12.8 to avoid bumping `pion/webrtc`.

## Notes
- The LiveKit identity used in these calls is the user ID, matching v2's current token issuance. When MM-68719 lands and switches the identity to a `userID___sessionID` composite, callers in `host_controls.go` need to pass that composite instead of `ust.UserID`.
- No `host_controls_test.go` exists on v2 — host controls are untested at the unit level. Worth a follow-up ticket for permission/precondition coverage; the LK enforcement itself needs E2E.

## Test plan
- [ ] `go build ./server/...`, `go vet ./server/...`, `make gomod-check` pass
- [ ] `go test ./server/` passes (TestStatsStore failure on v2 is pre-existing, not caused by this PR)
- [ ] Manual: with LiveKit configured, host mutes a participant → participant's mic stops on server; participant can click unmute to re-enable
- [ ] Manual: host removes a participant → after 3s grace, LK kicks them from the room
- [ ] Manual: `POST /calls/{id}/recording/start` → 501 "recording is not yet supported on this version"


server/livekit_admin.go
```diff
@@ -0,0 +1,82 @@
...
+	if lkURL == "" || cfg.LiveKitAPIKey == "" || cfg.LiveKitAPISecret == "" {
+		return nil, errLiveKitNotConfigured
+	}
+	return lksdk.NewRoomServiceClient(lkURL, cfg.LiveKitAPIKey, cfg.LiveKitAPISecret), nil
+}
+
+// livekitMuteParticipant force-mutes the participant's microphone track(s) on
+// the server. The user can still unmute themselves locally afterwards, matching
+// v1 host-mute semantics.
+func (p *Plugin) livekitMuteParticipant(room, identity string) error {
```
M-ZubairAhmed, line 38: Should we not do plugin DB update for muted participants or thats something we dont track in the plugin DB? Curious

server/host_controls.go
M-ZubairAhmed: Confusing name 
- is this for host removing the participant? or 
- host removing the complete session/room?

server/host_controls.go
```diff
@@ -259,8 +269,9 @@ func (p *Plugin) hostRemoveSession(requesterID, channelID, sessionID string) err
 			return
 		}
 
-		if err := p.closeRTCSession(ust.UserID, sessionID, channelID); err != nil {
-			p.LogError("hostRemoveSession: failed to close RTC session", "err", err.Error())
+		if err := p.livekitRemoveParticipant(channelID, ust.UserID); err != nil && !errors.Is(err, errLiveKitNotConfigured) {
```
M-ZubairAhmed: IF its session/room then shouldn't be just delete the room here

server/livekit_admin.go
bgardner8008: Mute state lives in LiveKit. It's best not to duplicate state unless we absolutely have to. And if we do track state in plugin, we need to use the DB to make sure state is synced in HA (if it depends on REST API, websocket accessed state can be local to server).

server/host_controls.go
bgardner8008: Good point. I'll rename to hostRemoveParticipant.

M-ZubairAhmed (changes requested):
The v2 merge introduced SetIdentity(userID + "___" + sessionID) in api.go (identity is sealed post-connect per the comment there), but here we pass UserID only. LiveKit's GetParticipant / RemoveParticipant will return NotFound, so host mute/remove will appear to work from the UI (WS event fires, local state updates) but won't actually touch LiveKit.

I will take this over.

M-ZubairAhmed (approved):

PR #1181, opened 2026-05-15
[MM-68563] Replace RTCMonitor with LiveKit ConnectionQuality

bgardner8008:
#### Summary
CallClient now subscribes to RoomEvent.ConnectionQualityChanged and emits a new CALL_EVENT.QUALITY_CHANGED for the local participant. The widget and expanded view consume that event in place of the v1 `mos` listener: Poor/Lost shows the degradedCallQuality banner, Excellent/Good clears it, Unknown is a no-op. Dismiss + lockout behavior is unchanged.

Drops the now-dead RTCMonitor wiring from the v1 CallsClient.

#### Note for reviewers
Semantic shift: v1's MOS estimated perceived audio quality client-side, so the banner only fired once degradation became audible. LiveKit's `ConnectionQuality` grades the network path, so the banner fires earlier — sometimes while Opus FEC is still keeping audio intelligible. This is intentional as an early warning.

#### Test plan
See the ticket for the local LAN + Network Link Conditioner test setup (attached `LiveKitLocalTlsDev.md`).

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-68563

### DIFF
--- a/webapp/src/clients/call/call_client.test.ts
+++ b/webapp/src/clients/call/call_client.test.ts
@@ -1,7 +1,7 @@
 // Copyright (c) 2020-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-import {Room, RoomEvent, Track} from 'livekit-client';
+import {ConnectionQuality, Room, RoomEvent, Track} from 'livekit-client';
 import RestClient from 'src/clients/rest';
 import {WebSocketClient} from 'src/clients/websocket';
 import {WEBSOCKET_EVENT} from 'src/clients/websocket/constants';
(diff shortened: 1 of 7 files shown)


webapp/src/clients/call/call_client.test.ts
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) 2020-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-import {Room, RoomEvent, Track} from 'livekit-client';
+import {ConnectionQuality, Room, RoomEvent, Track} from 'livekit-client';
```
M-ZubairAhmed, line 4: Little bit concerned that livekit leaks into the implementation

M-ZubairAhmed (approved):

PR #1183, opened 2026-05-18
[MM-68855] Calls widget 'Leave call' requires two clicks after closing the expanded view popout

M-ZubairAhmed:
#### Summary
* Upgraded `@mattermost/desktop-api` from version `5.10.0-1` to `6.2.0-1`
* Improved window closing logic in `CallWidget` by checking if `expandedViewWindow` is not already closed before attempting to close it, preventing potential errors.

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-68855

#### Release Note

```release-note

```


webapp/.eslintrc.json
```diff
@@ -355,7 +355,6 @@
       }
     ],
     "no-undef-init": 2,
-    "no-undefined": 2,
```
lieut-data, line 358: Why the change to this linting rule?

webapp/package-lock.json
```diff
@@ -37,7 +37,7 @@
         "@babel/preset-typescript": "7.16.0",
         "@formatjs/cli": "5.0.7",
         "@mattermost/client": "file:mattermost-webapp/webapp/platform/client",
-        "@mattermost/desktop-api": "5.10.0-1",
+        "@mattermost/desktop-api": "6.2.0-1",
```
lieut-data, line 40: To confirm, is the desktop-api compatible with old desktop versions?

webapp/.eslintrc.json
M-ZubairAhmed: Removed beacuse no-undefined is generally considered overly strict, initializing with undefined is perfectly valid in modern TypeScript.

webapp/package-lock.json
M-ZubairAhmed: The bump from 5.10.0-1 → 6.2.0-1 is type-only. It doesn't affect the runtime bundle at all. All actual desktop API calls in the plugin have legacy fallbacks, so nothing breaks on older desktop versions. The v5.x desktop line is fully supported.

lieut-data (approved):

M-ZubairAhmed:
Heads up @DHaussermann  This PR will be included in the 1.12 release

DHaussermann:
@M-ZubairAhmed Thanks for bringing this to my attention.

I can test this shortly and the we can merge and see about creating an 1.12.0-RC6 early in the week.

DHaussermann:
@M-ZubairAhmed I see an issue with this build.
I'm not sure the 2x leave issue is resolved. On desktop only (not Chrome) I see an issue with this build where join in not working. 
The join in the top right just hangs.   
image

Joining from a post is alos non-functional.
In both cases I see
image

I have tried clean rebuilds including removing _node_modules_ to avoid any deployment problems but I still see this issue. It's possible I'm having a deployment issue. Please advise if you can repro.
- Build the plugin with clean web
- Upload to test server and enable
- Start a call from desktop
Observed join does not worl 

Calls build `1.12.0-rc5+80b40306)`
Desktop `6.3.0 (257150344001)`


This does **not** repro for me on v1.12.0-rc5

M-ZubairAhmed:
Thanks @DHaussermann, looking into this now.

This PR doesn't actually touch the join-call path — connectCall → desktopAPI.joinCall → handleDesktopJoinedCall is byte-for-byte identical to rc5. The only changes are a TYPE ONLY bump of @mattermost/desktop-api, the CallWidget popout state→ref refactor (only runs after joining), an ESLint rule removal. None of these can change what desktopAPI.joinCall returns at runtime.

The error itself is handleDesktopJoinedCall receiving undefined from await window.desktopAPI.joinCall(payload). Before digging deeper, can you rule out a stale bundle:

1. In Desktop DevTools → Network, confirm the loaded com.mattermost.calls…bundle.js hash matches the freshly built one in dist/ (hard refresh with Cmd+Shift+R inside the
  Desktop window).
2. Grab the console output around the crash — we log desktopAPI.joinCall right before the await, so we should see whether the Desktop side returned anything.

I can't repro locally on Desktop 6.3.0.

DHaussermann:
@M-ZubairAhmed after taking another look I'm not clear on what's happening here. 

It seems I **can** repro on v1.12.0-rc5 
0/5 Perhaps something was cached in the desktop app before which was hiding the issue. This would explain why getting your updated JS bundle did not solve the problem. But it does not explain why you cannot repro the issue. 

In any case I feel I should open new issue against 1.12.0-rc5. Not sure if that means we want to merge this fix now? I see the bug with the scenario where you need to  end twice but I'm unable to join a call on desktop. 

**Edit**: I created https://mattermost.atlassian.net/browse/MM-68974 to investigate further as it seems unrelated to this PR

M-ZubairAhmed:
Can you try running this locally instructions are here at https://github.com/mattermost/mattermost-plugin-calls/blob/main/README.md#development

DHaussermann:
@M-ZubairAhmed I am already running this locally. Yes, I can check my versions of Go and Node to see if they are current. But I assume it would cause some type of build error if they were too old.

DHaussermann (approved):
Tested and passed
- Confirmed bug is resolved and it only takes one attempt to end call after modal was opened and closed
- Tested other cases of end call and end call for everyone
- Regression tested ending from widget in mobile
- Regression tested from browser and desktop

DHaussermann:
@M-ZubairAhmed from my local, I continue to see an issue on dekstop but it is unrelated to your PR. I built a light bundle I was able to upload to a cloud server and tested there instead so we can move this forward.

Please merge.

PR #1184, opened 2026-05-18
[MM-68308] Migrate standalone entry points to CallClient

M-ZubairAhmed:
#### Summary
This pull request refactors the initialization logic for the standalone embed app, streamlining the `standalone/src/init.ts` (now `index.ts`) entrypoint, and modernizing the way the app connects to calls.

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-68308

#### Release Note

```release-note

```


standalone/src/index.ts
```diff
@@ -105,44 +97,58 @@ function setBasename() {
 }
 
 function connectCall(
-    joinData: CallsClientJoinData,
-    clientConfig: CallsClientConfig,
+    connectPayload: {channelID: string; title?: string; threadID?: string},
```
cpoile: why are we not using CallsClientJoinData anymore?

standalone/src/index.ts
```diff
@@ -188,13 +194,6 @@ export default async function init(cfg: InitConfig) {
         return;
     }
 
-    const joinData = {
-        channelID,
-        title: getCallTitle(),
-        threadID: getRootID(),
-        jobID: getJobID(),
```
cpoile, line 195: now we're missing jobID (maybe we don't need it anymore for recording, is that why? but we're going to be doing recording eventually, if so what will we use?)

standalone/src/index.ts
```diff
@@ -105,44 +97,58 @@ function setBasename() {
...
+        let lastError: Error | undefined;
 
-        window.callsClient.on('close', (err?: Error) => {
+        callClient.on(CALL_EVENT.ERROR, (e: unknown) => {
+            if (e instanceof Error) {
+                lastError = e;
+            }
+        });
+        callClient.on(CALL_EVENT.CONNECTED, () => store.dispatch(setClientConnecting(false)));
+        callClient.on(CALL_EVENT.DISCONNECTED, () => {
```
cpoile: see webapp/src/clients/call/call_client.ts:550-55 -- we should translate non-client initiated reasons into an Error (or forward `reason` through `DISCONNECTED` and let the standalone code handle it), otherwish standalone won't know what the reason is if the server kicks or shuts down

standalone/src/index.ts
```diff
@@ -188,13 +194,6 @@ export default async function init(cfg: InitConfig) {
         return;
```
cpoile: (existing code, I know)  -- if `channelID` is falsy here (or `!channel` at L213), we silently `return` without throwing or calling `cfg.closeCb`. A few effects:

- the widget's `deinitWidget` wouldn't run
- wor recording (if we're going to), the entry uses `runWithRetry(() => initialiseEmbedApp(...))` which only retries on `throw`, so this path will give up silently with no retry

if we threw instead, or closed with an error the deinit path would run

standalone/src/index.ts
```diff
@@ -105,44 +97,58 @@ function setBasename() {
...
+        store.dispatch(setClientConnecting(true));
+        callClient.connect(connectPayload).catch((err: unknown) => {
             store.dispatch(setClientConnecting(false));
             logErr(err);
             if (closeCb) {
-                closeCb(err);
+                if (err instanceof Error) {
+                    closeCb(err);
+                } else {
+                    closeCb();
```
cpoile: Small worry on this `else closeCb()` branch: when `err` isn't an `Error` instance (LiveKit has occasionally rejected with plain objects, and any non-thrown rejection reason gets routed through here too), we lose the cause entirely. That routes to `deinitWidget(err?: Error)`, which then skips `sendDesktopError` — so a real failure ends up looking like a normal leave

Same idea applies to the outer `catch (err)` at L152-157 — a sync-throw from `new CallClient(...)` would land there and `closeCb()` is also called with no arg.

What about `closeCb(err instanceof Error ? err : new Error(String(err)))` in both spots?

cpoile (changes requested):
Hey Zubair, nice work!
Just some thoughts and questions, nothing too serious.  Thanks!

standalone/src/index.ts
M-ZubairAhmed: Added the job ID back

standalone/src/index.ts
M-ZubairAhmed: CallsClientJoinData was renamed to ConnectPayload

standalone/src/index.ts
M-ZubairAhmed: fixed

standalone/src/index.ts
M-ZubairAhmed: fixed

standalone/src/index.ts
M-ZubairAhmed: fixed

standalone/src/index.ts
```diff
@@ -184,17 +191,11 @@ export default async function init(cfg: InitConfig) {
 
     const channelID = getCallID();
     if (!channelID) {
-        logErr('invalid call id');
-        return;
+        const err = new Error('invalid call id');
+        cfg.closeCb?.(err);
+        throw err;
```
cpoile: we should probably pick one of these, either close with error or throw (then the widget would need to catch the throw, I think)

cpoile (approved):
Nice, thanks Zubair!  One more little change, but I'll assume it's easy to fix.

standalone/src/index.ts
M-ZubairAhmed: Done threw the errors and caught then in error handler

PR #1188, opened 2026-05-24
[MM-68793] Synchronize session identity across HA cluster so LiveKit token endpoint is HA-safe

M-ZubairAhmed:
#### Summary
This pull request strengthens the security and reliability of the LiveKit token generation endpoint by ensuring session ownership checks use the database as the source of truth.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-68793
Fixes https://mattermost.atlassian.net/browse/MM-68789

#### Release Note

```release-note

```


server/api.go
```diff
@@ -506,7 +512,7 @@ func (p *Plugin) handleGetLiveKitToken(w http.ResponseWriter, r *http.Request) {
 	}
 	at.SetVideoGrant(grant).
 		SetIdentity(composeLivekitIdentity(userID, sessionID)).
-		SetName(user.GetDisplayName(model.ShowFullName)).
+		SetName(user.Username).
```
M-ZubairAhmed: Although we won’t be using the LiveKit display name—since we’ll instead show the Mattermost name—the username would be a more stable identifier. Display name formats can vary significantly between users.

server/api.go
```diff
@@ -474,11 +474,17 @@ func (p *Plugin) handleGetLiveKitToken(w http.ResponseWriter, r *http.Request) {
...
-	us := p.getSessionByOriginalID(sessionID)
-	if us == nil || us.userID != userID {
+	// Require the session to belong to the requesting user to prevent unauthorized token minting.
+	callSession, err := p.store.GetCallSession(sessionID, db.GetCallSessionOpts{})
+	if err != nil {
+		if !errors.Is(err, db.ErrNotFound) {
+			p.LogError("failed to get call session", "err", err.Error(), "sessionID", sessionID)
+		}
+		res.Err = "Forbidden"
+		res.Code = http.StatusForbidden
```
cpoile, line 488: This should probably be a 500, to distinguish db issues from session issues
(then maybe we should have a test for that, so it doesn't regress)

server/api.go
```diff
@@ -474,11 +474,17 @@ func (p *Plugin) handleGetLiveKitToken(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+	if callSession.UserID != userID {
```
cpoile: should we also verify that the active callID is the one for the requested channelID? That will help prevent users from fiddling with params and tricking us into minting tokens

cpoile (changes requested):
Nice work Zubair!  Just a couple comments, (and the failing test, which you see I'm sure).  Thanks!

server/db/calls_sessions_store.go
```diff
@@ -122,7 +122,7 @@ func (s *Store) GetCallSession(id string, opts GetCallSessionOpts) (*public.Call
 	ctx, cancel := context.WithTimeout(context.Background(), time.Duration(*s.settings.QueryTimeout)*time.Second)
 	defer cancel()
 	if err := s.dbXFromGetOpts(opts).GetContext(ctx, &session, q, args...); err == sql.ErrNoRows {
-		return nil, fmt.Errorf("call session not found")
+		return nil, fmt.Errorf("call session not found: %w", ErrNotFound)
```
lieut-data: nit: In the Calls repo, is it idiomatic to return `nil, err` when something is not found? I typically lean towards `nil, nil` and assume that something not existing is not, necessarily, an error, but I recognize this is a convention we should be consistent on regardless.

server/api.go
```diff
@@ -474,11 +474,17 @@ func (p *Plugin) handleGetLiveKitToken(w http.ResponseWriter, r *http.Request) {
...
-	// The session must belong to the requesting user. Without this check a
-	// client could mint a token claiming any session_id, breaking the
-	// invariant that LiveKit identity == owner's plugin-WS session.
-	us := p.getSessionByOriginalID(sessionID)
-	if us == nil || us.userID != userID {
+	// Require the session to belong to the requesting user to prevent unauthorized token minting.
+	callSession, err := p.store.GetCallSession(sessionID, db.GetCallSessionOpts{})
+	if err != nil {
+		if !errors.Is(err, db.ErrNotFound) {
+			p.LogError("failed to get call session", "err", err.Error(), "sessionID", sessionID)
```
lieut-data: can we use `session_id` to match the snake case used by the server for logging parameters? (Again, not sure if this is a convention in Calls or not, but this should be universal across our code base.)

server/api.go
lieut-data: Should we use `user.Id` for maximum stability, e.g. in the face of potentially a user name changing mid-call?

server/api.go
M-ZubairAhmed: Done

server/api.go
M-ZubairAhmed: Good call, added

server/api.go
M-ZubairAhmed: Switched to user.Id

server/api.go
M-ZubairAhmed: done added snake case for logging

server/db/calls_sessions_store.go
M-ZubairAhmed: this repo uses `nil,err` for not found consistently

server/api.go
bgardner8008: The participant.name field, set by SetName when minting the token, is currently unused. We look up the display name from the redux store based on the unique identity, so we are currently able to change name dynamically. If we were ever to use LiveKit egress, it would pull display name from participant properties derived from the token. So I slightly prefer user.Username. But, it's moot as we could easily change this if we ever needed to use egress.

PR #1189, opened 2026-05-25
Include voip in notification payload to make the phone ring

enahum:
#### Summary
Sets the push notification for new calls to use a voip transport so that the devices that support it ring instead of just receiving a normal notification.

#### Ticket Link
TDB

#### Release Note
```release-note
Added capability to notify the phone that there is an incoming call and ring even while the app is closed.
```


enahum:
/test-analysis-override the Call push with voip is verified at the layer where it actually matters, in mattermost-server. TestSendPushNotificationsTransportRouting in channels/app/notification_push_test.go covers the full matrix of session/transport combinations end-to-end and asserts the resulting outbound message. TestSendPushNotificationsVoIPRemoveTracking covers the VoIP-token invalidation path. In the plugin, the only thing that would need its own assertion is the literal Transport: model.PushTransportVoIP field set on the struct in sendPushNotifications, and to capture that the maintenance to set it up is extremely high and worthless

enahum:
@lieut-data can you do your magic so that the model picks up the new property ?

lieut-data:
Absolutely, @enahum -- I've lifted the needed model changes from https://github.com/mattermost/mattermost/pull/36726 into https://github.com/mattermost/mattermost/pull/36829 so I can tag and release this for the plugin to consume. Will update here once merged.

lieut-data:
@enahum, tagged v0.4.2 with the `PushNotification` model changes and updated here to reflect.

lieut-data (approved):
Lovely and simple :)

PR #1193, opened 2026-05-27
[MM-68570] Wire E2E framework for LiveKit feature migration (PR 1/5)

bgardner8008:
## Summary

First PR in the MM-68570 series — bring the E2E suite back online against the LiveKit backend. This one is pure plumbing; no test currently runs that didn't run before.

- **Fix the shared helpers.** `e2e/page.ts` waits on `window.callsClient.connected` / `.closed` (lines 58, 66, 87). Neither field exists on the LiveKit `CallClient` (they live as private `isRoomConnected` / `isDisconnected`), so every test using `devPage.startCall()`/`joinCall()`/`leaveCall()` hung at `waitForFunction`. This PR adds public read-only getters that delegate to the existing private state. The smoke test in `livekit_smoke.spec.ts` worked only because it bypassed those helpers — now the helpers work too.
- **Add a WS-reconnect test hook.** `CallClient._e2eForceWebsocketClose()` (which calls a new `WebSocketClient.e2eForceClose()`) lets Playwright simulate a transport-level WS drop. Without this hook, the RTCD-era `unmuting after ws reconnect` test cannot be migrated.
- **Tag + quarantine the 14 RTCD-era spec files** so CI can grep the broader suite without breaking. Every test gains `{tag: '@livekit'}` and either `test.fixme` (Bucket A/B — to be revived in PRs 2–5) or `test.skip` (Bucket C — popout, recording, notifications, desktop — deferred until their underlying features ship or until separate scoping). The CI `--grep` in `e2e/scripts/run.sh` widens from `@livekit-smoke` to `@livekit-smoke|@livekit`.
- **Document the new conventions** in `e2e/README.md`.

Targets `v2` because the LiveKit migration lives there; `main` is still RTCD.

The next PRs in the series — Bucket A revive, start/join/audio, screen share + host controls, and net-new two-user tests — are scoped in `/Users/billg/.claude/plans/iterative-napping-pike.md` (Ultraplan-refined).

## Test plan

- [ ] `@livekit-smoke` test still passes (no regression — the smoke test bypasses the helpers we touched).
- [ ] `--grep '@livekit-smoke|@livekit'` finds the smoke test plus every quarantined test, reports the smoke test as passing and the rest as `fixme`/`skipped` — zero failures.
- [ ] No bare `test(` lines remain in any spec file except `livekit_smoke.spec.ts`.
- [ ] Each Bucket C spec file has a top-of-file comment naming the gating reason.


webapp/src/clients/call/call_client.ts
```diff
@@ -92,6 +92,23 @@ export default class CallClient extends EventEmitter {
         room.on(RoomEvent.ConnectionQualityChanged, this.handleConnectionQualityChanged.bind(this));
     }
 
+    // E2E helpers — public read-only views of internal state so Playwright
+    // tests can wait on `window.callsClient.connected` / `.closed`.
+    public get connected(): boolean {
```
M-ZubairAhmed:
```suggestion
    public get isConnected(): boolean {
```

webapp/src/clients/call/call_client.ts
```diff
@@ -92,6 +92,23 @@ export default class CallClient extends EventEmitter {
...
+        return this.isRoomConnected;
+    }
+
+    public get closed(): boolean {
```
M-ZubairAhmed:
```suggestion
    public get isDisconnected(): boolean {
```

M-ZubairAhmed (changes requested):

webapp/src/clients/call/call_client.ts
bgardner8008: Done in 9caf50d7. Renamed both private fields with an underscore prefix (`_isRoomConnected` / `_isDisconnected`) to free up the public names — without that, the public `isDisconnected` getter would collide with the existing private field of the same name. Updated the e2e callers (page.ts + the test.skip'd recordings.spec.ts) and cascaded the rename through PRs 2-4.

webapp/src/clients/call/call_client.ts
bgardner8008: Done in 9caf50d7 (along with the `connected` → `isConnected` rename above).

webapp/src/clients/call/call_client.ts
bgardner8008: linter did not like underscores, ended up using predicate isXXX() style for functions and roomConnected, disconnected variables.

M-ZubairAhmed (approved):

PR #1204, opened 2026-06-02
[MM-68565][MM-68566] Raise/lower hand, emoji reactions

M-ZubairAhmed:
#### Ticket Link
https://mattermost.atlassian.net/browse/MM-68565
https://mattermost.atlassian.net/browse/MM-68566

#### Release Note

```release-note

```


server/api.go
```diff
@@ -520,8 +520,9 @@ func (p *Plugin) handleGetLiveKitToken(w http.ResponseWriter, r *http.Request) {
 
 	at := auth.NewAccessToken(cfg.LiveKitAPIKey, cfg.LiveKitAPISecret)
 	grant := &auth.VideoGrant{
-		RoomJoin: true,
-		Room:     requestingChannelID,
+		RoomJoin:             true,
+		Room:                 requestingChannelID,
+		CanUpdateOwnMetadata: new(true),
```
bgardner8008: I don't think this will compile. I prefer setting this using an API call rather than allocating pointer to a boolean (bizarre golang convention to allow nil/true/false states).

```
  grant := &auth.VideoGrant{
      RoomJoin: true,
      Room:     requestingChannelID,
  }
  grant.SetCanUpdateOwnMetadata(true)
```

webapp/src/components/expanded_view/component.tsx
```diff
@@ -584,6 +589,15 @@ export default class ExpandedView extends React.PureComponent<Props, State> {
...
+
+            // The WS call_state seed carries the server's stale unmuted/raised_hand
+            // (those fields moved to LiveKit). Once the session list is populated,
+            // overlay the live mute + raised-hand state from the opener's CallClient.
+            // Runs once; incremental CALL_EVENTs keep it fresh afterward.
+            if (!this.hasReSyncedLiveKitState && this.props.sessions.length > 0) {
+                this.hasReSyncedLiveKitState = true;
+                getCallsClient()?.reSyncMuteAndHandState();
+            }
         }
```
bgardner8008, line 601: At some point we should remove hand raise and mute state tracking from server/DB and from call_state messages, but that can wait. For now it remains as vestigial Calls v1 artifacts.

bgardner8008 (changes requested):
Thanks Zubair, this looks great, except for the new(true) code. See comments.

bgardner8008 (approved):
Looks great, thanks.

PR #1207, opened 2026-06-04
[MM-69034] Single-path call teardown via Disconnected event

M-ZubairAhmed:
#### Summary
* Upgraded `livekit-client` from version `2.8.0` to `2.19.2`
**Call Teardown and Disconnect Handling**
* Refactored call teardown logic: removed direct calls to `window.callsClient?.disconnect()` in the widget and recording deinitialization, delegating disconnect handling to the call state logic.
* Improved Redux actions by introducing `getCallIDForChannel` and `getCallIDForCurrentCall` selectors, and updated the logic for ending and leaving calls to ensure notifications and teardown are handled consistently.
* See the tickets for bug fixes

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-68309
https://mattermost.atlassian.net/browse/MM-69034
https://mattermost.atlassian.net/browse/MM-69111
https://mattermost.atlassian.net/browse/MM-69162

#### Release Note

```release-note

```


bgardner8008 (changes requested):
We do need to be careful on this, because the current code works fine, and although there are redundant teardown paths that create some log noise, it does work. By isolating call teardown to the LiveKit disconnected event, we risk creating partial teardowns in some situations. I created this ticket, so it's on me!

I'm now wondering if we should wait on this until we also mate it with E2E tests that explicitly test the edge cases, like user hitting hangup before the room is created.

The key issue is that all teardown is gated on RoomEvent.Disconnected, which LiveKit does not emit if the room never connected. So if things so bad during call setup the UI can be created but we never connected so we don't get a disconnected event. Claude has a reasonable looking analysis and fix which I'll have it post in a subsequent comment.

bgardner8008:
### Teardown is fully gated on `RoomEvent.Disconnected`, which LiveKit doesn't emit if the room never connected

**Root cause**

All teardown now lives in `handleDisconnected()`, which only runs on `RoomEvent.Disconnected`. In `livekit-client@2.8.0`, `Room.disconnect()` early-returns **without emitting** that event when the room is still in its initial `Disconnected` state:

```js
// livekit-client.esm.mjs:21486
if (_this2.state === ConnectionState.Disconnected) {
    _this2.log.debug('already disconnected', ...);
    return;   // no RoomEvent.Disconnected
}
```

The room sits in `Disconnected` for the entire pre-`room.connect()` window — the WS handshake (`websocketClient.ready()`), the token fetch, and `prepareConnection()`. So `disconnect()` called in that window is a silent no-op:

- the plugin WS leaks and `sendLeave` is never sent (server still thinks the user is in the call);
- `CALL_EVENT.DISCONNECTED` never fires, so the widget never unmounts and `window.callsClient` is never cleared — the user is stuck in a dead call.

Reachable triggers:
- a hang-up/cancel while the call is still connecting (`call_widget/component.tsx onDisconnectClick` → `disconnect()`; the widget is mounted during connect);
- the `WebSocketErrorType.Join` path in `handleWebsocketErrored`, which emits `ERROR` then calls `disconnect()` without closing `websocketClient` — and a join failure is exactly the case where the room hasn't connected yet.

This is a regression: the old `disconnect()` closed the WS + sent leave unconditionally, regardless of room state.

---

### Fix A — primary (closes all triggers)

`server/.../webapp/src/clients/call/call_client.ts` — `disconnect()`

Choose the teardown trigger based on room state: if the room can fire the event, let it; otherwise drive `handleDisconnected()` directly (it's already idempotent via the `this.disconnected` guard and null-checks `websocketClient`).

```js
    // The single public entry point to end a call. Teardown is normally driven by the
    // resulting RoomEvent.Disconnected -> handleDisconnected(). But LiveKit's
    // room.disconnect() only fires that event once the room has moved past its initial
    // Disconnected state; called earlier (a hangup/cancel or a pluginWS Join error during
    // the connect handshake) it is a silent no-op. In that window we drive teardown
    // directly so the pluginWS is closed, sendLeave is sent, and the CALL_EVENT.DISCONNECTED
    // listeners (widget unmount, window.callsClient cleanup) still run.
    public disconnect(): void {
        if (this.room && this.room.state !== ConnectionState.Disconnected) {
            void this.room.disconnect();
            return;
        }
        this.handleDisconnected();
    }
```

| Trigger | Room state at `disconnect()` | Before | After |
|---|---|---|---|
| Hang-up on a connected call | `Connected` | event → teardown ✓ | unchanged ✓ |
| Hang-up/cancel mid-connect | `Disconnected` (pre-`room.connect()`) | **no-op — leak + stuck UI** | `handleDisconnected()` runs directly ✓ |
| `WebSocketErrorType.Join` | usually `Disconnected` | **no-op — WS leaks** | direct teardown closes WS ✓ |
| `WebSocketErrorType.ReconnectTimeout` | varies | no `DISCONNECTED` if room not up | direct teardown emits `DISCONNECTED` ✓ |
| Cancel while `room.connect()` in flight | `Connecting` | LiveKit aborts → event ✓ | unchanged ✓ |

This also makes `disconnect()` return `void` on every path, so the `consistent-return` rule disabled in this PR could be restored.

---

### Fix B — optional hardening (cosmetic)

With Fix A, a user cancel during the *pre-connect handshake* closes the plugin WS, which makes the in-flight `ready()`/token fetch reject — and `connect()`'s `catch` blocks then `emit(CALL_EVENT.ERROR)`, popping a spurious "connection error" modal on top of an intentional cancel. (This does **not** affect the `Join`-error path, where the error modal is correct.) To suppress it, guard the `catch` blocks in `connect()`:

```js
        } catch (err) {
            logErr('CallClient: pluginWS connection error', err);
            this.connectPayload = null;

            // Already torn down (e.g. user cancelled mid-connect): the WS is closed
            // elsewhere and this rejection is expected — don't surface it as an error.
            if (this.disconnected) {
                throw err;
            }

            this.websocketClient?.close();
            this.websocketClient = null;
            this.emit(CALL_EVENT.ERROR, err);
            throw err;
        }
```

Fix A is the required change; Fix B is polish for a narrow edge.

M-ZubairAhmed:
@coderabbitai review

M-ZubairAhmed:
@coderabbitai full review

M-ZubairAhmed:
@coderabbitai full review

bgardner8008 (approved):
The changes look good, addressing the teardowns during partial setup. Claude suggested adding unit tests which might be a good idea but we're up against the clock, and there are the coderabbit issues which I assume you are looking at. I'm going to approve.

M-ZubairAhmed:
@coderabbitai review

PR #1208, opened 2026-06-04
[MM-68569] Fix intermittent Desktop join 429 + LAN TLS proxy for testing

bgardner8008:
## Summary

Part of getting Calls v2 (LiveKit) working on the Desktop app (MM-68569).

### Fix intermittent "Something went wrong with calls" on call start/join (Desktop)

The plugin's per-user API rate limiter (`rate.NewLimiter(1, 10)`) wraps the main router, which also serves the `/standalone/` static bundle (widget html, JS, fonts). On Desktop the global widget is a separate window that loads that bundle over HTTP, so on a cold cache the static requests exhausted the burst-10 bucket before the `/livekit-token` fetch — 429'ing the join. Browsers use the in-page view and never request those assets, so the failure was intermittent (cache- and deploy-dependent).

- Exempt the `/standalone/` static route from the limiter, matching the exemption it already has in the auth middleware. Static-asset serving isn't an API and is cheap/cacheable; DOS protection belongs in the core per-IP rate limiter + proxy, not this per-user API throttle.
- Log rate-limit rejections, which were previously silent (the handler's audit log never runs on a 429), so this class of failure is visible server-side.

After the fix a Desktop join is ~4 API calls against burst 10; verified across 6 start/join cycles including a cold cache.

### Add a LiveKit TLS proxy for LAN testing

`make livekit-tls-docker-start` brings up a Caddy sidecar that terminates TLS for the LiveKit signaling WebSocket, so an external LAN machine (e.g. the Windows Desktop app) can connect to a local HTTPS Mattermost over `wss://`. The plaintext `livekit-docker-start` flow is unchanged. Dev tooling, kept here for convenience — could be split to its own PR.

## Related
- Screen share from Desktop: MM-68561
- Channel toast staleness: MM-69111

## Testing
- 6x Desktop call start/join including cold cache — no 429s.
- Screen share verified separately on the MM-68561 build.


M-ZubairAhmed (approved):
looks good

PR #1209, opened 2026-06-05
[MM-69107] Drive host lower-hand and mute through the LiveKit admin API

bgardner8008:
## Summary

Fixes the host **lower-hand** and **mute / mute-others** controls under LiveKit (v2), which were wired end-to-end but silently did nothing.

**Root cause:** all three gated on the server-side `session.RaisedHand` / `session.Unmuted` fields. The v2 `CallClient` moved hand state onto a LiveKit participant attribute and mute onto the track state, and it no longer sends the raise-hand / mute WS messages. So those server fields stay frozen at their join defaults (`RaisedHand=0`, `Unmuted=false`) and the guards always short-circuited:

- `hostLowerParticipantHand` — `if ust.RaisedHand == 0 { return nil }` always true
- `hostMuteParticipant` — `if !ust.Unmuted { return nil }` always true (so individual mute was also fully broken, not just "unverified")
- `hostMuteAllParticipants` — `if s.Unmuted && …` never true, so the whole loop was skipped

**Fix:** drop the vestigial guards and act unconditionally through the LiveKit admin API, which is idempotent:

- Lower-hand: new `livekitLowerParticipantHand` clears the `raised_hand` attribute via `UpdateParticipant` (empty string deletes it), mirroring the client's own `unraiseHand()`.
- Mute / mute-others: force-mute the mic track via the existing `livekitMuteParticipant` (no-ops if there is no unmuted mic track).

LiveKit then propagates the change to **every** participant — including observers — via `ParticipantAttributesChanged` / `TrackMuted`, which is the reliable observer-sync path. The targeted host-control WS events (`host_lower_hand`, `host_mute`) are retained as client-side UX notices and as a fallback when LiveKit is not configured.

## Relationship to other tickets

- **Depends on** #1204 (MM-68565/MM-68566) for the client-side LiveKit hand attribute — already on `v2`.
- The broader removal of the vestigial `Unmuted`/`RaisedHand`/`Video` server plumbing (columns, `UpdateCallSession`, dead WS handlers) is tracked separately in **MM-69116**; this PR only stops *relying* on those fields.
- Full host-control UI is being wired in **MM-68567**; some controls aren't reachable from the UI until that lands.

## Other host controls (unchanged)

`remove`, `stop screen share`, `make host`, and `end call` were not affected by the vestigial-state bug and are left as-is. Stop-screen-share relies on the Calls WebSocket round-trip to the target (no LiveKit fallback) by design.

## Testing

Build / vet / gofmt clean. Manual two-user end-to-end verification pending (the acceptance signal for this ticket).

### DIFF
--- a/server/host_controls.go
+++ b/server/host_controls.go
@@ -94,10 +94,11 @@ func (p *Plugin) hostMuteParticipant(requesterID, channelID, sessionID string) e
 		return ErrNotInCall
 	}
 
-	if !ust.Unmuted {
-		return nil
-	}
-
+	// NB: we deliberately do not gate on server-side mute state. Under LiveKit
+	// (v2) mute lives in the track state; the server's session.Unmuted is
+	// vestigial (never updated post-migration, see MM-69116) and would always
+	// read false here. livekitMuteParticipant is idempotent — it no-ops if the
+	// participant has no unmuted mic track.
 	if err := p.livekitMuteParticipant(channelID, composeLivekitIdentity(ust.UserID, sessionID)); err != nil && !errors.Is(err, errLiveKitNotConfigured) {
 		p.LogError("hostMuteParticipant: failed to mute participant via LiveKit",
 			"channelID", channelID, "sessionID", sessionID, "err", err.Error())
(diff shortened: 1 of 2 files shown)


M-ZubairAhmed (approved):

M-ZubairAhmed:
@coderabbitai review

M-ZubairAhmed:
@bgardner8008 do you want to add some tests?

bgardner8008:
Regarding tests, great idea, but unit tests are kindof pointless here, I'd prefer E2E tests and those should go in separate ticket. Because we don't want to add more to milestone 1, I've created a new ticket to complete the milestone 1 E2E tests, as part of milestone 2. :-) https://mattermost.atlassian.net/browse/MM-69158, and those tests will include host controls.

PR #1211, opened 2026-06-05
[MM-69148] Wire LiveKit-owned mute/speaking/raised-hand state into standalone bundles

bgardner8008:
## Summary

After the LiveKit migration, per-participant call state (mute, speaking/voice activity, raised hand, reactions) is owned by LiveKit and re-emitted by `CallClient` as `CALL_EVENT.*` rather than traveling over the plugin WebSocket. The standalone bundles (`standalone/src/index.ts`, shared by the **widget** and the **recording** page) never got the bridge that feeds this state into the UI, so the indicators were wrong:

- **Recording:** every participant stuck "muted", speaking outline never drew, hands/reactions never showed.
- **Widget (Desktop):** remote participants' mute/speaking indicators stale (local user was fine).

The main webapp (`webapp/src/index.tsx`) and popout (`expanded_view/component.tsx`) were already wired correctly; only the standalone surfaces were missing.

## Changes

- **`standalone/src/index.ts` — central bridge in `connectCall()`:** subscribe to `CALL_EVENT.MUTE`/`UNMUTE` → `userMuted`/`userUnmuted`, `USERS_VOICE_ACTIVITY_CHANGED` → `usersVoiceActivityChanged`, `RAISE_HAND`/`LOWER_HAND` → `userRaisedHand`/`userLoweredHand`, and `REACTION` → `dispatchReaction`, all using the event's real session/userID. Both widget and recording inherit this.
- **Replay-once seed overlay:** call `reSyncMuteAndHandState()` exactly when *both* the LiveKit room is connected *and* the WS `call_state` seed has populated the session list. The session reducers drop events for sessions they don't yet know about, so the at-connect emits race ahead of the seed and are lost — this covers the recording bot joining mid-call. Handles either arrival order.
- **`standalone/src/widget/index.tsx`:** removed the `on('mute')`/`on('unmute')` handlers that ignored the event args and hardcoded the local user (superseded by the shared bridge). Video handlers left untouched (out of scope).
- Removed the now-dead plugin-WS switch cases (`user_voice_on/off`, `user_muted/unmuted`, `user_raise_hand/unraise_hand`, `user_reacted`) and their unused imports — these events no longer travel over the plugin WS in v2.

## Testing

- Verified live: in a recording, mute + speaking (green outline) + raised hands update correctly; in the Desktop widget, remote participants' mute/speaking are correct.
- `npm run check-types` and `npm run lint` pass in `standalone/`.
- Main webapp and popout behavior unchanged.

## Notes

- Discovered while verifying MM-69144 (enable recording for Calls v2). The two changes are independent at the file level; this branch is based directly on `v2`.
- Reactions ride the LiveKit data-message path from MM-68566.
- The dead-WS-case removal overlaps the vestigial-v2-state cleanup tracked in MM-69116.

https://mattermost.atlassian.net/browse/MM-69148

### DIFF
--- a/standalone/src/index.ts
+++ b/standalone/src/index.ts
@@ -14,6 +14,7 @@ import {
     CallJobStateData,
     CallStartData,
     CallStateData,
+    EmojiData,
     EmptyData,
     HelloData,
     HostControlLowerHand,
@@ -22,13 +23,9 @@ import {
     UserDismissedNotification,
     UserJoinedData,
     UserLeftData,
-    UserMutedUnmutedData,
-    UserRaiseUnraiseHandData,
-    UserReactionData,
     UserRemovedData,
     UserScreenOnOffData,
     UserVideoOnOffData,
-    UserVoiceOnOffData,
     WebsocketEventData,
 } from '@mattermost/calls-common/lib/types';
 import {WebSocketMessage} from '@mattermost/client/websocket';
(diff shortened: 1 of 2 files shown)


M-ZubairAhmed (approved):

M-ZubairAhmed:
@coderabbitai full review

PR #1212, opened 2026-06-08
[MM-621] Ring on incoming calls in CM

### LINKED ISSUE #621: Feature request: Add ringtone alarm to the group channels
Hello!

We are aware of the considerations about the subject (for example here https://github.com/mattermost/mattermost-plugin-calls/issues/125) but still in our use cases we do need the ringtone to the channel group calls also (like in DM channels).

It would be wonderful if this could be a option for the admin to set and even more awesome if then the option could be still configurable per channel bases.

 
Admin-setting
Channel-setting

Judx:
## Summary
- Incoming call rings for channels

## Behavior
- Channel ringing uses existing push notification delivery.
- Only users with push notifications enabled for that channel will ring.
- Fixes #621 

## Test plan
- [X] Call in DM -> ringing still works
- [X] Call in GM - ringing still works
- [X] Call in CM -> rings

### DIFF
--- a/server/push_notifications.go
+++ b/server/push_notifications.go
@@ -47,6 +47,9 @@ func (p *Plugin) NotificationWillBePushed(notification *model.PushNotification, 
 		return nil, ""
 	}
 
+	// Mark as a calls notification so mobile clients ring for channel calls.
+	notification.SubType = model.PushSubTypeCalls
+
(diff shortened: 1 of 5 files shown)


bgardner8008:
We appreciate the submission, but... we don't want channel calls to ring, we only want DM/GM calls to ring. So the omission of this feature is by design.

PR #1215, opened 2026-06-09
[MM-69176] Don't fabricate errors for clean host-driven disconnects on Desktop

bgardner8008:
## Summary

Fixes the spurious **"Something went wrong with calls"** error modal that Desktop users saw when a host ended a call for everyone (or removed an individual user). The teardown itself was already correct — this is a client-side fix in the standalone widget bundle that Desktop runs.

## Root cause

When the host ends a call, the server deletes the LiveKit room and every client's LiveKit SDK fires `RoomEvent.Disconnected` with reason `ROOM_DELETED` (the single-path teardown from MM-69034). The standalone widget fabricated an `Error` from *any* non-`CLIENT_INITIATED` reason, which flowed through `closeCb` → `deinitWidget(err)` → `sendDesktopError('client-error', ...)` → the generic error modal in the main webapp window.

This was Desktop-only because `sendDesktopError` only fires when `window.desktopAPI` is present; in a plain browser the call runs through the webapp's `CallWidget`, which never hits this path. `PARTICIPANT_REMOVED` (host removes one user) had the same flaw.

## Fix

Treat `ROOM_DELETED` and `PARTICIPANT_REMOVED` as clean disconnects alongside `CLIENT_INITIATED`, so no error is fabricated for expected server-driven disconnects. Genuine media-plane failures (`SERVER_SHUTDOWN`, `JOIN_FAILURE`, transport errors that set `lastError`) continue to report as before.

## Test cases

- [ ] Host ends call for everyone on Desktop → user is removed with no error modal.
- [ ] Host removes an individual user on Desktop → no error modal.
- [ ] Genuine abnormal disconnects (server shutdown, join failure, transport errors) still surface an error.
- [ ] No `disconnected from room (reason: ...)` log line on a clean host-end or host-kick.

## Ticket

https://mattermost.atlassian.net/browse/MM-69176

Related: MM-69034 (single-path teardown via `RoomEvent.Disconnected`).


M-ZubairAhmed (approved):

standalone/src/index.ts
```diff
@@ -185,7 +185,18 @@ function connectCall(
...
             if (closeCb) {
                 let err = lastError;
-                if (!err && typeof reason === 'number' && reason !== DisconnectReason.CLIENT_INITIATED) {
+
+                // Server-driven disconnects that aren't failures: the user left, the
+                // host ended the call for everyone, or the host removed this user.
+                // These must not be fabricated into an error (otherwise Desktop shows
+                // a spurious "Something went wrong with calls" modal). Genuine media-
+                // plane failures still set lastError and report as before.
+                const cleanReasons = [
```
M-ZubairAhmed: lets also terse this a bit to two lines at max

bgardner8008: Agreed, done.

PR #1125, opened 2026-02-04
Add video statistics and metrics

bgardner8008:
#### Summary
This PR adds comprehensive video usage tracking and metrics to the Calls plugin. It tracks participant-seconds of video usage, video-enabled call counts, and screen sharing statistics in real-time during calls, storing them in the call stats JSON field. New database queries calculate aggregate metrics (average/total video duration, counts of calls with video/screen share) with cross-database support for MySQL and PostgreSQL. A background job updates Prometheus gauges hourly for monitoring and alerting, with metrics exposed via the existing /metrics endpoint. Performance is optimized by deferring database writes until users leave or calls end. Includes test coverage.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-66865

PR #1219, opened 2026-06-12
[MM-69233] Persist video stats on toggle to fix Calls video metrics

bgardner8008:
### Summary

The Grafana "Calls metrics" dashboard reported **Video Calls = 0** even when video was actively used, and **Avg/Total Video Duration** were likewise zero. Screen-share metrics were unaffected.

Root cause: the video on/off handler set `Stats.HasUsedVideo` and `Props.VideoStartAt` in memory but only persisted the session row via `UpdateCallSession` — it never called `UpdateCall`. Because every websocket handler reloads call state fresh from the DB (`lockCallReturnState` → `getCallState` → `GetActiveCallByChannelID`), those in-memory mutations were discarded before the next handler, the leave path, or call-end finalization could read them. So `has_used_video` was written back as `false`, and the duration-finalization loops found an empty `VideoStartAt` map and accumulated nothing.

The handler carried a comment claiming video stats were "persisted when users leave or call ends" — that assumption was wrong, since `state.Call` is reloaded from the DB on every invocation.

### Fix

Persist the call via `p.store.UpdateCall(&state.Call)` on every video toggle, in addition to the existing `UpdateCallSession`. This writes both `Stats.HasUsedVideo` and `Props.VideoStartAt`, fixing the call count and both duration metrics. This is the same pattern the screen-share handler already uses (restored in #1178 after the same regression was introduced for screen share by #1125). Video toggles are infrequent relative to audio mute, so the per-toggle write cost the original comment worried about is negligible.

Introduced by #1125 ("Add video statistics and metrics").

### Testing

Added `TestHandleClientMsgVideoStats`, which drives the real `handleClientMsg` handler and re-reads state from the store to exercise the reload-from-DB boundary:

- `video on` → `HasUsedVideo`, `VideoStartAt[connID]`, and the session video flag are all persisted.
- `video off` → `VideoDuration` is accumulated and persisted, the `VideoStartAt` entry is cleared, and `HasUsedVideo` remains set.

Confirmed the test fails without the fix and passes with it. Full `./server/` package suite passes.

### Acceptance criteria

- [x] After a call that used video, `stats->>'has_used_video'` is `true` in the `calls` row.
- [x] Dashboard "Video Calls" reflects the real count; "Avg/Total Video Duration" are non-zero when video was used.
- [x] Server test covering persistence of `HasUsedVideo` and `VideoDuration` across the reload-from-DB boundary.

https://mattermost.atlassian.net/browse/MM-69233

### DIFF
--- a/server/websocket.go
+++ b/server/websocket.go
@@ -450,8 +450,14 @@ func (p *Plugin) handleClientMsg(us *session, msg clientMessage, handlerID strin
 			return fmt.Errorf("failed to update call session: %w", err)
 		}
 
-		// Note: Video stats are persisted when users leave or call ends,
-		// no need to write to DB on every toggle for performance.
+		// Stats.HasUsedVideo and Props.VideoStartAt must be persisted on every
+		// toggle: getCallState reloads from the DB on each handler invocation, so
+		// without this write the next video_on/video_off, the leave path, and the
+		// call-end finalization would all see stale state — leaving has_used_video
+		// false and the video durations at zero (MM-69233).
+		if err := p.store.UpdateCall(&state.Call); err != nil {
+			return fmt.Errorf("failed to update call: %w", err)
+		}
 
 		evType := wsEventUserVideoOn
 		if msg.Type == clientMessageTypeVideoOff {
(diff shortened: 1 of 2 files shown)


cpoile (approved):
Nice!

PR #1221, opened 2026-06-16
fix issue not show icon start call in chaneel

davoodf1995:
…icense checker

- Updated session tests to allow calls in DMs only for unlicensed cloud environments and added a new test for self-hosted deployments.
- Modified the LicenseChecker to support group calls in self-hosted deployments without requiring a Professional license, based on environment variables.

#### Release Note

```release-note

```

### DIFF
--- a/server/enterprise/license.go
+++ b/server/enterprise/license.go
@@ -58,5 +58,15 @@ func (e *LicenseChecker) HostControlsAllowed() bool {
 }
 
 func (e *LicenseChecker) GroupCallsAllowed() bool {
-	return e.isAtLeastProfessionalLicensed() || os.Getenv("MM_CALLS_GROUP_CALLS_ALLOWED") == "true"
+	if os.Getenv("MM_CALLS_GROUP_CALLS_ALLOWED") == "true" {
+		return true
+	}
+	if os.Getenv("MM_CALLS_GROUP_CALLS_ALLOWED") == "false" {
+		return false
+	}
+	// Self-hosted deployments support calls in public/private channels without Professional.
+	if !license.IsCloud(e.api.GetLicense()) {
+		return true
+	}
+	return e.isAtLeastProfessionalLicensed()
 }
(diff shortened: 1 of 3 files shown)


bgardner8008:
Thanks for the contribution! A few concerns from review, most significant first:

**1. `server/enterprise/license.go` — `MM_CALLS_GROUP_CALLS_ALLOWED=false` now hard-disables group calls even for licensed servers (regression).**
The old logic was `isAtLeastProfessionalLicensed() || env=="true"`, so a Professional/Enterprise/cloud-paid server kept group calls regardless of the env var. The new code returns `false` immediately when the var is `"false"`, *before* the license check. Any operator who set `MM_CALLS_GROUP_CALLS_ALLOWED=false` (previously a no-op for licensed servers) would now lose group calls entirely — this turns a soft override into one that beats a paid license.

**2. `server/enterprise/license.go` — the `!IsCloud(...)` branch removes the Professional-license requirement for group calls on *all* self-hosted deployments.**
This is a licensing-policy change rather than just an icon fix: every unlicensed self-hosted server would now get group calls for free, where previously Professional (or the env var) was required. This is much broader than "show the missing button" and would need product/licensing sign-off before it could merge.

**3. `webapp/src/components/channel_header_dropdown_button/index.ts` — uses `isDmGmChannel` where the server only exempts DMs.**
The server gate in `server/session.go` (`channelType != model.ChannelTypeDirect && !GroupCallsAllowed()` → `errGroupCallsNotAllowed`) exempts only `ChannelTypeDirect`; a GM is not exempt. The sibling `channel_header_button.tsx` correctly uses `isDMChannel(channel)`. With `isDmGmChannel` (DM **or** GM), when group calls aren't allowed the dropdown button appears in a GM, the user clicks "Start call," and the server rejects it. This should use `isDMChannel` to match the server gate and the sibling component.

M-ZubairAhmed:
This removes the Professional-license permission check for group calls on self-hosted deployments effectively bypassing the licensing gate rather than just fixing the missing icon. Closing this off.

PR #1222, opened 2026-06-16
[MM-69319] Fix screen share alert wording (recording -> sharing)

bgardner8008:
## Summary

When a user cancels or is denied a screen **share**, Calls shows the `missingScreenPermissions` alert, but its copy referred to screen **recording**, which is misleading (reported by Dylan).

This reword fixes the two affected strings in `CallAlertConfigs.missingScreenPermissions`:

- Banner: "Screen **recording** access is not currently allowed or was canceled." → "Screen **sharing** access…"
- Tooltip subtext: "Allow screen **recording** access to Mattermost." → "Allow screen **sharing** access…"

The tooltip text "No screen sharing permissions" was already correct and is unchanged. `i18n/en.json` regenerated accordingly.

## Note

This alert also fires on a plain picker cancel (not just permission denial). Distinguishing cancel from denial is unreliable cross-browser and is out of scope here — this is a wording fix only.

## Ticket

https://mattermost.atlassian.net/browse/MM-69319


DHaussermann (approved):
Tested and passed
- Confirmed I now see an explicit message about screen sharing failure
- Confirmed screen recording messages are still functional 

LGTM!

PR #1223, opened 2026-06-16
[MM-69320] Default "Share sound with screen" to on

bgardner8008:
## Summary

The per-user **Share sound with screen** setting (Settings → Calls → Screen sharing settings) defaulted to **off**. While off, Calls requests `getDisplayMedia({audio: false})`, so the browser never shows any share-audio option — users had no way to discover audio sharing and ended up telling each other to flip a switch that's hidden by default.

This flips the default to **on** (`STORAGE_CALLS_SHARE_AUDIO_WITH_SCREEN`):

- `shareAudioWithScreen()` now returns `true` unless localStorage is explicitly `'off'`.
- The settings radio shows **On** by default when the user hasn't chosen.

A user who previously selected **Off** still has `'off'` stored and keeps that behavior.

## Why this is safe

- Chrome/Edge couple the audio option to the capture request: the checkbox only appears when we request audio (pre-checked for tab shares, opt-in for screen/window). No audio is captured without the picker showing the control.
- Firefox/Safari ignore audio entirely.

## Release notes

Tab-audio becomes on-by-default in Chromium when sharing your screen.

https://mattermost.atlassian.net/browse/MM-69320

### DIFF
--- a/webapp/src/components/user_settings/screen_sharing_settings_section.tsx
+++ b/webapp/src/components/user_settings/screen_sharing_settings_section.tsx
@@ -14,7 +14,7 @@ export default function ScreenSharingSettingsSection() {
     const description = formatMessage({defaultMessage: 'Configure your screen sharing settings.'});
     const editLabel = formatMessage({defaultMessage: 'Edit'});
 
-    const [shareAudio, setShareAudio] = useState(window.localStorage.getItem(STORAGE_CALLS_SHARE_AUDIO_WITH_SCREEN) || 'off');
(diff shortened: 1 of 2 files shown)


bgardner8008:
/test-analysis-override We are changing the audio share default from false to true; we don't need unit tests for that

DHaussermann (approved):
Tested and passed
- Confirmed the setting to share audio is `on` by default
- As a precaution confirmed both On and Off states work to show / hide the option on supported platform

LGTM!

PR #1226, opened 2026-06-18
[MM-69189] Restore live participant list for call-post observers

bgardner8008:
## Summary

The participant list on a call's channel post stopped updating for users who hadn't joined the call — it only populated once the viewing user joined. On `main` (v1) the webapp registered channel-wide `user_joined`/`user_left` websocket handlers that kept observers' Redux state in sync. v2 (LiveKit) dropped those two registrations; the only remaining dispatcher of `joinUser`/`leaveUser` is the local LiveKit room, which exists only after you've joined. The server still broadcasts the events channel-wide and the handler functions still existed — they were just orphaned.

## Changes

- Re-register `custom_<pluginId>_user_joined` / `_user_left` handlers in `index.tsx`, reusing the existing `handleUserJoined`/`handleUserLeft`.
- Gate both handlers to skip the channel the viewer is currently connected to: for that channel the LiveKit room stays the source of truth, which avoids racing the local participant feed and the `USER_JOINED` reset of `unmuted/voice/video/raised_hand` to defaults. The gate is per-channel, so a call in a *different* channel still updates live.
- Rename the session reducer's misleading `initialState` parameter to `currentState` (it's the live current state Redux passes, not the initial state). Pure rename, no behavior change.

## Test

Verified locally with two sessions: an observer viewing the channel post without joining sees a second user appear when they join the call and disappear when they leave.

## Still open (separate follow-up)

The server writes the `participants` post prop only at call end (`plugin.go`), never during the call, so a fresh page load / permalink for a pure observer relies on `call_state` hydration. Not addressed here.

Jira: https://mattermost.atlassian.net/browse/MM-69189

### DIFF
--- a/webapp/src/state/session/reducer.ts
+++ b/webapp/src/state/session/reducer.ts
@@ -29,7 +29,7 @@ type State = {
 
 const emptyState: State = {};
 
-export const reducer: Reducer<State, Actions> = (currentState = emptyState, action) : State => {
+export const reducer: Reducer<State, Actions> = (initialState = emptyState, action) : State => {
     switch (action.type) {
     case UN_INITIALIZED: {
         return emptyState;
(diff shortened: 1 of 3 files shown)


webapp/src/state/session/reducer.ts
M-ZubairAhmed, line 1: I don’t think we need to rename initialState to currentState. In this context, initialState refers to the state when the reducer is invoked, while the next state is what the reducer derives from it.

M-ZubairAhmed (changes requested):

webapp/src/state/session/reducer.ts
bgardner8008: OK, I reverted that change

M-ZubairAhmed:
Oops you forgot to git push

M-ZubairAhmed (approved):

PR #1229, opened 2026-06-19
[MM-69368] SIP outbound dev & E2E test tooling (LiveKit SIP stack)

bgardner8008:
## Summary

Tooling to develop and test SIP/VOIP outbound calling (Milestone 2, MM-68360) both locally and in CI. Outbound SIP requires the LiveKit SIP bridge — which needs Redis (it talks to LiveKit only over the PSRPC bus, flipping LiveKit into distributed mode) — plus a SIP endpoint to receive the outbound INVITE.

This is the **tooling/machinery** ticket. Comprehensive SIP E2E **coverage** is tracked separately under the umbrella **MM-69369**, which runs on the harness built here.

## Local dev — additive compose profiles in `docker-compose.livekit.yaml`

- _(no profile)_ — **LiveKit only**, unchanged daily WebRTC driver.
- `sip` — **+ dedicated Redis + LiveKit SIP bridge** (dial an external provider).
- `sink` — **+ Asterisk** auto-answer endpoint (`Answer → Playback → Echo → Hangup`) for manual bidirectional-media validation.

Because the `redis:` block can't be conditional (LiveKit won't boot if it's set but Redis is down), the SIP path runs from a separate `livekit-sip.yaml`; the default `livekit.yaml` stays lean. Three make targets mirror the layers:

- `make livekit-docker-start` (unchanged)
- `make livekit-sip-docker-start` (`sip`)
- `make livekit-sip-sink-docker-start` (`sip,sink`)

The dedicated Redis is deliberately not Mattermost's (avoids an availability dependency + a pub/sub cross-DB collision). The outbound destination (bundled Asterisk vs external provider) is a runtime LiveKit API concern, so the sink stays cleanly opt-in.

## CI — SIPp harness in the E2E stack (`e2e/docker/`)

Adds the same Redis + SIP bridge plus a **SIPp** auto-answer UAS (built from Debian `sip-tester`; public images only ship a floating `latest`). A single `@sip-smoke` test drives an outbound INVITE through the bridge to the SIPp sink **via the LiveKit SIP API** (`createSipOutboundTrunk` → `createSipParticipant` with `waitUntilAnswered`) — proving the harness end to end, independent of the not-yet-shipped SIP feature. The SIP-bridge config (`sip.yaml`) is shared with the dev compose to avoid drift; the e2e LiveKit just gains the Redis block, a second `devkey` keypair, and a `livekit-dev` network alias so the shared config resolves unchanged.

## Testing

Verified locally end to end: booted Redis + LiveKit + SIP bridge + SIPp, drove a real outbound INVITE, and confirmed the call is answered (`sipCallId` returned) with the INVITE captured in SIPp's message trace. `tsc` and eslint pass on the new test.

## Out of scope

- The SIP outbound **feature** (dialing, trunk/dispatch wiring, UI) — MM-68360.
- Comprehensive SIP E2E coverage — MM-69369.

### DIFF
--- a/e2e/tests/sip_smoke.spec.ts
+++ b/e2e/tests/sip_smoke.spec.ts
@@ -1,0 +1,73 @@
+// Copyright (c) 2020-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+/* eslint-disable no-process-env */
+
+import {SIPTransport} from '@livekit/protocol';
+import {expect, test} from '@playwright/test';
+import {RoomServiceClient, SipClient} from 'livekit-server-sdk';
+
+// LiveKit API endpoint + credentials, supplied to the playwright container by
+// e2e/scripts/run.sh. The `devkey` pair is the one shared with the SIP bridge
+// config (e2e/docker/livekit.yaml + ../../sip.yaml).
+const livekitHost = process.env.LIVEKIT_HOST || 'http://livekit:7880';
+const livekitApiKey = process.env.LIVEKIT_API_KEY || 'devkey';
+const livekitApiSecret = process.env.LIVEKIT_API_SECRET || 'this-is-a-32-plus-character-dev-secret';
+
+// Address of the SIPp UAS sink the bridge dials (host:port on the e2e network).
+const sipSinkAddress = process.env.SIP_SINK_ADDRESS || 'sipp:5060';
+
+// This proves the SIP *harness* end to end — LiveKit SIP service is reachable
+// over Redis, an outbound trunk + CreateSIPParticipant makes it send an INVITE,
+// and the SIPp sink answers it — independent of the SIP outbound feature
+// (MM-68360), which the plugin doesn't ship yet. Comprehensive SIP coverage
+// that builds on this harness is tracked in MM-69369.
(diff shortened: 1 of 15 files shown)


e2e/tests/sip_smoke.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+// Address of the SIPp UAS sink the bridge dials (host:port on the e2e network).
+const sipSinkAddress = process.env.SIP_SINK_ADDRESS || 'sipp:5060';
+
+// This proves the SIP *harness* end to end — LiveKit SIP service is reachable
+// over Redis, an outbound trunk + CreateSIPParticipant makes it send an INVITE,
+// and the SIPp sink answers it — independent of the SIP outbound feature
+// (MM-68360), which the plugin doesn't ship yet. Comprehensive SIP coverage
+// that builds on this harness is tracked in MM-69369.
+test.describe('SIP harness smoke', {tag: '@sip-smoke'}, () => {
+    test('outbound INVITE reaches the SIPp sink', async () => {
```
M-ZubairAhmed, line 26: i assume this test is for checking if the sip stack works or not right? because this uses the library as not testing anything of the app

M-ZubairAhmed (approved):

PR #1230, opened 2026-06-22
[MM-69251] Port outbound SIP dialing (server) from livekit-poc to v2

bgardner8008:
## Summary

Ports the **server-side** of LiveKit SIP outbound phone dialing from the `livekit-poc` branch (commit `4c019e7b`) to `v2`. A Mattermost user can dial an external phone number; LiveKit places the outbound SIP call and joins the remote party to a call hosted on the user↔Calls-bot DM channel. This is a port-and-adapt onto v2's existing webhook/session-tracking model (MM-68698 / #1173), not a clean cherry-pick.

Client/webapp/standalone work (tel: interception, phone-mode widget) is tracked in the companion client ticket and is **not** in this PR.

## What's included

- **Config:** `EnableSIPOutbound` toggle and `LiveKitSIPOutboundTrunkID` text setting (both in `plugin.json` + `configuration`). The two are independent gates — turning the toggle off disables dialing without clearing the trunk config.
- **`server/sip.go`:** `createSIPParticipant()` + a `livekitHTTPURL()` helper. Uses the raw `livekit.NewSIPProtobufClient` with explicit SIP (`Admin`+`Call`) and video (`RoomAdmin`+`RoomCreate`) grants — the video grant lets LiveKit auto-create the room when the SIP leg reaches it before the user's join.
- **`POST /phone-call`** (`handlePhoneCall`) + `normalizePhoneNumber()` — resolves the user↔bot DM, validates the toggle/trunk, dials in a goroutine, returns `{channel_id}`.
- **`dial` slash command** — registered with autocomplete; the server handler returns the ephemeral "web/desktop only" fallback (the webapp intercepts it on web/desktop).
- **Teardown:** ends the phone call whenever **either** side leaves. SIP hangup is handled in v2's existing `handleLiveKitSIPParticipantLeft`; the last-human-leaves case is handled in `removeUserSession` (deletes the LiveKit room → SIP BYE), so the PSTN leg is never orphaned.

## Testing

- Unit tests for `normalizePhoneNumber`/`livekitHTTPURL`, the `/phone-call` gating (disabled toggle, missing trunk), the webhook end-call path, and the last-human-leaves teardown.
- Manually verified end to end against a local LiveKit + SIP-bridge + Asterisk sink stack: dial connects with bidirectional audio, and all three teardown paths (SIP hangup, leave, end-for-all) drop the phone leg.
- `go build`, `go vet`, and `gofmt` clean. (golangci-lint couldn't run locally due to a toolchain version mismatch; relying on CI.)

### DIFF
--- a/server/api.go
+++ b/server/api.go
@@ -623,7 +623,9 @@ func (p *Plugin) handlePhoneCall(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	// The call is hosted on the user<->bot DM channel (the phone-call container).
+	// The call is hosted on the user<->bot DM channel. The client creates/joins
+	// the call on this channel before calling /phone-call, so the SIP leg dials
+	// into the existing call's room and the join webhook finds an active call.
 	dmChannel, appErr := p.API.GetDirectChannel(userID, botID)
 	if appErr != nil {
 		res.Err = fmt.Errorf("failed to get bot DM channel: %w", appErr).Error()
--- a/server/api.go
+++ b/server/api.go
@@ -576,6 +576,77 @@ func (p *Plugin) handleGetLiveKitToken(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+// handlePhoneCall dials an external phone number via LiveKit SIP and joins the
+// remote party to a call hosted on the requesting user's DM channel with the
+// Calls bot. LiveKit auto-creates the room on demand, so we dial in a goroutine
+// without waiting for the user's own join flow to create the room first.
+func (p *Plugin) handlePhoneCall(w http.ResponseWriter, r *http.Request) {
+	var res httpResponse
+	defer p.httpAudit("handlePhoneCall", &res, w, r)
+
+	userID := r.Header.Get("Mattermost-User-Id")
+
+	var req struct {
+		Number string `json:"number"`
+	}
+	if err := json.NewDecoder(http.MaxBytesReader(w, r.Body, requestBodyMaxSizeBytes)).Decode(&req); err != nil {
+		res.Err = "invalid request body"
+		res.Code = http.StatusBadRequest
+		return
+	}
+
+	number := normalizePhoneNumber(req.Number)
+	if number == "" {
+		res.Err = "number is required"
+		res.Code = http.StatusBadRequest
+		return
+	}
+
+	cfg := p.getConfiguration()
+	if cfg.EnableSIPOutbound == nil || !*cfg.EnableSIPOutbound {
+		res.Err = "outbound dialing is disabled. Enable it in the admin console."
+		res.Code = http.StatusBadRequest
+		return
+	}
+
+	trunkID := cfg.LiveKitSIPOutboundTrunkID
+	if trunkID == "" {
+		res.Err = "outbound dialing is not configured. Set the SIP Outbound Trunk ID in the admin console."
+		res.Code = http.StatusBadRequest
+		return
+	}
+
+	botID := p.getBotID()
+	if botID == "" {
+		res.Err = "bot not initialized"
+		res.Code = http.StatusInternalServerError
+		return
+	}
+
+	// The call is hosted on the user<->bot DM channel (the phone-call container).
+	dmChannel, appErr := p.API.GetDirectChannel(userID, botID)
+	if appErr != nil {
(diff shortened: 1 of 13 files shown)


bgardner8008:
@dryrunsecurity nit drs_441fe159

Accepted, known scope for this milestone. Outbound dialing is **admin opt-in and off by default** (`EnableSIPOutbound` plus a required `LiveKitSIPOutboundTrunkID`), `/phone-call` runs under the per-user API rate limiter (1/s, burst 10), and each call is created on the initiating user's own Calls-bot DM (attributable + logged via `created SIP participant for outbound call`).

Per-**team** trunk authorization (`TeamIDs` on the trunk definition) is the planned access-control mechanism and is explicitly a future tier (see the "LiveKit Calls Outbound SIP Dialing" design doc). For now an admin controls exposure via the enable toggle and which trunk is configured; restricting "all authenticated users" further is deferred to that future work.

server/api.go
```diff
@@ -576,6 +576,77 @@ func (p *Plugin) handleGetLiveKitToken(w http.ResponseWriter, r *http.Request) {
...
+	dmChannel, appErr := p.API.GetDirectChannel(userID, botID)
+	if appErr != nil {
+		res.Err = fmt.Errorf("failed to get bot DM channel: %w", appErr).Error()
+		res.Code = http.StatusInternalServerError
+		return
+	}
+	channelID := dmChannel.Id
+
+	go func() {
+		if err := p.createSIPParticipant(trunkID, number, channelID, req.Number); err != nil {
```
M-ZubairAhmed, line 638: does this create livekit room internally?

server/api_router.go
```diff
@@ -160,6 +160,9 @@ func (p *Plugin) newAPIRouter() *mux.Router {
 	// LiveKit
 	router.HandleFunc("/livekit-token", p.handleGetLiveKitToken).Methods("GET")
 
+	// Outbound phone call
+	router.HandleFunc("/phone-call", p.handlePhoneCall).Methods("POST")
```
M-ZubairAhmed, line 164: more explicit api name? `create-outbound-sip-call`? or do you plan to add other methods to the same endpoint?

server/configuration.go
```diff
@@ -97,6 +99,9 @@ type ClientConfig struct {
 	EnableSimulcast *bool
 	// When set to true it enables ringing for DM/GM channels.
 	EnableRinging *bool
+	// When set to true it enables outbound SIP phone dialing. Requires a configured
+	// LiveKitSIPOutboundTrunkID; this toggle disables dialing without clearing it.
+	EnableSIPOutbound *bool
```
M-ZubairAhmed, line 104: would require admin console knobs for this as well

server/api.go
bgardner8008: Yes. LiveKit auto-creates the room on demand when the SIP participant connects — that's why the SIP token carries the RoomCreate video grant. Basically all LiveKit room joins will auto-create the room.

However, the client code should create the call on the calls bot channel BEFORE issuing the phone-call dial request. Otherwise the LK room will create and the participant_join web hook will hit the plugin before the client has created call session in DB, and the join hook will be ignored. So in practice the client will be creating the room prior to issuing the outbound phone call.

I need to update the call flow confluence doc to capture the SIP flows as well.

server/api_router.go
bgardner8008: Yes, more methods are planned on the same endpoint: the design doc has DELETE /phone-call/{id} (hang up) and POST /phone-call/{id}/dtmf (send digits). I'll create a followup ticket to implement those.

I'm also seeing that LiveKit supports both call ringback and no-answer timer. The ringback is already enabled, but I forgot to set the 60 sec cancellation timer, will do that.

If/when we implement Calls DM ringback tone and cancel timer, we may revisit this and use the Calls ringback tone and timer but for now it's easiest to use builtin LiveKit functions.

server/configuration.go
bgardner8008: Already done — plugin.json has settings_schema entries for both EnableSIPOutbound ("Enable outbound phone dialing") and LiveKitSIPOutboundTrunkID ("SIP Outbound Trunk ID"), in both the grouped LiveKit section and the flat settings list. And I can confirm they work because I already used them.

Changes take effect after saving (I just had to confirm that!).

M-ZubairAhmed (approved):

PR #1238, opened 2026-06-29
[MM-69553] Replace vendored Mattermost createSelector with the reselect npm package

M-ZubairAhmed:
#### Summary
This pull request updates how the `reselect` library is handled in both the `standalone` and `webapp` projects. The main change is to add `reselect` as a direct dependency, replacing previous custom path mappings to an internal implementation. Additionally, selector definitions in `webapp/src/selectors.ts` have been cleaned up by removing unnecessary string keys passed to `createSelector`. These changes ensure that `reselect` is used as a standard, maintained dependency and simplify selector code to match the official API.

#### Ticket Link

Fixes https://mattermost.atlassian.net/browse/MM-69553

#### Release Note

```release-note

```


M-ZubairAhmed:
@coderabbitai full review

bgardner8008 (approved):
I don't fully understand why the custom selector was used with the apparently unused name, but this seems like nice cleanup. Note that some of the tests failing which I fixed in PR #1243, this branch needs a rebase to pass CI.

M-ZubairAhmed:
> I don't fully understand why the custom selector was used with the apparently unused name, but this seems like nice cleanup. Note that some of the tests failing which I fixed in PR #1243, this branch needs a rebase to pass CI.

It was because we have metrics for selector usage and performance which we dont have anymore in the webapp.

PR #1141, opened 2026-03-25
[MM-68052] Upgrade plugin to  Node.js 24.xx

M-ZubairAhmed:
#### Summary
* Updated the required Node.js version to `^24.14` and npm to `^11` in `.nvmrc`, `standalone/package.json`, `standalone/package-lock.json`, `webapp/package.json`, and `e2e/package.json`
* Upgraded `@types/node` to `24.12.0` in all relevant `package.json` and `package-lock.json` files to match the new Node.js version.

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-68052

PR #1244, opened 2026-07-01
[MM-69618] Fix E2E report generation: use Playwright blob reporter + native merge-reports

bgardner8008:
### Summary

The **E2E Report Generation** workflow has failed on every run since **2026-04-27**, when the Node.js 24 upgrade (#1141) regenerated `e2e/package-lock.json` and floated `@playwright/test` from **1.51.1 → 1.58.2** via its caret range (`^1.51.1`). The merge step relied on the abandoned third-party `playwright-merge-html-reports` (last released 2023), which cannot parse the modern Playwright HTML report format and fails with:

```
Error: index.html does not contain report data
```

This replaces it with Playwright's built-in `merge-reports` (available since 1.37), which merges sharded runs via the `blob` reporter.

### Changes

- **`e2e/playwright.config.ts`** — emit a `blob` report per shard in CI instead of standalone `html` (keep `json` for pass/fail counts).
- **`e2e/scripts/run.sh` / `run-core.sh`** — copy each shard's `blob-report/` out of the container instead of `playwright-report/`.
- **`.github/workflows/generate-e2e-report.yml`** — merge with `npx playwright merge-reports --reporter html` using the pinned version (`npm ci` in `e2e/`); drop the unpinned `npm install` of the dead library.
- Remove the now-dead `e2e/scripts/merge-reports.js`.

Pass/fail counting (`pw-results.json`) and the S3 upload / `e2e/report-summary` status steps are unchanged.

### ⚠️ Testing catch-22: this cannot be fully validated on the feature branch

`E2E Report Generation` is a **`workflow_run`-triggered** workflow (it fires when `E2E Testing` completes). GitHub **always runs a `workflow_run` workflow using the version of that workflow file on the default branch** — never the PR's version. So:

- The `E2E Report Generation` run that fires for this PR executes **`main`'s current (broken) `generate-e2e-report.yml`**, *not* the version in this PR. It will keep failing until this merges — that failure is **expected** and is not a signal about this change.
- Only the `E2E Testing`-side changes (`playwright.config.ts`, `run.sh`, `run-core.sh`) actually run on this branch. This PR's E2E run does produce the new `blob` artifacts and all shards pass, so the "produce blob reports" half is validated live.

**How the un-runnable half was de-risked:** the new merge was exercised locally against the **real `blob` artifacts** produced by this PR's E2E run (Actions run `28523075321`) — downloaded them and ran `playwright merge-reports --reporter html` with the pinned Playwright 1.58.2, confirming a merged HTML report is produced. (Local validation only; see note below.)

**Definitive confirmation happens on merge:** the first `E2E Report Generation` run on `main` after this lands should go green and post the `e2e/report-summary` status. Recommend watching that run immediately post-merge and reverting if it doesn't recover.

### Follow-up

- Same bug exists on the `v2` branch (Playwright 1.59.1) — needs its own port; tracked to follow this PR.
- Optional hardening: pin `@playwright/test` exactly (drop the `^`) in `e2e/package.json` so future lockfile regenerations don't silently bump Playwright. (Already done on `v2`.)

Jira: MM-69618


M-ZubairAhmed (commented):
i was wondering about this one as well last week about why we never ran this?

bgardner8008:
I know these E2E reports are written to s3 bucket, but from there who looks at them? They don't seem to be the ones we see in the PR CI results. I only noticed it because I started getting tons of GitHub bot notifications.

M-ZubairAhmed (approved):

PR #1249, opened 2026-07-03
[MM-69654] Fix /call logs from expanded-view popout

bgardner8008:
## Summary

- Exposes `window.callsClientFlushAndGetLogs()` on the main window — combines flush + read in a single cross-window call, mirroring the existing `callsClientLogAppend` pattern.
- In the `/call logs` handler, detects when running in the popout and delegates the entire flush+read to `window.opener.callsClientFlushAndGetLogs()` instead of operating on the popout's own (nearly empty) buffer and isolated `sessionStorage`.
- Wires `window.addEventListener('error')` and `window.addEventListener('unhandledrejection')` into the calls log buffer in both realms, so uncaught exceptions and rejected promises appear in `/call logs` uploads.

## Root cause

`/call logs` called `flushLogsToAccumulated()` + `getClientLogs()` in the current realm. In the popout:
1. The popout's in-memory `clientLogs` is ~empty — every line is forwarded to `opener.callsClientLogAppend()` and the popout's own buffer is skipped.
2. On web, `getPersistentStorage()` returns `sessionStorage`, which is **not shared** between windows, so the popout's read sees a store that never received the opener's logs.

## Test plan

- [ ] Join a call on web (Firefox or Chrome), open the expanded view popout, run `/call logs` from the popout's chat panel — verify the uploaded file contains full session logs
- [ ] Run `/call logs` from the main window — verify output is identical (same accumulated logs)
- [ ] Trigger an uncaught JS error in the calls plugin (e.g. via devtools) — verify it appears in the next `/call logs` upload
- [ ] Verify no regression on Desktop: `/call logs` from main window still works, and running it from the popout also returns full logs

Note: I ran these tests locally with Safari and Desktop, "/call logs" works from popout channel (log post appears in main thread), and unhandled rejections and uncaught exceptions are both logged.
 
Fixes https://mattermost.atlassian.net/browse/MM-69654

### DIFF
--- a/webapp/src/log.ts
+++ b/webapp/src/log.ts
@@ -17,6 +17,18 @@ let clientLogs = '';
 // usually more useful, entries. Strings and Error stacks are not capped.
 const maxObjectLogLength = 256;
 
+// Flush the in-memory buffer to storage once it exceeds this size. Keeps
+// memory bounded between calls and during plugin-inactive periods when the
+// window error/unhandledrejection listeners are still writing to the buffer.
+// String .length is O(1) in JS so this check is cheap on every write.
+const maxInMemoryLogSize = 50 * 1024;
+
+function maybeFlush() {
+    if (clientLogs.length > maxInMemoryLogSize) {
+        flushLogsToAccumulated();
+    }
+}
+
 function stringifyLogArg(arg: unknown): string {
     if (typeof arg === 'string') {
         return arg;
@@ -37,6 +49,7 @@ function stringifyLogArg(arg: unknown): string {
 // opener's buffer rather than persisting separately.
 function appendLogLine(line: string) {
     clientLogs += line;
+    maybeFlush();
 }
 
 function appendClientLog(level: string, ...args: unknown[]) {
@@ -64,6 +77,7 @@ function appendClientLog(level: string, ...args: unknown[]) {
     }
 
     clientLogs += line;
+    maybeFlush();
 }
 
 // Expose this realm's appender and flush+getter so an expanded-view popout can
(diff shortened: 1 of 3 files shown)


webapp/src/log.ts
```diff
@@ -66,10 +66,29 @@ function appendClientLog(level: string, ...args: unknown[]) {
...
+// Expose this realm's appender and flush+getter so an expanded-view popout can
+// write logs through to (and read them back from) the opener's realm.
 if (typeof window !== 'undefined') {
     window.callsClientLogAppend = appendLogLine;
+    window.callsClientFlushAndGetLogs = flushAndGetLogs;
+
+    // Wire uncaught JS errors and unhandled promise rejections into the client-
+    // log buffer. Without this, exceptions that crash a handler go only to
+    // console.error and never appear in /call logs uploads.
+    window.addEventListener('error', (event: ErrorEvent) => {
```
M-ZubairAhmed, line 92: would it make sense to removeEventListener as well?

M-ZubairAhmed (approved):

webapp/src/log.ts
bgardner8008: Good question. The addEventListener is happening at module load time, so it only happens once at page load. If the user was to disable the plugin, that page would accumulate error logs indefinitely until the next page load. But, if we remove the listeners in uninitialize(), we introduce a bug that disabling/enabling plugin causes the error listener to be removed. So I don't like that, better to leave them.

But this does remind me that log buffer does accumulate potentially without bound, with some non calls specific messages (like translations loading). I used to have a periodic cleanup on a timer every few minutes, which I removed because Jesse was fussing about it. But I like the idea of capping the memory length and will add a check on every log write to prevent boundless accumulation, without a timer. I'll put it on this PR because I plan to backport these changes to v1.

PR #1250, opened 2026-07-03
[MM-69672] Send leave when LiveKit join fails during WS reconnect

bgardner8008:
## Summary

- Adds `sendLeaveAndClose()` to `WebSocketClient`: sends leave immediately when the WS is OPEN; when the WS is CONNECTING (mid-reconnect), queues the leave via `once('open')` so it fires as soon as the pending connection opens and then closes — closing off the ~87s server-side session orphan that occurs when LiveKit fails while the plugin WS is reconnecting
- Replaces the three separate `sendLeave()` + `close()` call pairs in `CallClient` with `sendLeaveAndClose()`
- Fixes a bug in `handleWebsocketErrored` where `ReconnectTimeout` nulled `websocketClient` before calling `disconnect()`, preventing `sendLeaveAndClose()` from running cleanup

## Root cause

When LiveKit returns `JOIN_FAILURE` while the plugin WS is in CONNECTING state (e.g. a 5-second ping timeout triggered a reconnect), `sendLeave()` calls through to `send()` which checks `ws.readyState === WebSocket.OPEN` and silently drops the message. No leave ever reaches the server, so the `CallSession` row is only reaped when the server's own WS pong timeout fires (~87s).

Note: if the underlying TCP connection is dead (not just a transient blip), no client-side fix can help — the server pong timeout remains the reaper in that case. This fix closes the gap for the more common case where TCP is healthy but the application-level ping interval triggered a reconnect.

## Test plan

- [ ] Unit tests: `sendLeaveAndClose` describe block in `websocket_client.test.ts` (3 new tests covering OPEN, CONNECTING, and already-closed cases)
- [ ] Existing disconnect tests updated to assert `sendLeaveAndClose` instead of separate `sendLeave` + `close`
- [ ] All 423 webapp tests pass

### DIFF
--- a/webapp/src/clients/call/call_client.ts
+++ b/webapp/src/clients/call/call_client.ts
@@ -302,8 +302,7 @@ export default class CallClient extends EventEmitter {
 
             // The plugin WS is already open and the join was sent, so tell the server we're
             // leaving before closing — otherwise it keeps the session until its own timeout.
-            this.websocketClient?.sendLeave();
-            this.websocketClient?.close();
+            this.websocketClient?.sendLeaveAndClose();
             this.websocketClient = null;
 
             this.emit(CALL_EVENT.ERROR, err);
@@ -341,8 +340,7 @@ export default class CallClient extends EventEmitter {
 
             // The plugin WS is already open and the join was sent, so tell the server we're
             // leaving before closing — otherwise it keeps the session until its own timeout.
-            this.websocketClient?.sendLeave();
-            this.websocketClient?.close();
+            this.websocketClient?.sendLeaveAndClose();
             this.websocketClient = null;
 
             this.emit(CALL_EVENT.ERROR, err);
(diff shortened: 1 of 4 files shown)


webapp/src/clients/call/call_client.ts
```diff
@@ -302,8 +302,7 @@ export default class CallClient extends EventEmitter {
 
             // The plugin WS is already open and the join was sent, so tell the server we're
             // leaving before closing — otherwise it keeps the session until its own timeout.
-            this.websocketClient?.sendLeave();
-            this.websocketClient?.close();
```
M-ZubairAhmed, line 306: `sendLeave` are now pretty much not used should we remove ?

M-ZubairAhmed (approved):

webapp/src/clients/call/call_client.ts
bgardner8008: Good find. Removing.

PR #1255, opened 2026-07-10
[MM-69685] Fix /call logs from expanded-view popout (v1 backport)

bgardner8008:
## Summary

Backport of MM-69654 / PR #1249 (merged to `v2`) to the `main` (v1/RTCD) branch. All changes are in the webapp layer with no v2-specific dependencies.

- **Fix `/call logs` from the expanded-view popout**: exposes `window.callsClientFlushAndGetLogs()` on the main window. The `/call logs` handler detects when it's running in the popout and delegates flush+read to the opener's realm. Previously the popout's local storage read missed the opener's accumulated logs (on web, `sessionStorage` is per-window), and the popout's in-memory buffer was mostly empty since each line is forwarded to the opener via `window.opener.callsClientLogAppend`.
- **Wire uncaught exceptions into the calls log buffer**: adds `window.addEventListener('error', ...)` and `window.addEventListener('unhandledrejection', ...)` at module load time. Uncaught JS exceptions and unhandled promise rejections now appear in `/call logs` uploads instead of going only to `console.error`.
- **Auto-flush in-memory buffer at 50 KB**: adds a `maybeFlush()` call after every write to `clientLogs`. Prevents unbounded in-memory accumulation between calls and during plugin-inactive periods when the error/unhandledrejection listeners are active.

## Test plan

- [ ] Open a call and join via the expanded-view popout
- [ ] Run `/call logs` from the popout — confirm the upload contains the full call log (not empty/near-empty)
- [ ] Run `/call logs` from the main window — confirm it still works as before
- [ ] Trigger an uncaught JS exception (e.g. via DevTools console: `setTimeout(() => { throw new Error("test") }, 100)`) — confirm it appears in the next `/call logs` upload

### DIFF
--- a/webapp/src/log.ts
+++ b/webapp/src/log.ts
@@ -21,7 +21,12 @@ const maxInMemoryLogSize = 50 * 1024;
 
 function maybeFlush() {
     if (clientLogs.length > maxInMemoryLogSize) {
-        flushLogsToAccumulated();
+        try {
+            flushLogsToAccumulated();
+        } catch {
+            // Storage quota or security error — keep only the most recent portion in memory.
+            clientLogs = clientLogs.slice(-maxInMemoryLogSize);
+        }
     }
 }
 
@@ -90,7 +95,7 @@ if (typeof window !== 'undefined') {
     window.addEventListener('unhandledrejection', (event: PromiseRejectionEvent) => {
         const reason = event.reason instanceof Error ?
             (event.reason.stack || `${event.reason.name}: ${event.reason.message}`) :
-            String(event.reason);
+            formatArg(event.reason);
         appendClientLog('error', `[unhandledrejection] ${reason}`);
     });
 }
--- a/webapp/src/log.ts
+++ b/webapp/src/log.ts
@@ -13,6 +13,18 @@ let clientLogs = '';
 
 const maxArgLength = 256;
 
+// Flush the in-memory buffer to storage once it exceeds this size. Keeps
+// memory bounded between calls and during plugin-inactive periods when the
+// window error/unhandledrejection listeners are still writing to the buffer.
+// String .length is O(1) in JS so this check is cheap on every write.
+const maxInMemoryLogSize = 50 * 1024;
+
+function maybeFlush() {
+    if (clientLogs.length > maxInMemoryLogSize) {
+        flushLogsToAccumulated();
+    }
+}
+
 function formatArg(a: unknown): string {
     if (a instanceof Error) {
         return a.message;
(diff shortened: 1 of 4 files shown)


webapp/src/log.ts
```diff
@@ -9,10 +9,34 @@ import {getPersistentStorage} from 'src/utils';
...
+}
+
 let clientLogs = '';
 
 const maxArgLength = 256;
 
+// Flush the in-memory buffer to storage once it exceeds this size. Keeps
+// memory bounded between calls and during plugin-inactive periods when the
+// window error/unhandledrejection listeners are still writing to the buffer.
+// String .length is O(1) in JS so this check is cheap on every write.
```
lieut-data, line 26: nit: this comment about `.length` feels out of place: we're documenting the const, not the code that later uses it. Maybe move (or remove, if trivial)?

webapp/src/log.ts
```diff
@@ -28,8 +52,59 @@ function formatArg(a: unknown): string {
     return String(a);
 }
 
+// Appends a fully-formatted log line to this realm's in-memory buffer. Exposed
```
lieut-data, line 55: nit: we talk about exposing it, but that's "how it's used", not what this does.

webapp/src/log.ts
```diff
@@ -28,8 +52,59 @@ function formatArg(a: unknown): string {
...
+
+    clientLogs += line;
+    maybeFlush();
+}
+
+// Expose this realm's appender and flush+getter so an expanded-view popout can
+// write logs through to (and read them back from) the opener's realm.
+if (typeof window !== 'undefined') {
+    window.callsClientLogAppend = appendLogLine;
+    window.callsClientFlushAndGetLogs = flushAndGetLogs;
```
lieut-data, line 89: nit: instead of "polluting" `window` directly, do we have the option of extending an existing calls global (i.e. `window.callsClient.*`)?

More salient, the asymmetry between `LogAppend` and `appendLogLine` makes this harder to understand and grep: can we converge?

webapp/src/slash_commands.tsx
```diff
@@ -177,8 +177,24 @@ export default async function slashCommandsHandler(store: Store, joinCall: joinC
...
-        flushLogsToAccumulated();
-        const allLogs = getClientLogs();
+        // When running in the expanded-view popout, delegate flush+read to the
+        // opener's realm. The popout's own in-memory buffer is ~empty (every
+        // appended line is forwarded to opener.callsClientLogAppend), and on
+        // web sessionStorage is per-window so a local read would miss the
+        // opener's accumulated logs entirely.
+        let allLogs: string;
+        try {
+            const opener = window.opener as Window | null;
```
lieut-data, line 187: It feels backward to have every client of these methods have to know about `window.opener`. Could we flip the interface to instead have a common method we call, but that itself resolves the window.opener? This would help focus the code in the caller to the relevant parts (flush, get logs) instead of having to sprinkle awareness of the opener distinction everytime we need to use these methods.

lieut-data (changes requested):

M-ZubairAhmed:
I think this is advantageous to merge in v1 considering v1 isn't going anywhere anytime soon

bgardner8008:
@M-ZubairAhmed I got stuck on Jesse's review, I didn't have time to dive in, so I just dropped it. But it would be nice if call logs could work from expanded view in v1.

bgardner8008:
@M-ZubairAhmed Oh, it's already in v2, this was a back port to v1.

M-ZubairAhmed:
> @M-ZubairAhmed Oh, it's already in v2, this was a back port to v1.

Sorry i mistyped i meant its good to have this in v1

PR #1256, opened 2026-07-10
[MM-69717] Guard join button against concurrent join attempts

bgardner8008:
## Summary

Ports the join re-entrancy guard from MM-69670 (v2) to `main` (v1/RTCD).

- Early-returns from `connectToCall` when `clientConnecting` is already `true`, so rapid or slow-connection clicks cannot start multiple join flows in parallel.
- Adds `isClientConnecting` to the `disabled` prop on the channel header button, so the button is natively disabled (not just visually styled) during the in-flight window.

## Jira

https://mattermost.atlassian.net/browse/MM-69717

## Test plan

- [ ] Click "Join call" rapidly while on a slow/throttled connection — only one join attempt fires.
- [ ] Button is visually and functionally disabled while a join is in progress (spinner showing).
- [ ] Normal join and start-call flows still work.

### DIFF
--- a/webapp/src/components/channel_header_button.tsx
+++ b/webapp/src/components/channel_header_button.tsx
@@ -71,7 +71,7 @@ const ChannelHeaderButton = () => {
         <CallButton
             id='calls-join-button'
             className={'style--none call-button ' + (inCall || restricted ? 'disabled' : '')}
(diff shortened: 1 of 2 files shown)


bgardner8008:
/test-analysis-override changes are bug fixes, not new features

lieut-data (approved):

PR #1264, opened 2026-07-17
[MM-69793] Port blob reporter to v2 branch

bgardner8008:
## Summary

- Switch CI Playwright reporter from `html` to `blob` in `e2e/playwright.config.ts`
- Copy `blob-report` (instead of `playwright-report`) out of the E2E container in `e2e/scripts/run.sh`

This is the v2 follow-up to PR #1244, which fixed the broken `E2E Report Generation` workflow on `main`. The `generate-e2e-report.yml` workflow always runs from `main` (GitHub `workflow_run` semantics) and now expects blob-report artifacts; v2-based feature branches were still producing the old HTML-format artifacts, causing the report generation step to fail with `Error: No report files found in .../all-blob-reports`.

No changes needed to `generate-e2e-report.yml` or `e2e.yml` — the report workflow was already fixed on `main`, and the artifact upload step picks up `results/blob-report-*` automatically.

Jira: MM-69793

### DIFF
--- a/e2e/playwright.config.ts
+++ b/e2e/playwright.config.ts
@@ -71,7 +71,7 @@ const config: PlaywrightTestConfig = {
         },
     ],
     reporter: process.env.CI ? [
-        ['html', {open: 'never'}],
+        ['blob'],
         ['json', {outputFile: 'pw-results.json'}],
         ['list'],
     ] : 'list',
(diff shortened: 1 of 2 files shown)


M-ZubairAhmed:
@bgardner8008 is this PR ready or does it need more work?

PR #1265, opened 2026-07-17
[MM-69745] Parallelize host mute-all; drop redundant plugin WS events

bgardner8008:
## Summary

- `hostMuteAllParticipants` was a sequential loop making two blocking LiveKit Cloud RPCs per participant (`GetParticipant` + `MutePublishedTrack`), causing ~400ms/participant latency — 5+ seconds for a 13-person call observed at the July 8 dev meeting.
- Replaces the sequential loop with a concurrent fan-out (`sync.WaitGroup`), collapsing latency to roughly one RTT pair regardless of participant count.
- Removes the `wsEventHostMute` plugin WebSocket event from both `hostMuteParticipant` and `hostMuteAllParticipants`. `MutePublishedTrack` causes LiveKit to deliver a `RemoteMute` signal over each participant's own signaling connection; the livekit-client SDK handles it by calling `pub.mute()` on the local track — the same effect as the WS-driven path, making it redundant.
- Removes the dead `handleHostMute` handler from `webapp` and `standalone`.

## Test plan

- [ ] Host mute-all in a call with several participants — all participants mute simultaneously rather than sequentially
- [ ] Host mute individual participant — participant is muted
- [ ] Muted participant can still self-unmute afterward (matching v1 semantics)

I tested this feature locally with host and two users and it worked fine.

### DIFF
--- a/server/host_controls.go
+++ b/server/host_controls.go
@@ -5,6 +5,7 @@ package main
 
 import (
 	"fmt"
+	"sync"
 	"time"
 
 	"github.com/mattermost/mattermost/server/public/model"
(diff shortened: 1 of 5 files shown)


bgardner8008:
@coderabbitai review

server/host_controls.go
```diff
@@ -134,21 +135,28 @@ func (p *Plugin) hostMuteAllParticipants(requesterID, channelID string) error {
...
-			"session_id": id,
-		}, &WebSocketBroadcast{UserID: s.UserID, ReliableClusterSend: true})
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			if err := p.livekitMuteParticipant(channelID, composeLivekitIdentity(s.UserID, id)); err != nil && !errors.Is(err, errLiveKitNotConfigured) {
+				p.LogError("hostMuteAllParticipants: failed to mute participant via LiveKit",
+					"channelID", channelID, "sessionID", id, "err", err.Error())
+			}
+		}()
```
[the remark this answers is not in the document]
bgardner8008: Skipping the semaphore — fan-out is bounded by call size, which has a hard cap well below the point where goroutine overhead or LiveKit rate-limiting would be a concern. This is a one-shot, human-triggered operation, not a continuous workload. A semaphore adds complexity without meaningful benefit here.

webapp/src/websocket_handlers.ts
```diff
@@ -402,21 +402,6 @@ export function handleCaption(store: Store, ev: WebSocketMessage<LiveCaptionData
     }, LIVE_CAPTION_TIMEOUT);
 }
 
-export function handleHostMute(store: Store, ev: WebSocketMessage<HostControlMsg>) {
```
M-ZubairAhmed, line 405: i see so no handling of host mute is needed because of Livekit WS

M-ZubairAhmed (approved):

webapp/src/websocket_handlers.ts
bgardner8008: Yes, along with the parallel change I also stripped out the host mute logic from calls websocket. This is part of the "pare it down to nothing, then remove" Calls websocket plan. After that active calls in Calls v2 should be able to withstand pod rolls with no extra work, plus the benefit of simplicity and minimalism.

PR #1266, opened 2026-07-20
[MM-69754] Server: DM auto-end when the other party leaves

bgardner8008:
## Summary

- In `removeUserSession`, after a participant leaves a 1:1 DM call, publishes `wsEventCallEnd` to the remaining participant so they disconnect immediately.
- A 5-second safety-net goroutine (same pattern as `hostEnd`) force-closes any lingering RTC sessions and runs `cleanCallState` if the client doesn't disconnect cleanly on its own.
- Non-DM channels (group messages, open/private) are unaffected — the logic is gated on `model.ChannelTypeDirect`.
- Two unit tests added: one verifying `wsEventCallEnd` is published for DM channels, one verifying it is not published for non-DM channels.

Fixes: https://mattermost.atlassian.net/browse/MM-69754
Part of: https://mattermost.atlassian.net/browse/MM-69735

## Test plan

- [ ] Start a DM call with two users; have one leave — confirm the call ends for both immediately.
- [ ] Start a group message call with 3+ users; have one leave — confirm remaining users stay in the call.
- [ ] `go test ./server -run TestRemoveUserSessionDMAutoEnd` passes.

I tested locally with DM/GM/channel calls and all works as expected.

### DIFF
--- a/server/session.go
+++ b/server/session.go
@@ -483,6 +483,46 @@ func (p *Plugin) removeUserSession(state *callState, userID, originalConnID, con
 		}()
 	}
 
+	// DM auto-end: end the call when a real user leaves and another real user is still connected.
+	if len(state.sessions) > 0 && !state.onlyUserLeft(p.getBotID()) {
+		ch, appErr := p.API.GetChannel(channelID)
+		if appErr != nil {
+			p.LogError("failed to get channel for DM auto-end check", "err", appErr.Error(), "channelID", channelID)
+		} else if ch.Type == model.ChannelTypeDirect {
+			p.LogDebug("DM auto-end: participant left, ending call", "channelID", channelID, "userID", userID)
+			p.publishWebSocketEvent(wsEventCallEnd, map[string]interface{}{}, &WebSocketBroadcast{
+				ChannelID: channelID, ReliableClusterSend: true,
+			})
+
+			callID := state.Call.ID
+			nodeID := state.Call.Props.NodeID
+
+			go func() {
+				time.Sleep(5 * time.Second)
+
+				call, err := p.store.GetCall(callID, db.GetCallOpts{})
+				if err != nil {
+					p.LogError("DM auto-end: failed to get call", "err", err.Error())
+				}
+
+				sessions, err := p.store.GetCallSessions(callID, db.GetCallSessionOpts{})
+				if err != nil {
+					p.LogError("DM auto-end: failed to get call sessions", "err", err.Error())
+				}
+
+				for _, s := range sessions {
+					if err := p.closeRTCSession(s.UserID, s.ID, channelID, nodeID, callID); err != nil {
+						p.LogError("DM auto-end: failed to close RTC session", "err", err.Error())
+					}
+				}
+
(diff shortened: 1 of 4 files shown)


bgardner8008:
@dryrunsecurity fp drs_35cac69a The scenario requires user B to both fail to disconnect cleanly (which is why the goroutine exists) and also successfully rejoin — those are mutually exclusive. If B can disconnect cleanly, the call ends via the normal path before the goroutine fires and it exits early on EndAt != 0. If B can't disconnect cleanly, they can't rejoin either.

server/session.go
```diff
@@ -483,6 +483,46 @@ func (p *Plugin) removeUserSession(state *callState, userID, originalConnID, con
 		}()
 	}
 
+	// DM auto-end: end the call when a real user leaves and another real user is still connected.
+	if len(state.sessions) > 0 && !state.onlyUserLeft(p.getBotID()) {
```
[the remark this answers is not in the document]
bgardner8008: This was fixed

server/session.go
```diff
@@ -483,6 +483,46 @@ func (p *Plugin) removeUserSession(state *callState, userID, originalConnID, con
...
+			p.LogDebug("DM auto-end: participant left, ending call", "channelID", channelID, "userID", userID)
+			p.publishWebSocketEvent(wsEventCallEnd, map[string]interface{}{}, &WebSocketBroadcast{
+				ChannelID: channelID, ReliableClusterSend: true,
+			})
+
+			callID := state.Call.ID
+			nodeID := state.Call.Props.NodeID
+
+			go func() {
+				time.Sleep(5 * time.Second)
```
M-ZubairAhmed: is this standard in golang?

server/session.go
```diff
@@ -483,6 +483,46 @@ func (p *Plugin) removeUserSession(state *callState, userID, originalConnID, con
...
+					p.LogError("DM auto-end: failed to get call", "err", err.Error())
+				}
+
+				sessions, err := p.store.GetCallSessions(callID, db.GetCallSessionOpts{})
+				if err != nil {
+					p.LogError("DM auto-end: failed to get call sessions", "err", err.Error())
+				}
+
+				for _, s := range sessions {
+					if err := p.closeRTCSession(s.UserID, s.ID, channelID, nodeID, callID); err != nil {
```
M-ZubairAhmed, line 518: how would this port to LK look like?

M-ZubairAhmed (approved):

server/session.go
bgardner8008: I assume this is a common way to start a goroutine that will wake up in 5 sec and do a check, though it's actually waking immediately, waiting 5 sec, and proceeding.

server/session.go
bgardner8008: In v2, we would call the LiveKit API to remove the participant from the room, that triggers event on client which closes the client. This is the non-websocket way.

server/session.go
bgardner8008: I updated to use time.After and select on p.stopCh so it will pick up plugin deactivation

PR #1268, opened 2026-07-22
[MM-69757] Play ringback tone for caller of DM/GM call

bgardner8008:
## Summary

- Adds a `useRingback` hook that plays an outbound ringback tone while the caller of a DM/GM call waits for the first other participant to answer
- Ringback stops on connect (another user joins), timeout (30s, caller is disconnected), cancel, or leave
- Reuses the existing `ringForCall`/`stopRingingForCall` ring infrastructure, gated by the `EnableRinging` admin setting
- `RingbackContainer` is a render-nothing global component that mounts the hook once

## Notes

- The ringback sound currently uses `'Calm'` as a placeholder. A dedicated ringback audio asset (`calls_ringback.mp3`) will be added in a follow-up PR to the server webapp, at which point `DEFAULT_RINGBACK_SOUND` here will be updated to point to it.
- The ringback is intentionally fixed (not user-selectable) — it is outbound caller feedback, not an incoming notification.

## Test plan

- [ ] Start a DM call — ringback plays immediately after your own session joins
- [ ] Other user answers — ringback stops
- [ ] Let 30s elapse with no answer — ringback stops, call will persist (unless dm-timer PR has landed)
- [ ] Leave the call manually before timeout — ringback stops
- [ ] Verify ringback does not play for channel calls (non-DM/GM)
- [ ] Verify ringback does not play if you are not the call owner
- [ ] Verify ringback does not play when `EnableRinging` is off in admin console
- [ ] Verify ringback obeys current Calls audio output device

I did a bunch of testing with webapp (Firefox, Safari) and desktop, on GM/DM/channel calls and changed devices. Everything looks good.


webapp/src/index.tsx
```diff
@@ -397,6 +398,7 @@ export default class Plugin {
         registry.registerGlobalComponent(injectIntl(SwitchCallModal));
         registry.registerGlobalComponent(injectIntl(ScreenSourceModal));
         registry.registerGlobalComponent(injectIntl(IncomingCallContainer));
+        registry.registerGlobalComponent(RingbackContainer);
```
M-ZubairAhmed: Do we need this to be a global component? Can we not use this component in the local widget

webapp/src/constants.ts
```diff
@@ -9,7 +9,16 @@ export const REACTION_TIMEOUT_IN_REACTION_STREAM = 10000;
...
+
+// RINGBACK_TIMEOUT is how long the ringback tone plays before the call is
+// automatically cancelled if no other participant has joined.
+export const RINGBACK_TIMEOUT = 30000;
 export const DEFAULT_RING_SOUND = 'Calm';
+
+// DEFAULT_RINGBACK_SOUND is the tone the caller hears while waiting for the
+// other party to answer. Decoupled from the caller's incoming ringtone
+// preference — a ringback tone is a fixed, standardized signal.
+export const DEFAULT_RINGBACK_SOUND = 'Calm';
```
M-ZubairAhmed: Can we add in the ringback tone in here itself?

webapp/src/index.tsx
bgardner8008: Good find, code has been revised to mount on CallWidget component.

webapp/src/constants.ts
bgardner8008: Yes, and doing so means I don't have to make PR on MM repos. Implemented in latest changes.

M-ZubairAhmed (approved):

PR #1277, opened 2026-07-31
[MM-69735] Improve DM call user experience on Web

M-ZubairAhmed:
#### Summary

Makes DM calls behave like a phone call instead of a channel call that happens to have two people in it.

**Ringing and answering**
- The caller's widget and expanded view show `Calling…` (pulsing) until the other party answers, instead of a duration counting up from a call nobody is in yet. Once answered, the timer starts from `0:00` — ring time is excluded.
- The answered timestamp is client-side and shared through the calls window, so a popout opened mid-call picks up the same start time instead of restarting the timer.
- The caller hears a ringback tone while the callee is being rung.

**Ending a DM call**
Four distinct end states, each reflected in the call post:

| What happened | Post status |
|---|---|
| Call completed normally | `ended` |
| Caller hung up before the callee answered | `canceled_by_caller` |
| Callee never answered within 30s | `no_answer` |
| Callee declined | `declined` |

- New `POST /calls/{channel_id}/decline` endpoint lets the callee decline without joining first. It disconnects the caller, ends the call, and dismisses the callee's own incoming-call notification.
- A 30s no-answer timer auto-cancels an unanswered DM call. It is cancelled as soon as a second person joins.
- When one party leaves an answered DM call, the call ends for both rather than leaving the other person alone in an empty call. Clients are asked to disconnect themselves first, with a 5s grace period before sessions are force-closed.

**Call post card**
Reworked into a `PostTypeEvent` component so a single post communicates the outcome, replacing the previous call-card state logic.

#### Behavior changes worth calling out

Two changes here affect more than DM calls, or apply regardless of configuration:

1. **Host "End call for everyone" no longer shows a confirmation modal.** Selecting it from the leave menu ends the call immediately (from #1274). Non-hosts also get single-click leave. This applies to all call types, not just DMs.
2. **The 30s no-answer auto-cancel is not gated on `EnableRinging`.** DM calls are phone-like regardless of that setting, so an unanswered DM call is cancelled after 30s even on servers where ringing is off — which is the default (`EnableRinging` defaults to `false`). With ringing disabled the callee has no incoming-call UI, so they have 30 seconds to notice the call post in the channel. This is intentional; flagging it because it changes behavior on default-configured servers without any setting being touched.

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-69735

#### Release Note
```release-note
Improved the DM call experience to behave like a phone call.
```


webapp/src/components/call_widget/call_widget.scss
```diff
@@ -0,0 +1,189 @@
...
+
+    &>span.noOneSpeakingText {
+      font-weight: 400;
+    }
+  }
+
+  .callStatusTimerCallingText {
+    font-weight: 600;
+    letter-spacing: 0.02em;
+    color: var(--center-channel-color-64, rgba(63, 67, 80, 0.64));
```
matthewbirtch:
```suggestion
    color: var(--center-channel-color-64, rgba(63, 67, 80, 0.75));
```

for improved contrast

webapp/src/components/call_widget/call_widget.scss
```diff
@@ -0,0 +1,189 @@
...
+  border: solid 1px rgba(var(--center-channel-color-rgb), 0.16);
+  box-shadow: 0 6px 14px 0 rgba(0, 0, 0, 0.12);
+}
+
+a.calls-channel-link {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  border-radius: 4px;
+  color: rgba(var(--center-channel-color-rgb), 0.64);
```
matthewbirtch:
```suggestion
  color: rgba(var(--center-channel-color-rgb), 0.75);
```

webapp/src/components/call_widget/call_widget.scss
```diff
@@ -0,0 +1,189 @@
...
+  i {
+    &::before {
+      color: rgba(var(--center-channel-color-rgb), 0.56);
+    }
+  }
+
+  &:link,
+  &:focus,
+  &:visited {
+    color: rgba(var(--center-channel-color-rgb), 0.64);
```
matthewbirtch:
```suggestion
    color: rgba(var(--center-channel-color-rgb), 0.75);
```

matthewbirtch (changes requested):
Nice work on this @M-ZubairAhmed. All looks good, just a few minor things I noticed from my testing:

1. While in the calling state, I wonder if we should disable or hide the record button. Then once it's connected, it should be enabled.
2. While in the calling state, can we update the tooltip on the hangup button to read 'Cancel call' instead of 'Leave call'?
3. I am getting an outbound ring for GM calls. I think we should just leave that for DMs only since we're treating GMs like channels in most other aspects as [stated in the spec doc](https://mattermost.atlassian.net/wiki/spaces/DES/pages/4653645825/Calls+UX+Improvements+for+DM+Calls#GM-Calls). 
4. This is probably not an issue with this PR, more a general issue, but when trying to start a call in poor connectivity (or no connectivity), it infinitely stays in the 'starting call' state and I can't cancel it or close it. Even if I start the call with wifi off, then turn it back on, it doesn't seem to connect.
image

Also, just confirming that we are leaving the 'decline' action as a follow-up for later.

server/i18n/en.json
```diff
@@ -19,10 +19,22 @@
     "id": "app.admin.concurrent_sessions_warning.team",
     "translation": "We highly recommend switching to [Mattermost Enterprise Edition](https://mattermost.com/pl/install-enterprise-install-upgrade) and [deploying the RTCD service](https://mattermost.com/pl/calls-deployment-the-rtcd-service) to offload calls processing to a separate instance in order to maintain the performance, scalability, and reliability of your main Mattermost server."
   },
+  {
+    "id": "app.call.canceled_by_caller_message",
+    "translation": "Call cancelled"
```
matthewbirtch:
```suggestion
    "translation": "Call canceled"
```

American spelling is 'canceled'

bgardner8008:
@matthewbirtch Regarding the ringback tone on GM calls, I'd vote to keep it, as the GM callees are being notified via ringtones. So even though the GM experience still uses the old conference paradigm, the ringback tone mirrors the ringtones at the other endpoints.

bgardner8008 (approved):
I've already reviewed all PRs going onto dm-calls integration branch, so naturally I'm OK with these changes. I do think we should test with video 1:1 enabled to make sure the widget and popout window work as expected before merging.

matthewbirtch:
@M-ZubairAhmed a couple more things I thought of:

- When starting a DM call, it immediately plays the existing call start sound, then plays the outbound ringtone. Can we remove the call start sound in this case? I don't think we need both.
- Also wondering if we can supress the `You are now the host' notice that shows in this context
image

> @matthewbirtch Regarding the ringback tone on GM calls, I'd vote to keep it, as the GM callees are being notified via ringtones. So even though the GM experience still uses the old conference paradigm, the ringback tone mirrors the ringtones at the other endpoints.

@bgardner8008 the weird thing is that the call timer starts immediately with GM calls, so it doesn't feel quite right to have the outbound ring for the caller.

bgardner8008:
@matthewbirtch I don't have strong feelings on this, I just happened to like the symmetry of hearing ringback when the other endpoints are playing ringtones. But I'm OK playing ringback only on DM and leaving GM experience unchanged. We could revisit GM calling experience in future. Also OK suppressing host popup and call start tone in DM just to fully tweak the experience.

webapp/src/components/call_widget/call_widget.scss
M-ZubairAhmed: done

webapp/src/components/call_widget/call_widget.scss
M-ZubairAhmed: done

webapp/src/components/call_widget/call_widget.scss
M-ZubairAhmed: done

server/i18n/en.json
M-ZubairAhmed: done

M-ZubairAhmed:
Thanks @matthewbirtch , all addressed:

Decline UI : confirming that is a follow-up.

matthewbirtch (approved):
> 1. While in the calling state, I wonder if we should disable or hide the record button. Then once it's connected, it should be enabled.

The recording button looks properly disabled in the widget menu, but not in the popout window.
image

And one more new item: just before the calling state shows, it says 'Starting call' and in behind that i shows the 'no one is talking' text. Can we remove this?

Approving this to unblock

DHaussermann:
Hi @M-ZubairAhmed 
I did a round of testing on this and so far it seems in order. I will take one more look soon. 
Just to confirm the scope:
- This is only for webapp so far and does not cover mobile?
- In the UX doc for the card updates there is a UX shown for declined call but as far as I known, you can't decline a call. I don't see anything implemented for this.
image

M-ZubairAhmed:
Hi @DHaussermann 

> * This is only for webapp so far and does not cover mobile?

Yes only webapp

> * In the UX doc for the card updates there is a UX shown for declined call but as far as I known, you can't decline a call. I don't see anything implemented for this.

Yes its not implemented in this Epic

server/api.go
```diff
@@ -292,6 +292,118 @@ func (p *Plugin) handleDismissNotification(w http.ResponseWriter, r *http.Reques
...
+		p.unlockCall(channelID)
+		return http.StatusBadRequest, fmt.Errorf("no call ongoing")
+	}
+
+	// The callee has no active session; the caller does. Reject if the requester
+	// owns a session (they would be the caller, not the callee).
+	for _, sess := range state.sessions {
+		if sess.UserID == userID {
+			p.unlockCall(channelID)
+			return http.StatusForbidden, fmt.Errorf("caller cannot decline their own call")
```
carlisgg: the comment could be wrong in case the session belongs to the callee having answered on another device already, right?

server/api.go
```diff
@@ -292,6 +292,118 @@ func (p *Plugin) handleDismissNotification(w http.ResponseWriter, r *http.Reques
...
+		}
+	}
+
+	// If not exactly one session, the callee may have already joined (race). Bail silently.
+	if len(state.sessions) != 1 {
```
carlisgg: is it possible that some session belongs to a bot? I guess not for DM calls since the call screen is not shown until the call is accepted right?
But maybe even the caller has the channel with the ringing call open in 2 devices, will it make this condition satisfied?

server/api.go
```diff
@@ -292,6 +292,118 @@ func (p *Plugin) handleDismissNotification(w http.ResponseWriter, r *http.Reques
...
+	}
+	nodeID := state.Call.Props.NodeID
+	callID := state.Call.ID
+	postID := state.Call.PostID
+	participants := mapKeys(state.Call.Props.Participants)
+
+	if state.Call.Props.DismissedNotification == nil {
+		state.Call.Props.DismissedNotification = make(map[string]bool)
+	}
+	state.Call.Props.DismissedNotification[userID] = true
```
carlisgg: `setCallEnded` on next line will set `call.Props.DismissedNotification = nil` so there is not point on this map creation and assignment right?

server/api.go
```diff
@@ -292,6 +292,118 @@ func (p *Plugin) handleDismissNotification(w http.ResponseWriter, r *http.Reques
...
+
+	setCallEnded(&state.Call)
+
+	if err := p.store.UpdateCall(&state.Call); err != nil {
+		p.LogError("declineCall: failed to update call", "channelID", channelID, "err", err.Error())
```
carlisgg: hmmm, just logging an continuing, won't it leave a zombie call in DB? Maybe we can unlock and return 500 and let the call ring to an end or callee retrying instead of bailing.

server/state.go
carlisgg, line 413: with current ordering when `declining` is clearing `RTCDHost` right?

server/dm_timer.go
```diff
@@ -0,0 +1,111 @@
...
+
+// A var rather than a const so tests can shorten it.
+var dmNoAnswerTimeout = 30 * time.Second
+
+// How long we wait after telling clients a DM call ended before force-closing any RTC
+// sessions still connected, giving them the chance to disconnect on their own first.
+var dmAutoEndGracePeriod = 5 * time.Second
+
+func (p *Plugin) startDMNoAnswerTimer(channelID, callID string) {
+	p.dmNoAnswerTimersMut.Lock()
```
carlisgg, line 55: should this run even when `EnableRinging` is set to `false`?

server/dm_timer.go
```diff
@@ -0,0 +1,111 @@
...
+	delete(p.dmNoAnswerTimers, channelID)
+	p.dmNoAnswerTimersMut.Unlock()
+
+	state, err := p.lockCallReturnState(channelID)
+	if err != nil {
+		p.LogError("handleDMNoAnswer: failed to lock call", "channelID", channelID, "err", err.Error())
+		return
+	}
+
+	if state == nil || state.Call.ID != callID || len(state.sessions) != 1 {
```
carlisgg: could this be the caller having 2 devices open reporting more than 1 session?

server/session.go
```diff
@@ -473,16 +473,77 @@ func (p *Plugin) removeUserSession(state *callState, userID, originalConnID, con
...
-			_, err := p.updateCallPostEnded(state.Call.PostID, mapKeys(state.Call.Props.Participants))
+			_, err := p.updateCallPostEnded(state.Call.PostID, participants, reason)
 			if err != nil {
 				p.LogError("failed to update call post ended", "err", err.Error(), "channelID", channelID)
 			}
 		}()
 	}
 
+	// DM auto-end: end the call when a real user leaves and another real user is still connected.
+	if userID != p.getBotID() && len(state.sessions) > 0 && !state.onlyUserLeft(p.getBotID()) {
```
carlisgg: can a user join a DM call from 2 devices?

webapp/src/components/leave_call_menu.tsx
```diff
@@ -18,13 +20,30 @@ type Props = {
...
 export const LeaveCallMenu = ({channelID, isHost, numParticipants, leaveCall}: Props) => {
     const {formatMessage} = useIntl();
+
+    const dispatch = useDispatch();
+
     const isAdmin = useSelector(isCurrentUserSystemAdmin);
-    const showEndCall = (isHost || isAdmin) && numParticipants > 1;
+    const shouldShowWarningMenuItemForEndingCall = (isHost || isAdmin) && numParticipants > 1;
+
+    async function handleHostEndCallForEveryone() {
```
carlisgg, line 29: this is now the default, maybe we should mention in PR summary

webapp/src/selectors.ts
```diff
@@ -262,6 +278,57 @@ export const callOwnerIDForCallInChannel = (state: GlobalState, channelID: strin
...
+
+// Checks if the current user owns the DM call, has joined, and no one else has joined yet
+// i.e. the caller is waiting for the other party to answer.
+export const isCurrentDMCallInCallingState: (state: GlobalState) => boolean =
+    createSelector(
+        'isCurrentDMCallInCallingState',
+        isCurrentUserOwnerOfCurrentCall,
+        isCurrentUserInSessionForCurrentCall,
+        channelForCurrentCall,
+        sessionsForOtherUsersInCall,
```
carlisgg, line 309: maybe we can also add `dmCalleeAnsweredAtForCurrentCall` with `!answeredAt` to make sure this is actually happening on a never answered state.

webapp/src/components/call_status_timer/elapsed_timer.tsx
```diff
@@ -38,12 +37,7 @@ export default function CallDuration(props: Props) {
...
     });
 
-    const style = props.style || {};
-    if (!style.fontWeight) {
-        style.fontWeight = 600;
-    }
-
     return (
-        <div style={style}>{getCallDuration(adjustedStartAt)}</div>
+        <div className='callDurationContainer'>{getCallDuration(adjustedStartAt)}</div>
```
carlisgg: couldn't find `callDurationContainer` in the codebase

M-ZubairAhmed: added the styles

webapp/src/selectors.ts
M-ZubairAhmed: changed

server/api.go
M-ZubairAhmed: added the filter

server/api.go
M-ZubairAhmed: correct

server/api.go
M-ZubairAhmed: returned

server/state.go
M-ZubairAhmed: yes

server/dm_timer.go
M-ZubairAhmed: filtered

server/session.go
M-ZubairAhmed: yes two sessions of same user can be in a dm with another

server/dm_timer.go
M-ZubairAhmed: DM calls are meant to be phone-like regardless of the ringing setting

server/api.go
M-ZubairAhmed: fixed

DHaussermann (approved):
Tested and passed
- Explored all UX changes discussed in the doc 
- Calls widget goes into a 'calling state' and shows the name of the person you're calling and does **not** yet start a timer
- Calls widget timer starts when 2nd participant joins
- Tested expanded view and timer while in the "Calling" state  
- Tested that UI elements in expanded view such as  chat, participants, raise hand etc... are functional  during `calling` state 
- If you start a call in the same thread as an exiting call that ended the ringing and calling state still work as expected
- Tested that changes re: calling state are specific to DMs and not other channel types 
- Tested simplification to leave call UX
- Leave UX changes for DM apply to widget, expanded view and Call Card 
- End call for all confirmation for admin removed from all channel types
- Call in a DM w\ 3 participants (1 user has 2 clients connected) still follows DM leave logic
- Briefly regression tested mobile

LGTM! 
Thanks @M-ZubairAhmed

hmohammed-prodsec (approved):
LGTM :)

carlisgg (approved):
Thanks for addressing comments. LGTM

webapp/src/components/expanded_view/component.tsx
```diff
@@ -131,6 +132,8 @@ interface Props extends RouteComponentProps {
     openModal: <P>(modalData: ModalData<P>) => void;
     enableVideo: boolean;
     otherSessions: UserSessionState[];
+    isDMCalling: boolean;
+    clientConnecting: boolean;
```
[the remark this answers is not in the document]
M-ZubairAhmed: expanded view is not allowed to opened in the clientConnecting state, so this is not a problem

PR #1287, opened 2026-09-01
[MM-70435] Migrate DM-calls feature to v2

M-ZubairAhmed:
#### Summary
Port of https://github.com/mattermost/mattermost-plugin-calls/pull/1277 to v2

#### Ticket Link

Fixes https://mattermost.atlassian.net/browse/MM-70435

#### Release Note

```release-note

```


server/api.go
```diff
@@ -291,6 +291,98 @@ func (p *Plugin) handleDismissNotification(w http.ResponseWriter, r *http.Reques
...
+
+	// More than one distinct user connected means the call was answered elsewhere before this
+	// decline landed. Nothing to do, and nothing to report as an error either.
+	if len(state.distinctNonBotUserIDs(p.getBotID())) != 1 {
+		return http.StatusOK, nil
+	}
+
+	callID := state.Call.ID
+
+	p.endDMCallRoom("declineCall", channelID)
```
bgardner8008, line 329: Should also call cancelDMNoAnswerTimer here to stop the timer.

bgardner8008:
An issue not in diff: In call_client.ts:1082 DISCONNECTED should be emitted before this.room is nulled. The handler calls getSessionID() which tests room and returns null if room is null so then leaveUser is not dispatched on this path, though it should later be dispatched by user_left handler.

```
  this.emit(CALL_EVENT.DISCONNECTED, reason);
  this.room = null;

```

bgardner8008 (changes requested):
Looks great, thanks. Just a few things discovered.

server/api.go
M-ZubairAhmed: fixed

M-ZubairAhmed:
> An issue not in diff: In call_client.ts:1082 DISCONNECTED should be emitted before this.room is nulled. The handler calls getSessionID() which tests room and returns null if room is null so then leaveUser is not dispatched on this path, though it should later be dispatched by user_left handler.
> 
> ```
>   this.emit(CALL_EVENT.DISCONNECTED, reason);
>   this.room = null;
> ```

Yes also added in the hostendcall path as well

bgardner8008 (approved):
Thanks Zubair. I pushed some fixes to get the E2E tests working again. I will approve and merge.

PR #1293, opened 2026-09-09
Bump Go to 1.26.7 on v2

bgardner8008:
Mirrors #1286, which bumped Go to 1.26.7 on `main` but was never applied to `v2`.

## Why now

`make i18n-check` runs `go install github.com/mattermost/mattermost/tools/mmgotool@latest`. That resolved today to `v0.0.0-20260909101603-f45e25d54b17`, which requires Go >= 1.26.7:

```
go: github.com/mattermost/mattermost/tools/mmgotool@latest:
    github.com/mattermost/mattermost/tools/mmgotool@v0.0.0-20260909101603-f45e25d54b17
    requires go >= 1.26.7 (running go 1.26.4; GOTOOLCHAIN=local)
make: *** [Makefile:175: i18n-check] Error 1
```

`main` satisfies that; `v2` was on `.go-version` 1.26.4 and `go.mod` 1.26.3. With `GOTOOLCHAIN=local` in CI there is no auto-upgrade, so it hard-fails.

**This is not specific to any one PR — every PR against `v2` is red until this lands.** PRs whose runs predate today's upstream publish look green but would fail on re-run.

Same failure mode as MM-68971 (`e757b53b`), which was the previous time `mmgotool` moved its Go floor ahead of ours.

## Changes

`.go-version` and the `go` directive in `go.mod`, `build/go.mod`, `lt/go.mod`, `server/public/go.mod` — all 1.26.4/1.26.3 → 1.26.7.

Excludes main's `.cursor/Dockerfile` change from #1286; that file doesn't exist on this branch. It was the only cherry-pick conflict.

## Verification

- `go build ./server/...` clean
- `make gomod-check` passes
- `go install github.com/mattermost/mattermost/tools/mmgotool@latest` now succeeds, resolving the same version CI failed on

Note for anyone pulling this: `go.work` is gitignored and pins its own `go` directive, so re-run `make setup-go-work` after updating or local builds will still report the old version.

## Follow-up worth its own ticket

`mmgotool@latest` in `i18n-check` is unpinned, so any upstream Go bump breaks CI on every branch that hasn't caught up — twice now. Pinning it would turn these into deliberate updates rather than surprise breakages.


bgardner8008:
@coderabbitai review

M-ZubairAhmed (approved):

PR #1295, opened 2026-09-10
Calls v1: clear stale join-call popups on WS reconnect [MM-70670]

bgardner8008:
## Summary

- The `incomingCalls` reducer had no `UNINIT` case, so a \"Join Call\" popup could survive indefinitely if the `call_end` WS event arrived during a reconnect dead window.
- Adding `case UNINIT: return []` clears all pending notifications on reconnect; `onActivate`'s channel fetch re-adds only still-active calls.
- Added reducer tests for add/remove and UNINIT behaviour.

## Test plan

- [ ] Start a DM call (User A). Confirm User B sees the \"Join Call\" popup.
- [ ] Simulate User B's WS disconnect (e.g. `! fiber-walk` reconnect trick or brief network blip).
- [ ] End the call while User B is reconnecting.
- [ ] Confirm the popup is gone after User B reconnects (no stale popup, no accidental new call on click).
- [ ] Repeat with a call that is still active when User B reconnects — confirm popup reappears after reconnect.
- [ ] `make test` passes.

### DIFF
--- a/webapp/src/reducers.test.ts
+++ b/webapp/src/reducers.test.ts
@@ -2,7 +2,7 @@
 // See LICENSE.txt for license information.
 
 import type {AnyAction} from 'redux';
-import {CALL_END, CALL_STATE, DM_CALLEE_ANSWERED_AT, UNINIT} from 'src/action_types';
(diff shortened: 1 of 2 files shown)


M-ZubairAhmed (approved):
Nice find

PR #1297, opened 2026-09-11
[MM-70690] Calls v2: SIP outbound phone number allowlist

bgardner8008:
## Summary

Adds an optional outbound phone number allowlist to the SIP dialing feature, intended for development and testing environments to prevent accidental calls to real numbers.

- Two new server-only config fields (not exposed to clients): `EnableSIPOutboundAllowlist` (bool toggle, default off) and `SIPOutboundAllowlist` (longtext textarea — one number per line; commas and semicolons also accepted as separators)
- When the toggle is on, `handlePhoneCall` rejects any number not in the list with HTTP 403
- Entries are normalized via the existing `normalizePhoneNumber` before comparison, so formatting variations (`+1 555-123-4567`, `15551234567`) match correctly
- Empty allowlist with toggle on = fail-closed (all calls rejected)
- Settings appear in the admin console SIP Outbound section (on-prem only)

## Test plan

- [x] `EnableSIPOutboundAllowlist = false` (default): all numbers pass through
- [x] `EnableSIPOutboundAllowlist = true`, number not in list: HTTP 403 returned
- [x] `EnableSIPOutboundAllowlist = true`, number in list: call proceeds
- [x] `EnableSIPOutboundAllowlist = true`, empty list: all calls rejected
- [x] Unit tests: `TestIsNumberInAllowlist` (9 cases — separators, normalization, empty list)
- [x] Integration tests: 4 new subtests in `TestHandlePhoneCall`
- [x] Manually validated end-to-end with local LiveKit SIP stack

Jira: https://mattermost.atlassian.net/browse/MM-70690

### DIFF
--- a/server/configuration.go
+++ b/server/configuration.go
@@ -58,6 +58,10 @@ type configuration struct {
 	LiveKitAPISecret string
 	// SIP outbound trunk ID for outbound phone calls (e.g., ST_xxx). Empty disables outbound dialing.
 	LiveKitSIPOutboundTrunkID string
+	// When set to true, only numbers in SIPOutboundAllowlist may be dialed. Intended for dev/test environments.
+	EnableSIPOutboundAllowlist *bool
+	// Newline/comma/semicolon-separated list of E.164 phone numbers permitted for outbound dialing when the allowlist is enabled.
+	SIPOutboundAllowlist string
 	// When set to true live captions will be enabled when starting transcription jobs.
 	EnableLiveCaptions *bool
 	// The speech-to-text model size to use to transcribe live captions.
@@ -158,6 +162,9 @@ func (c *configuration) SetDefaults() {
 	if c.EnableSIPOutbound == nil {
 		c.EnableSIPOutbound = model.NewPointer(false)
 	}
+	if c.EnableSIPOutboundAllowlist == nil {
+		c.EnableSIPOutboundAllowlist = model.NewPointer(false)
+	}
 	if c.TranscriberModelSize == "" {
 		c.TranscriberModelSize = transcriber.ModelSizeDefault
 	}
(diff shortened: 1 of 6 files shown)


bgardner8008:
@coderabbitai review

server/configuration.go
```diff
@@ -58,6 +58,10 @@ type configuration struct {
 	LiveKitAPISecret string
 	// SIP outbound trunk ID for outbound phone calls (e.g., ST_xxx). Empty disables outbound dialing.
 	LiveKitSIPOutboundTrunkID string
+	// When set to true, only numbers in SIPOutboundAllowlist may be dialed. Intended for dev/test environments.
+	EnableSIPOutboundAllowlist *bool
+	// Newline/comma/semicolon-separated list of E.164 phone numbers permitted for outbound dialing when the allowlist is enabled.
```
M-ZubairAhmed: why support these many seperator? just a comma would have been sufficient

M-ZubairAhmed (changes requested):
Do we need the System console part of this as well?

server/configuration.go
bgardner8008: Probably my fault for asking claude about use of semicolon and comma in phone numbers for hard pause and delayed pause. Allowing newline does let us copy/paste from a list, one number per line. So I'd want both newline and comma, might a well allow semicolon too. It's not clear to me yet if the whitelist is just for us to deploy on production (legal wants it), or is it potentially useful to customers.

bgardner8008:
@M-ZubairAhmed Regarding system console, yes we need it to be able to edit the whitelist on production or cloud test.

M-ZubairAhmed:
> @M-ZubairAhmed Regarding system console, yes we need it to be able to edit the whitelist on production or cloud test.

We are planning to include that in PR?

server/configuration.go
bgardner8008: If I had to pick one, I would choose newline for easy editing of list.

bgardner8008:
CleanShot 2026-09-14 at 14 39 02@2x

server/configuration.go
M-ZubairAhmed: yes i think new line is easier to read for phone numbers as well. lets fix on this one itself then

M-ZubairAhmed (approved):

PR #1298, opened 2026-09-11
MM-70684 Update Calls for Mattermost v12

cpoile:
#### Summary
Calls device selectors can crash on React 19 because React Select 3 calls the removed `ReactDOM.findDOMNode` API and uses legacy context. Upgrade the selector dependency and adapt its callbacks while preserving the existing styling.

Require Mattermost 12 and resolve ReactDOM client and JSX runtime imports to the host exports, keeping plugins and their dependencies aligned with the host React. Use development mode for debug builds. The independent desktop widget and recording applications retain their bundled React 18.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-70684

#### Release Note
```release-note
Updated Calls for React 19 and fixed device selector compatibility. Calls now requires Mattermost 12.0 or later.
```


bgardner8008 (approved):
Hi Chris, thanks for picking this up, otherwise I would have been wondering why Calls was crashing in v12! I'm not a react expert at all, but I want through everything with claude and it all looks good.

M-ZubairAhmed (approved):

PR #1300, opened 2026-09-14
Update Calls for Mattermost v12

M-ZubairAhmed:
#### Summary
Cherry-pick of #1298 onto `v2`.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-70684

#### Release Note
```release-note
Updated Calls for React 19.
```


M-ZubairAhmed:
> [!IMPORTANT]
> **Do not merge yet — blocked on the Mattermost 12 release.**
> This branch declares `min_server_version: 12.0.0`, but Mattermost 12 is not
> released: there is no `release-12.x` branch or `v12*` tag in
> `mattermost/mattermost`, and no `release-12*` or `12.x` image tag in the dev
> registry. `master` is currently the only ref whose `version.go` reports
> `12.0.0`, so E2E had to be pointed there to run at all.
>
> Before merging, re-pin `MM_SERVER_IMAGE_TAG` from `master` to `release-12.0`
> (or whichever ESR is designated) to restore the immutable pin this branch
> previously had on `release-11.7`.

M-ZubairAhmed:
Unblocked — `release-12.0` now exists upstream, so the temporary `master` target is gone.

- `mattermost/mattermost` has a `release-12.0` branch whose `version.go` reports `12.0.0`.
- It carries the React 19 host exports this branch needs: `window.ReactJSXRuntime`, `window.ReactJSXDevRuntime`, and `ReactDOM` merged with `react-dom/client`.
- The dev registry has a matching `release-12.0` image tag, so the single variable still works as both the image tag and the `mattermost/mattermost` checkout ref.

`MM_SERVER_IMAGE_TAG` is now `release-12.0`, which restores the fixed-release pin the workflow had on `release-11.7` and keeps upstream `master` churn out of Calls E2E.

One caveat before marking this ready: Mattermost 12 has not reached GA. The only v12 release so far is `v12.0.0-rc1` (2026-09-18); the release branch is cut but the final tag is not out. Leaving this as a draft until E2E is green and 12.0.0 ships.

M-ZubairAhmed:
### E2E results on Mattermost 12

The `release-12.0` pin worked: the plugin installs cleanly and **the React 19 host externals are confirmed working against a real Mattermost 12 server.** 53 tests passed, including full call join and screenshare flows, with no `ReactJSXRuntime` / `ReactJSXDevRuntime` / `createRoot` / `findDOMNode` or undefined-global errors anywhere in the logs. That was the one genuinely unverified risk in this change.

Two tests fail, neither caused by this PR:

```
1) tests/livekit_smoke.spec.ts:16:13 › LiveKit URL field renders @livekit-smoke
   Error: strict mode violation:
   getByTestId('PluginSettings.Plugins.com+mattermost+calls.livekiturlinput') resolved to 2 elements

2) tests/start_call.spec.ts:356:9 › switching products › playbooks @livekit
   Locator: locator('#product-switcher-menu-dropdown div').filter({ hasText: 'Playbooks' })
   Error: element(s) not found

  2 failed  83 skipped  53 passed (15.3m)
```

**1. Duplicate LiveKit URL field — pre-existing `v2` bug, fixed by #1310.**
`plugin.json` on `v2` declares `LiveKitURL` twice: once under `settings_schema.sections` (line 23) and again in the flat `settings` array (line 356), so the admin console renders two inputs and the strict-mode locator fails. #1310 ([MM-70874] Fix duplicate LiveKit settings in admin console) is open against `v2` and fixes exactly this.

**2. Playbooks missing from the product switcher.**
The Playbooks entry isn't in the MM12 product switcher menu. This looks like a test that needs updating for v12 rather than a Calls regression.

Both tests passed on the `release-11.7` image a day earlier and both fail deterministically across retries on `release-12.0`, so this is a Mattermost 12 environment difference surfaced by the image pin, not flake and not this PR's code. Worth noting the suite does not normally pass 100%; upstream #1298 reported 182/256 passing and was still green.

**Blocked on #1310.** This PR can't go green until that merges into `v2` and this branch is rebased. The Playbooks test needs a separate decision — someone who knows the MM12 product-switcher changes should confirm whether that assertion is still valid.

bgardner8008 (approved):
Looks great, thanks for this.

PR #1301, opened 2026-09-14
[MM-69502] Replace dedicated Calls WebSocket with LiveKit connection

M-ZubairAhmed:
#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-69502

#### Release Note

```release-note
NONE
```


bgardner8008:
Two issues caught during review and fixed on top of the branch:

**`ScreenStartAt` units mismatch** (`livekit_webhook.go`): `handleLiveKitTrackPublished` was storing `time.Now().UnixMilli()` into `state.Call.Props.ScreenStartAt`, but all three consumers (`clearScreenSharingState`, `endEmptyCall`, `removeUserSession`) pass that value to `secondsSinceTimestamp()` which calls `time.Unix(ts, 0)` — treating milliseconds as seconds, resolving to year ~55000, and producing a large negative `ScreenDuration` in call stats for every screen share. Fixed to `time.Now().Unix()`, matching the RTCD path.

**`livekitDeleteRoom` held under call lock in SIP hangup path** (`api.go`): When the last SIP participant leaves a phone call channel, `livekitDeleteRoom` was called synchronously while the call lock was still held, risking up to a 5-second block (the LK API timeout) on any concurrent webhook or join for that channel. The other two callsites in `livekit_webhook.go` already run this in a goroutine for exactly this reason. Fixed to match.

bgardner8008:
**Host badge not updating in popout after host transfer** (`host_controls.spec.ts:326`)

The `HOST_CHANGED` event fires on the shared `callsClient` (via LiveKit room metadata) but only `index.tsx` registered a listener — which runs in the main window context. The popout has its own independent Redux store, so the dispatch never reached it. The host badge and host-changed notice were left stale in the popout for both participants.

Fixed by adding an `onClient(CALL_EVENT.HOST_CHANGED)` handler inside the `window.opener` gate in `expanded_view/component.tsx`, consistent with how `USER_JOINED`, `USER_LEFT`, and other LiveKit events are already handled for the popout's store.

bgardner8008 (approved):
Eureka, it finally passes CI and E2E tests! Thanks! I feel like we should merge this and move ahead. Though if you want to run coderabbit we could possible do that and suffer through it now. It will happen eventually when we merge to main. I leave it to you.

PR #1302, opened 2026-09-15
[MM-68638] E2E for disconnect cleanly on tab close to remove the 30s server-side timeout window

M-ZubairAhmed:
#### Summary
* Added a new test suite `abandoned session` in `e2e/tests/join_call.spec.ts` to check that closing a participant's tab removes them from the call for other users.

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-68638

#### Release Note

```release-note
NONE
```


bgardner8008 (approved):
Thanks Zubair. Happy to approve this but the test is somewhat moot when client-over-lk lands. This is because the beforeUnload will cause immediate participant disconnect in LiveKit and client will be notified quickly, not through Calls WebSocket disconnect and 10 sec wait for reconnect (Calls v1 and Calls v2 prior to client-over-lk). So this test should probably be modified then, or just wait until that lands and refactor.

M-ZubairAhmed:
> Thanks Zubair. Happy to approve this but the test is somewhat moot when client-over-lk lands. This is because the beforeUnload will cause immediate participant disconnect in LiveKit and client will be notified quickly, not through Calls WebSocket disconnect and 10 sec wait for reconnect (Calls v1 and Calls v2 prior to client-over-lk). So this test should probably be modified then, or just wait until that lands and refactor.

Yes we have to change the implementation for the test again after that, so merging this now is good so we can change that once the other lands.

PR #1305, opened 2026-09-17
[MM-69461] Calls v2: SIP outbound — report dial failures to caller

bgardner8008:
Fixes https://mattermost.atlassian.net/browse/MM-69461

#### Summary

Two commits:

**1. Synchronous dial + host-controls SIP identity fix**

`POST /phone-call` previously dispatched the SIP dial in a goroutine: a bad trunk ID, carrier rejection, or unresponsive LiveKit returned `200 {channel_id}` and left the caller in a silent call with the error only in server logs.

- Dial is now synchronous, bounded by a 5s timeout. With `WaitUntilAnswered` unset, LiveKit dispatches the INVITE asynchronously and returns near-instantly (confirmed in LiveKit SIP source: `DialAsync` spawns a goroutine and returns before the INVITE is sent) — the timeout only guards against an unresponsive LiveKit SIP service.
- On failure, the handler returns `500 "failed to dial number"`.
- Drive-by: SIP twirp client now uses `cfg.getLiveKitURL()` so `MM_CALLS_LIVEKIT_URL` override applies.

`livekitIdentity(s, sessionID)` helper fixes all five host-control operations (mute, mute-all, lower-hand, remove, screen-off) targeting SIP participants: they were composing `sip:+15551234___PA_abc123` instead of the actual identity `sip:+15551234` (set as `ParticipantIdentity` at dial time, no suffix).

**2. Self-contained endpoint**

`POST /phone-call` now handles the full outbound call setup in one request. The client no longer needs to pre-join the bot DM channel before calling this endpoint.

Server-side flow:
1. Validate number, check allowlist, check SIP + LiveKit config.
2. `GetDirectChannel(userID, botID)` — resolve (or create) the bot DM channel.
3. Create a LiveKit session (`addUserSession` + `mintLiveKitToken`).
4. Dial via `createSIPParticipant`.
5. If either token minting or dial fails, roll back the session (and call row if this was the first joiner).
6. Announce the call only after a successful dial — observers and the DM ring timer should not fire for a call that never connected.
7. Return `{session_id, token, url, call_id, channel_id, sip_call_id}`.

`rollbackSession` helper extracted and used in both `handlePhoneCall` and `handleCreateLiveKitSession`.

#### Note on dropped endpoints

The original ticket scope included `DELETE /phone-call/{id}` (hang-up) and `POST /phone-call/{id}/dtmf`. Both were implemented and deliberately removed — see the Jira ticket. Short version: every teardown trigger is already covered by existing paths, and LiveKit has no server-side DTMF RPC (`publishDtmf` in livekit-client is the right path for MM-69423).

#### Release Note

```release-note
NONE
```


bgardner8008:
@coderabbitai review

bgardner8008:
@coderabbitai review

M-ZubairAhmed:
@bgardner8008 can you take a look at these conflicts

server/session.go
```diff
@@ -793,6 +793,21 @@ func (p *Plugin) cancelDMNoAnswerTimerIfAnswered(state *callState, userID, chann
...
+// it deletes the session row, and if this join created the call row (createdCall),
+// deletes that too. Nothing was announced, so this is a plain delete rather than
+// the call-ended path.
+func (p *Plugin) rollbackSession(sessionID, callID string, createdCall bool) {
+	if err := p.store.DeleteCallSession(sessionID); err != nil {
+		p.LogError("rollbackSession: failed to delete call session", "err", err.Error(), "sessionID", sessionID)
+	}
+	if createdCall {
+		if err := p.store.DeleteCall(callID); err != nil {
+			p.LogError("rollbackSession: failed to delete call", "err", err.Error(), "callID", callID)
```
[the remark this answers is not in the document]
bgardner8008: This is a large refactor for a failure mode that's already harmless: if DeleteCallSession succeeds but DeleteCall fails, the orphan call row has no session and will be swept up by the reconciliation loop. Skipping this.

server/api.go
```diff
@@ -793,21 +809,173 @@ func (p *Plugin) handlePhoneCall(w http.ResponseWriter, r *http.Request) {
...
+		p.rollbackSession(sessionID, state.Call.ID, createdCall)
+		res.Err = "Internal server error"
+		res.Code = http.StatusInternalServerError
+		return
+	}
+
+	// Dial synchronously so a failed dial reaches the caller as an error. We
+	// don't set WaitUntilAnswered, so LiveKit returns once the SIP leg is created
+	// (INVITE dispatched) rather than once the callee picks up.
+	info, err := p.createSIPParticipant(trunkID, number, channelID, req.Number)
```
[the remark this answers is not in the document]
bgardner8008: fixed

server/activate.go
```diff
@@ -161,6 +161,11 @@ func (p *Plugin) OnActivate() (retErr error) {
 
 	go p.reconciler()
 
+	p.publisherWg.Add(1)
```
M-ZubairAhmed: why add another waiting group?

webapp/src/types/global.d.ts
```diff
@@ -11,6 +11,7 @@ import type {CallActions, CurrentCallData} from 'src/types/types';
 declare global {
     interface Window {
         callsClient?: CallClient,
+        callsClientLastChannelID?: string,
```
M-ZubairAhmed: where are we using this?

webapp/src/index.tsx
```diff
@@ -175,6 +174,7 @@ import {
     shouldRenderDesktopWidget,
 } from './utils';
 import {
+    applyCallHostChanged,
```
M-ZubairAhmed: where are we using this?

server/activate.go
bgardner8008: Good catch, this is duplicate code from rebase. Fixed.

webapp/src/types/global.d.ts
bgardner8008: It's to ensure user gets an error modal when kicked from call due to user_removed. When server removes a user from channel, it kicks user from LK room, and the disconnect may hit the client before the user_removed message. So this is how we know the channel ID of the last call. It's possible we should do this a different way.

webapp/src/index.tsx
bgardner8008: applyCallHostChanged is called from two different paths, either from mattermost websocket message for observers not in call, or from LK room metadata (for participants in call).

webapp/src/types/global.d.ts
M-ZubairAhmed: But does it have to be on the window would be my next question?

bgardner8008: From claude: The only reason it's on window is that both index.tsx (where the call join happens and the value needs to be set) and websocket_handlers.ts (where it needs to be read) are separate modules with no direct import relationship — window is the shared mutable state they already use for callsClient itself.

I'm looking into whether we could simply show the modal when user kicked from room, it could be due to host command or user removed from channel, just show a generic "you were removed from call" message. That eliminates the race and hacky logic.

bgardner8008: It's a bit complicated and we just get a generic "you were removed from call" message losing the specific reason. Here's possible logic:

  Host-kick path:
  1. host_removed WS → show modal, call client.disconnect()
  2. LiveKit DISCONNECTED (CLIENT_INITIATED) → no modal (not PARTICIPANT_REMOVED)
  3. user_removed WS → no modal (user is no longer in call, check fails)
  
  Channel-remove path:
  1. LiveKit DISCONNECTED (PARTICIPANT_REMOVED) → show modal
  2. user_removed WS → no modal (just Redux cleanup)
  
  User leaves call normally:
  1. client.disconnect() → DISCONNECTED (CLIENT_INITIATED) → no modal
  2. No user_removed event 
  
  Reconciler sweep (edge):
  1. LiveKit DISCONNECTED (PARTICIPANT_REMOVED) → show modal ("you were removed from the call") — acceptable, something was
     wrong anyway

  The only loose end: handleUserRemovedFromChannel currently shows different messages depending on whether the user left
  themselves (userLeftChannelErr) or was removed (userRemovedFromChannelErr). Moving to PARTICIPANT_REMOVED loses that
  distinction — you'd show one generic message for both channel-leave-while-in-call and being-kicked-from-channel. But since the
  user is asking for "removed from call" without specificity, that's by design.

  One thing to verify: does handleUserRemovedFromChannel also handle the case where the current user leaves the channel (not
  removed by someone else)? If so, that path currently shows userLeftChannelErr, which is a different message. You'd want to
  confirm that case also routes through CLIENT_INITIATED (since the user chose to leave).

bgardner8008: What I'd like to do is send LK data message from server which would put up the appropriate message, but no way to guarantee delivery of that before server kicks the user, which then means no message delivery.

M-ZubairAhmed: It does not need to be on window. index.tsx already imports the websocket handlers, so a module variable is enough. Please remember the channel id only for a PARTICIPANT_REMOVED disconnect, and clear it on a normal leave and after the modal is shown. As written, the id is never cleared, so a later channel leave shows the in-call error. Also, callsClientLastChannelID is declared twice in global.d.ts

server/api.go
```diff
@@ -799,21 +815,173 @@ func (p *Plugin) handlePhoneCall(w http.ResponseWriter, r *http.Request) {
 	}
 	channelID := dmChannel.Id
 
-	go func() {
-		if err := p.createSIPParticipant(trunkID, number, channelID, req.Number); err != nil {
-			p.LogError("handlePhoneCall: failed to create SIP participant",
-				"err", err.Error(), "number", number, "channelID", channelID)
+	sessionID := model.NewId()
+
+	state, err := p.lockCallReturnState(channelID)
```
M-ZubairAhmed: handlePhoneCall holds the cluster call lock across the SIP dial. The lock is taken before the session is saved and released only after createSIPParticipant returns. That call can take the full 5 second timeout. publishCallRoomMetadata drops this same lock before it talks to LiveKit, because holding it blocks every other operation on that call, including plugin shutdown. Do the dial after unlockCall, then lock again to announce or roll back.

server/api.go
```diff
@@ -799,21 +815,173 @@ func (p *Plugin) handlePhoneCall(w http.ResponseWriter, r *http.Request) {
...
 	}); err != nil {
 		p.LogError("failed to encode phone-call response", "err", err.Error())
 	}
 }
 
+// handleOutboundCall dials an external phone number into an existing call. The
+// client must already have a live session in the bot DM channel before calling
+// this endpoint. Useful for testing and future multi-party dial-in scenarios.
+// Returns {call_id, channel_id, sip_call_id} without a LiveKit token.
+func (p *Plugin) handleOutboundCall(w http.ResponseWriter, r *http.Request) {
```
M-ZubairAhmed: Still trying to understand difference between outboundcall and phonecall end points, and which to use

M-ZubairAhmed: why does handleOutboundCall not checks that that the user is on an allowed team?

server/api.go
```diff
@@ -729,10 +729,21 @@ func (p *Plugin) mintLiveKitToken(userID, channelID, sessionID string) (string,
...
+	URL       string `json:"url"`
+	CallID    string `json:"call_id"`
+	ChannelID string `json:"channel_id"`
+	SIPCallID string `json:"sip_call_id"`
+}
+
+// handlePhoneCall is a self-contained outbound-call endpoint: it resolves the
+// bot DM channel, creates a LiveKit session, dials the SIP number, and returns
+// everything the client needs to connect in a single round trip.
 func (p *Plugin) handlePhoneCall(w http.ResponseWriter, r *http.Request) {
```
M-ZubairAhmed: Can you please add call_state to the /phone-call response. Please use the same CallStateClient shape that POST /livekit-token returns. The client will skip /livekit-token and join with the session and token from /phone-call, so without call_state the widget has no call data until later events arrive.

server/api.go
```diff
@@ -799,21 +815,173 @@ func (p *Plugin) handlePhoneCall(w http.ResponseWriter, r *http.Request) {
...
+
+	// Announce only after a successful dial: observers and the DM ring timer
+	// should not fire for a call that never connected.
+	if len(state.sessions) == 1 {
+		p.announceCallStarted(state, userID, channelID, "", "", dmChannel.Type)
+	}
+	p.cancelDMNoAnswerTimerIfAnswered(state, userID, channelID, dmChannel.Type)
+	p.maybeSendConcurrentSessionsWarning()
+
+	p.LogInfo("phone call session created",
```
M-ZubairAhmed: Please set props.type = "phone" (plus phone_number and display_number) when /phone-call creates the call, and include those props in the client call state. Phone mode in the widget needs a reliable way to tell phone calls apart, and the Outbound SIP spec keys on props.type rather than "DM with the Calls bot". The server could also move its own isPhoneCallChannel checks to this later.

server/api.go
```diff
@@ -799,21 +815,173 @@ func (p *Plugin) handlePhoneCall(w http.ResponseWriter, r *http.Request) {
...
+		res.Err = "outbound dialing is not configured. Set the SIP Outbound Trunk ID in the admin console."
+		res.Code = http.StatusBadRequest
+		return
+	}
+
+	if cfg.sipOutboundAllowlistEnabled() && !cfg.isNumberInAllowlist(number) {
+		res.Err = "number is not in the outbound calling allowlist"
+		res.Code = http.StatusForbidden
+		return
+	}
```
M-ZubairAhmed: Right now "number not on the allowlist" and "user not in an allowed team" both return 403 with free-text messages. Could the response include a stable error id for each, like sip_number_not_allowed and sip_team_not_allowed? Same for the 400s: invalid_number, outbound_disabled, outbound_not_configured. The UI shows different text for each, and I'd rather not match on message strings.

server/api.go
M-ZubairAhmed: On Desktop, the call runs in the separate widget window, not the main window. The widget only gets call_id, title, root_id and job_id from its URL, so it can't receive the session /phone-call created. It would call /livekit-token and create a second session for the same user.

The first session is never confirmed. The reconciler skips unconfirmed sessions, so it stays in the call. When the user hangs up, that leftover session still counts as a person, so "last human left, hang up SIP" never fires. The phone leg stays connected to an empty call and keeps using a trunk line. I found this by reading the code, not by running it.

For the Desktop fix, I see two options:

- Dial from the widget. This needs the Desktop app to pass the number in the widget URL.
- Have the server drop or reuse the pending session when the same user joins the same phone call again.

server/api.go
bgardner8008: Agreed, the pattern in the code is to do the LK API calls in goroutines or outside of lock so we don't hold the lock.

server/api.go
bgardner8008: You want phone-call. outbound-call is the old way of making a SIP call from an existing call, which is useful for debugging in the absence of UI, and *might* have use in future.

bgardner8008: Because claude just ported the old code which didn't test team. Good catch, will fix.

server/api.go
bgardner8008: Sounds good.

server/api.go
bgardner8008: OK

server/api.go
bgardner8008: So maybe return json with fields including error code, or include an easy to parse code in the text, or a stable text id. Seems we're not using the text (client would need to translate) so we really just need a code, the text is only useful when developing with curl or whatnot. Text IDs work for both.

Seems you prefer the text IDs like "sip_team_not_allowed", good idea, I'll use those.

server/api.go
bgardner8008: Oof. Need to think about that a bit more. I don't like the idea of always creating and dropping a session. Is there no way to pass the number as a URL parameter?

server/api.go
M-ZubairAhmed: Can we be more explicit with the endpoint name to show its usage?

bgardner8008: How's about phone-call and add-phone-call, or sip-call and add-sip-call?

webapp/src/types/global.d.ts
bgardner8008: OK, I revised the code. Also, the new variable is participantRemovedChannelID to be more explicit.

server/api.go
bgardner8008: OK, that is fixed.

server/api.go
bgardner8008: changed to phone-call and add-phone-call, that's more accurate.

bgardner8008: done

server/api.go
bgardner8008: done

server/api.go
bgardner8008: done

server/api.go
bgardner8008: done

server/api.go
bgardner8008: OK, implemented the second "reuse" option above, which was pretty easy and clean. It also allows easy testing by issuing phone-call via curl and then joining call on calls-bot DM channel.

PR #1306, opened 2026-09-19
[MM-70870] Cover with test for DM auto-end when the last of a user's two devices leaves

M-ZubairAhmed:
## Summary

An existing subtest asserts that closing one of two devices does **not** end the call, but nothing asserted that closing the *second* device does end it (the remaining-user count drops to the callee alone). This adds that subtest.

Fixes https://mattermost.atlassian.net/browse/MM-70870


bgardner8008 (approved):
Despite claude trying to nitpick absolutely everything, I think this looks fine, it tests what it's supposed to.

PR #1309, opened 2026-09-22
Add minimal AGENTS and CURSOR markdown files for AI

M-ZubairAhmed:
#### Summary
This is not the final document, its a continuous improved one

#### Release Note

```release-note
NONE
```


bgardner8008 (approved):
Hi Zubair, thanks for this. I don't use cursor so no comments there but the agents instruction looks useful. I would want to add note to always check style before pushing commits (to avoid linter CI errors). Another is to remind agent of developer builds, running "make deploy" locally fails on MacOS unless you have MM_SERVICESETTINGS_ENABLEDEVELOPER=true set. I suppose we'll have tons of ideas, we just need to agree on a minimal set that we can both live with.

